Issue #811, opened 2026-07-19
Warriors of Whimsy: the level-jump ability, and the WOW/Coven modifier swap


pixieofhugs:
Warriors of Whimsy's identity is being rebuilt (#784 moved its pink `.exe` aesthetic to Cozy Coven). This issue gives WOW its **mechanical** identity; the yellow visual identity is a sibling issue.

## Decisions (settled by the owner — do not re-litigate)

- **WOW's ability: once per level, sign up for ONE task one level above the character's level.** At level 3, a WOW member may sign up for a single level-4 task. Using it consumes the allowance until they reach level 4.
- **WOW drops to full baseline modifiers.** `own_task_modifier`, `other_task_modifier`, `collab_own_modifier`, `collab_other_modifier` all `1.0`. The level-jump is WOW's only perk.
- **Cozy Coven gains the collaboration bonus:** `collab_own_modifier=1.1`. Everything else stays `1.0`.
- Verified: WOW was the ONLY faction carrying a non-1.0 modifier. After this change Coven is the only one.

## The seam

`backend/services/praxis.py:584` `meets_task_level(character_level, task) -> bool` is documented as "the single home for the task-level gate" (#292). It returns `character_level >= task.level_required`. `check_signup_eligibility` consults it and returns `SignupDenialReason.below_level`.

Because that gate is shared by every signup mode (solo, collab, duel), extending it there gives the ability consistent behaviour everywhere for free. **Do not add per-mode special cases.**

## Storage — the cheap shape

Add ONE nullable column to the per-era character stats row: `level_jump_used_at_level: int | None`.

The allowance is available iff `level_jump_used_at_level != stats.level`. This needs **no reset job**: levelling up changes `stats.level`, which makes the allowance available again automatically. And because stats rows are already per-(character, era), an era reset clears it for free.

Do NOT build a counter plus a reset task. If a counter seems necessary, the approach has drifted.

## Config

`FactionConfig` (`backend/game_config.py`) has no field for this. Add one — do not hardcode the slug. Never branch on `slug == "wow"` in a service; read the config field. Suggested: an int (levels above own that may be reached, `0` for everyone else, `1` for WOW), which reads honestly and leaves room without adding speculative machinery.

## Defaults taken (state them in the PR; flag if you disagree)

- **Consumed on signup, never refunded.** Abandoning or deleting the praxis does NOT return the allowance — otherwise it can be farmed by signing up and backing out.
[...]

Issue #292, opened 2026-06-30
Extract named homes for bundled gating sub-rules (level/vote/flag/sign-up)


pixieofhugs:
## Why

Spun out of #171 (faction seam). While extracting `faction_permits`, we noticed the
same smell recurs across the codebase: **a meaningful gating sub-rule buried inside a
broader "eligibility" predicate that ANDs unrelated axes together.**

The motivating example: `is_task_eligible_for_character` (praxis.py) bundles a **level**
check and a **faction** check in one function — so "where do faction rules live?" lands
on a function whose name says "eligible" and whose first gate is level. #171 carved the
faction half out into a named home; the level half (and likely the sign-up / vote / flag
predicates) still bundle unrelated sub-rules.

## Scope (to be designed, not yet decided)

- Audit the gating predicates (`is_task_eligible_for_character`, the sign-up guard
  `_check_create_preconditions`, vote eligibility, flag eligibility, comment gates) for
  bundled-but-unrelated sub-rules.
- Decide which axes deserve their own named single-purpose home (like `faction_permits`)
  vs. which are fine bundled.
- This is **code hygiene with no gameplay-rule change** — unlike #171 it has no
  SPEC-game-rules obligation. Keep it off the gameplay-design review surface.

## Non-goals

- No behavior change. Pure extraction/renaming so each axis has a discoverable home.
- Not blocking #171 — that ships the faction seam first; this generalizes the pattern after.

## Related

- #171 — established the pattern (`faction_permits` in `faction_service.py`).
- ADR-0024 (lands with #171) — its "Consequences" notes this general smell.

pixieofhugs:
**ADR-number note:** the "ADR-0024 (lands with #171)" reference above is stale — 0024 already shipped as `0024-in-progress-praxis-is-private.md`. #171's ADR will land under the next unused number (0025+); this issue's "Consequences" pointer should follow whatever number #171 actually takes.

pixieofhugs:
The **sign-up** slice of this audit is now filed as a concrete, ready-to-build issue: #330 (single `evaluate_signup` predicate, fixing the `can_submit_praxis` bank-cap drift). Split out rather than folded in because it carries a behavior change (correcting the flag) that this issue's non-goals exclude. #292 stays the umbrella for the remaining vote / flag / level axes — tick the sign-up bullet when #330 lands.

pixieofhugs:
[...]

PR #825, opened 2026-07-20
Document that duel acceptance bypasses the task-level gate

### LINKED ISSUE #815: Duel acceptance bypasses the shared task-level gate
Found while building #811 (the Warriors of Whimsy level-jump ability), verified against `origin/main`. **Pre-existing — not introduced by that work.**

## The gap

`meets_task_level` in `backend/services/praxis.py` is documented as "the single home for the task-level gate" (#292), and `check_signup_eligibility` owns "the four gates every claim shares". That holds for solo signups, collab signups, and *initiating* a duel.

It does **not** hold for *accepting* one. `respond_to_duel_challenge` builds the opponent's praxis inline and checks only:

- `era.duel_level_required`
- the bank cap
- active-member status

It never consults `task.level_required`.

## Consequence

**Accepting a duel already puts any character on an above-level task**, regardless of level, for free. The level gate that every other signup path enforces is simply absent on this one.

This also means the invariant #811 relies on is narrower than stated: the shared gate covers every signup mode *except* duel acceptance.

## Why it needs its own issue rather than a drive-by fix

Tightening this could break **in-flight duels** — a challenge already issued to a character who cannot clear the task's level would become unacceptable, potentially stranding it. That needs a deliberate decision, not a quiet patch:

- Do existing pending challenges get grandfathered, or invalidated?
- Should the challenge be blocked at *issue* time instead (validate the opponent can clear the level before the challenge is created), which is friendlier than failing at accept time?
- Should a WOW member be able to spend their #811 level-jump allowance to accept an above-level duel? (Consistent, but means duel acceptance must consume the allowance too.)
- Or is above-level duel acceptance actually *intended* — a duel being a deliberate reach beyond your level?

That last possibility is real and should be ruled in or out before any code changes. If it is intended, the fix is documentation, not enforcement.

## Suggested first step

Confirm intent with the owner before designing. If enforcement is wanted, gating at challenge-creation time is likely both friendlier and cheaper than gating at acceptance.

Related: #292 (the gate consolidation), #811 (level-jump ability).

pixieofhugs:
Closes #815

## What changed

Doc-only, per the owner's ruling on #815 (2026-07-19): duel acceptance intentionally has no task-level gate — `era.duel_level_required` is the only level check on the opponent, by design. No enforcement behavior changes.

1. **`backend/services/praxis.py`** — fixed `meets_task_level`'s docstring, which falsely claimed "identical behaviour across solo, collab and duel." It now states the gate covers solo signup, collab signup, and duel *initiation* (the challenger) only, and calls out duel *acceptance* as a deliberate exception, citing ADR-0049.
2. **`backend/services/duel.py`** — added a `# ponytail:` intent comment (matching the repo's existing convention) beside the `duel_level_required` check in `respond_to_duel_challenge`, so a future gate-consolidation pass doesn't silently re-add a task-level check here.
3. **`docs/adr/0049-duel-acceptance-bypasses-the-task-level-gate.md`** — new ADR recording the ruling, narrowing the #292 "single task-level gate" invariant to exclude duel acceptance, cross-referencing ADR-0011, #811, #292, #815.
4. **`backend/tests/integration/test_duel_service.py`** — new pin test: an opponent at `era.duel_level_required` but well below `task.level_required` still accepts a duel successfully via `respond_to_duel_challenge`.

## Out of scope (left untouched)

`backend/tests/unit/test_meets_task_level.py` (~line 30-32) has a comment repeating the same now-stale "identical behaviour across solo/collab/duel" claim. It doesn't assert anything false, but the wording drifted from the new docstring/ADR — worth a follow-up cleanup, not part of this issue's 4-item scope.

## Test results

- `pytest tests/integration/test_duel_service.py -v` → 1 passed
- `pytest tests/unit/test_meets_task_level.py tests/unit/test_praxis_mode_gates.py tests/integration/test_praxes.py tests/integration/test_duel_service.py -v` → 100 passed
- Full suite: `pytest --cov=. --cov-fail-under=80` → **679 passed**, coverage **89.06%**

---
_Generated by [Claude Code](https://claude.ai/code/session_01LLM7XgppATXJFZfKjAFcjM)_


pixieofhugs:
**Renumbered 0049 → 0051.** ADR-0049 and ADR-0050 were claimed by the praxis-card post-mortem (PR #845, merged as `492249a`) while this PR sat open as a draft. This branch now carries `docs/adr/0051-duel-acceptance-bypasses-the-task-level-gate.md`, and the four in-code references in `duel.py`, `praxis.py` and `test_duel_service.py` were updated to match.

My collision, not this PR's — I wrote 0049/0050 without checking open PRs for reserved numbers. No behaviour change; left as a draft since merging it wasn't part of the current batch.

Issue #840, opened 2026-07-20
WOW + Coven praxis cards, stamps and vote widgets — rebuild from source (ADR-0050)


pixieofhugs:
## Parent
#818

## What to build

Rebuild the WOW and Coven praxis cards, stamps and vote widgets from the vendored design sources. These two are their own slice because they were the most wrong and they share the copy catalog.

Read `.design-sync/praxis-cards/LABELS.md` first. Per **ADR-0050**, the design artifacts label these two backwards — go by tokens and metaphor, never filenames.

### WOW — the chronicle

- **Card**: cream ground, gold rule, plum-purple ink. Eyebrow `CHRONICLE OF PROOF · № 663`. Dashed gold media slot. Archaic register throughout: `for the quest —`, `Sealed by the Court`, `here, an illumination`.
- **Score box**: `base` + numeral + a **purple** `×0.80` chip, `+ 4 from votes`, gold rule.
- **Total mark**: the total followed by **`✦`** — the one place the retired `✦` survives (design README explicitly carves this out).
- **Widget**: googly **balloons** with the per-tier size ramp, plus the rainbow sweep behind the row. Caption `Cast thy Verdict`; picked tag reads `thou dubbed it legendary!`.
- **Tier ladder**: `…excellent · legendary`.

### Coven — the pink sticker

- **Card**: `#fffdfa` cream-pink ground, `1px solid #f3b6d2`, `border-radius: 16px`, Caveat throughout. **Build the dark variant too** — the design has one (deep plum-black ground, pink dashed sticker) and it is not a cascade freebie.
- **Score box**: a `2px dashed` **sticker** at `rotate(-3deg)`, radius 16, with `✨` after the total. This is Coven's total mark — it is not a rectangle.
- **Widget**: **moon phases** on a night plate — port from `FactionVoteWidgets.jsx` (`WowVote` in the bundle; the moons are Coven's, see ADR-0050). Caption `how'd this land?`; picked reads `magical · YAY!`.
- **Tier ladder**: `sweet · lovely · wonderful · magical · iconic`.

### Rules that apply to both

- Port the widget from `FactionVoteWidgets.jsx`; do not reimplement.
- **Ornament exemption**: take the design's caption sizes (Wow 19px Caveat), not `--text-content`.
- Motion stays reduced-motion-gated via CSS class, never inline `animation:`.
[...]

Issue #838, opened 2026-07-20
Swap the WOW and Coven palettes back (ADR-0050) — token repoint, blocks the epic


pixieofhugs:
## Parent
#818

## What to build

Swap the WOW and Coven palettes back, at the token layer only. **This blocks all card, stamp and widget work in the epic** — building nine stamps against a palette assignment already known to be wrong is wasted work, and flipping the praxis card alone would leave each faction contradicting itself across its other surfaces.

Full lineage is in **ADR-0050**. Short version: the faction was `gestalt` (pink, moons) → renamed Warriors of Whimsy → #784 split Cozy Coven off it, taking the then-current gold/plum chronicle → #814 gave WOW yellow. #821 then read a stale design heading and inverted the two.

Target assignment:

| | **WOW** | **Coven** |
|---|---|---|
| palette | cream / gold / plum chronicle | pink sticker (light **and** dark) |
| glyph | `✦` | `✨` |

Coven currently owns gold/plum across 27 registered surfaces; WOW owns 3. Every surface already reads `var(--faction-<slug>-*)`, so moving the **values** moves all 30 through the cascade in one change.

Scope:

- Repoint `--faction-coven-*` / `--coven-*` to the pink family (`#ec5f99`, `#f3b6d2`, `#fffdfa`, plus the moon tokens the pre-split design system carried as `--gestalt-moon-lit` / `--gestalt-moon-shadow`).
- Repoint `--faction-wow-*` to cream / gold / plum. Retire the invented `--faction-wow-balloon-*` yellow ramp authored in #830 — it has no design source.
- Swap the inverted tier words in `locales/en/votes.json`: `…excellent · legendary` is **WOW's**; `sweet · lovely · wonderful · magical · iconic` is **Coven's**. Same for idle/tag copy (`Cast thy Verdict` / `thou dubbed it legendary!` vs `how'd this land?` / `magical · YAY!`).

Two things that do **not** move:

- **The rainbow spine.** WOW's yellow stop in `--faction-default-rainbow` is spectrum membership, not skin. Leave it.
- **Dark mode is re-derived, not translated.** Coven's dark surfaces were tuned for plum-on-black; pink does not clear the same contrast slots by substitution. Check every dark slot and fix what fails.

## Acceptance criteria

- [ ] All 30 WOW/Coven surfaces render the swapped palettes with no per-surface edits.
- [ ] Tier words and caption copy swapped in `votes.json`.
[...]

Issue #837, opened 2026-07-20
Vendor the praxis-card design sources into the repo (delete on epic close)


pixieofhugs:
## Parent
#818

## What to build

Vendor the design sources into the repo so agents can **open** them. This is the root cause of the #821 drift: the vote widgets came out near-perfect (their source, `FactionVoteWidgets.jsx`, was readable) and the card chrome came out wrong (its source, the `.dc.html`, was not). Subagents cannot reach `claude.ai` design URLs from a worktree, and a design summarised into prose cannot carry geometry.

Create `.design-sync/praxis-cards/` containing:

- `design_handoff_faction_vote_stamps/` — the full local handoff bundle (`README.md`, `FactionVoteWidgets.jsx`, `animations.css`, `Faction Praxis Cards.dc.html`, `Singularity Mobile Card.dc.html`, `image-slot.js`, `enso-detailed.svg`, `lotus.svg`).
- `wow-kit/` — DesignSync project `d0a6fdd7-a2f3-4f61-834a-3a06ab4acf07` (`Crest.dc.html`, `WoW Faction Kit.dc.html`, `WoW Task Card.dc.html`).
- `coven-kit/` — DesignSync project `32c7198b-e7e7-43be-ad4f-590309b1093d` (the nine `Warriors of Whimsy *` surfaces + `whimsy-kit.jsx` + `tokens/`).

**Then correct the labels in the vendored copy.** Per ADR-0050, every artifact here is named from the pre-split `gestalt` era and the two factions are crossed throughout — filenames, section headings and sample bylines all disagree with what they depict. Add a `LABELS.md` at the root stating plainly which visual belongs to which slug, and name the two kit directories for what they actually contain (`wow-kit` = the cream/gold/plum chronicle, `coven-kit` = the pink card).

Add a `README.md` noting the directory is **temporary** and is deleted by the epic's final cleanup issue.

## Acceptance criteria

- [ ] `.design-sync/praxis-cards/` contains all three bundles, committed.
- [ ] `LABELS.md` records the WOW/Coven assignment from ADR-0050.
- [ ] Kit directories named for the faction they depict, not their filenames.
- [ ] `README.md` marks the directory delete-on-completion.
- [ ] No lint/CI rule trips on the vendored HTML/JSX (exclude from eslint if needed).

Issue #821, opened 2026-07-20
Praxis card: redesign all 8 faction skins + bespoke vote widgets (desktop + mobile, light + dark)


pixieofhugs:
## Parent
#818

## What to build

Apply the redesign to all **8 faction skins** — desktop frame + mobile frame + bespoke vote widget + faction-styled score stamp, **light and dark** — following the pattern established in #820. The mock widgets are canonical; the shipped widgets (placards, ink-stamps, hearts, seals, rubber-stamps, signal-keys) are **retired**. One issue because the desktop skins share `PraxisCard.tsx` and the manifest list `factions/index.ts`.

Design: [Faction Praxis Cards](https://claude.ai/design/p/14f9466e-6fb1-4643-93f6-21a0889f1f72?file=Faction+Praxis+Cards.dc.html) (each faction section, light + dark) · [Singularity Mobile exemplar](https://claude.ai/design/p/14f9466e-6fb1-4643-93f6-21a0889f1f72?file=Singularity+Mobile+Card.dc.html) (mobile reflow = kit skin single-column + horizontal score strip + shared widget).

| slug | frame | widget | note |
|---|---|---|---|
| ua | ensō | mandala | |
| snide | evidence-tag (3A) | amp meter | |
| singularity | system slab | decode-strip | widget can refine the existing signal-keys |
| everymen | broadsheet | gearworks | |
| ephemerists | codex | constellation stars | |
| coven | gold/plum chronicle | moon-phases | **recombine**: chronicle *frame* + moon-phases *widget* (the mock bundles balloons with the chronicle and moon-phases with the retired pink card) |
| wow | chronicle recolored to **WOW yellow** | balloons | needs a new manifest; use `--faction-wow-*` tokens (light `#e0a800` / dark `#f5c542`). **Do NOT use gold/plum — that is coven's.** First piece of WOW's kit; rest lands later |
| albescent | na spectrum **+ slow rainbow drift** | inherits na (spectrum-sweep) | one manifest row; render as NA + drift, not a from-scratch skin (ADR-0048, secret-society tell) |

Each faction's vote widget is one shared `VoteUI` (desktop + mobile both use it) reusing `useVote` + `VoteShell` states. Each faction's score stamp uses #820's shared conditional logic, styled in the faction's voice.

## Acceptance criteria

- [ ] All 8 factions render their redesigned desktop card, mobile card, and vote widget — light and dark.
[...]

PR #856, opened 2026-07-20
Praxis-card foundation: split PraxisCard.tsx, scoreStamp surface, factionMarks, font wiring, frame geometry

### LINKED ISSUE #839: Praxis-card foundation: split PraxisCard.tsx, scoreStamp surface, factionMarks, font wiring, frame geometry
## Parent
#818

## What to build

The foundation every faction slice depends on. Five things, one PR, because they all touch the same shared files.

### 1. Split `components/PraxisCard.tsx`

It is 700 lines holding all nine desktop archetypes in one file, which serialises every faction slice in this epic behind one merge queue. Mobile is already split per-faction under `components/praxisCard/mobile/` — mirror that layout exactly (`components/praxisCard/desktop/<Faction>PraxisCard.tsx`). Pure move, no behaviour change, so the slices that follow have **disjoint footprints**.

### 2. `scoreStamp` becomes a manifest surface (ADR-0049)

- Register `scoreStamp` in `factions/manifest.ts` + `SURFACE_KEYS`, dispatched by slug like every other surface, with `Default*` fall-through (ADR-0039).
- Keep `scoreBreakdown()` exactly as it is — it is the shared ADR-0047 row-selection logic and one test keeps all nine honest.
- Delete `PraxisScoreStamp`'s single presentation and `PraxisScoreStrip`. A faction's stamp is **size-agnostic**; mobile reuses it (per the design's mobile guidance) rather than getting an invented horizontal strip.
- Ship the `Default`/`na` stamp here — the design's plain cream sheet, score box + rainbow-clipped total, **not** the current rainbow band frame.

### 3. `components/factionMarks/`

React SVG components, not files in `public/`. An external `.svg` behind an `` cannot read CSS custom properties, so it would reintroduce hardcoded hex and break in dark mode. Seed with the two marks that already exist as assets in the bundle — `Enso` and `Lotus` — every `fill` reading a token. Other factions' marks land with their own slices.

### 4. Fix the font wiring

Straight bugs, not design conflicts:

- `--faction-ua-card-font` resolves to `--font-faction-gilt` → **IM Fell English**. The design says **Cormorant Garamond**, which is already loaded.
- The UA card asks for `'Playfair Display'` and `'Marcellus SC'`. **Neither is loaded anywhere**, so both silently render as generic serif. This is most of why the UA card reads wrong before any layout discussion.
- `--faction-default-card-font` → `--font-display` → **Lora** (a serif) where the design says **Bebas Neue**.
- Audit every `--faction-*-card-font` against the design's stated face; fix mismatches and load any missing family.

### 5. Frame geometry

Adopt the design's frame: **~394px**, `border-radius: 7px`, **2px** faction border, **22px** padding. Today the frame is fluid (`flex: 1 1 280px; minWidth: 280`) and five archetypes (UA, Everymen, Snide, Ephemerists, Singularity) set **no border-radius at all**. Score box radius `6px`, media slot `4px`.

## Not in scope

Metatask presentation and the logged-out/empty states are going back to design. Build the pattern with what we have today; they slot in later.

## Acceptance criteria

- [ ] `PraxisCard.tsx` split per-faction mirroring `praxisCard/mobile/`; no behaviour change.
- [ ] `scoreStamp` registered as a surface with `Default*` fall-through; `scoreBreakdown()` untouched.
- [ ] `PraxisScoreStrip` deleted; mobile reuses the faction stamp.
- [ ] `components/factionMarks/` exists with tokenised `Enso` + `Lotus`.
- [ ] Every `--faction-*-card-font` resolves to the design's stated family, and every family used is actually loaded.
- [ ] Frame geometry matches the design on all nine archetypes.
- [ ] PR body lists deviations and the rule that forced each.

## Blocked by

- #837 — vendor the design sources.
- #838 — WOW/Coven palette swap.

pixieofhugs:
Closes #839

Foundation slice of the praxis-card rebuild (#818). Five structural changes, one PR, because they all land on the same shared files. Six commits, each independently green.

## 1. `PraxisCard.tsx` split per-faction

The nine desktop archetypes now live one per module under `frontend/src/components/praxisCard/desktop/`, mirroring the existing `praxisCard/mobile/` layout exactly. `components/PraxisCard.tsx` keeps only the dispatcher and re-exports `ArchetypeProps`, so the faction manifests' import site stays valid; the manifests themselves now import their own card module, as they already did on mobile. `ChroniclePraxisCard` (shared by Coven + WOW) and `PraxisBody`/`frameBase` got their own modules. Pure move — the only content edits are in later commits.

## 2. `scoreStamp` is a manifest surface (ADR-0049)

- Registered in `FactionManifest` + `SURFACE_KEYS`, dispatched on the task's faction slug with `Default*` fall-through.
- `scoreBreakdown()` moved to `scoreStamp/scoreBreakdown.ts` **unchanged, byte for byte**.
- `PraxisScoreStamp`'s presentation and `PraxisScoreStrip` are deleted. Mobile renders the same `ScoreStamp` component as desktop.
- Ships the `Default`/`na` stamp: score box + rainbow-clipped total. The rainbow band that #821 used as the stamp's frame is gone.
- No per-faction stamps authored — #840/#841/#842 own those, and every non-default slug falls through until then.

## 3. `components/factionMarks/`

`Lotus` (inline React SVG, every fill/stroke from `color`) and `Enso` (see Deviations). Both tokenised, both covered by a test that fails on any hex.

## 4. Font wiring

| token / literal | was | now | why |
|---|---|---|---|
| `--faction-ua-card-font` | `--font-faction-old` to IM Fell English | `--font-faction-gilt` to Cormorant Garamond | design face |
| `--faction-ua-body-font` | *(did not exist)* | `--font-faction-codex-script` to EB Garamond | UA's second face |
| `'Playfair Display'` x5 files | **never loaded**, generic serif | `var(--faction-ua-card-font)` | not a design face |
| `'Marcellus SC'` / `'Marcellus'` x6 files | **never loaded**, generic serif | `var(--faction-ua-body-font)` | not a design face |
| `--faction-default-card-font` | `--font-display` to Lora | `--font-accent` to Bebas Neue | design face |
| `--faction-wow-card-font` | `--font-display` to Lora | `--font-faction-medieval` to MedievalSharp (added to loader) | design face; the old comment deferred this pick to exactly this redesign |

Audited and left alone because they already resolve to a face the design names: Snide (Permanent Marker), Coven (Caveat), Ephemerists (Cinzel), Singularity (Share Tech Mono), Everymen (Bebas Neue).

New guard `src/utils/__tests__/fontsLoaded.test.ts` compares every family named anywhere in `src` against the families `index.html` requests — negative-tested, it catches a reintroduced `'Playfair Display'`. This class of bug is invisible to tsc, eslint and the style ratchet because the fallback *is* the rendering.

## 5. Frame geometry

`frameBase` now carries `flex: 1 1 394px` and `borderRadius: 7`; borders to 2px and padding to `--space-xl` across all nine. Five archetypes (UA, Everymen, Snide, Ephemerists, Singularity) had no radius at all.

## Where the brief disagreed with the code

Reporting rather than splitting the difference:

- **The design bundle is not in the repo.** `.design-sync/praxis-cards/` does not exist on `origin/main` (#837 is still open). I worked from the copy in the additional working directory (`Downloads/Faction Vote Stamps Design (1)/`), which contains the README, the `.dc.html` prototype and both SVGs. `enso.svg` is vendored into `frontend/public/factionMarks/` by this PR because the code needs it at runtime.
- **`--faction-ua-card-font` did not resolve through `--font-faction-gilt`.** It resolved through `--font-faction-old`. `--font-faction-gilt` was *already* Cormorant Garamond — it was simply never pointed at. The bug (IM Fell English) was exactly as described; the mechanism was not.
- **`scoreBreakdown()` had no test.** The brief says its test must still pass untouched; there wasn't one, anywhere. I wrote one (7 cases pinning the ADR-0047 row rules) as part of this slice, since the function is now the epic's shared invariant.
- **The README's frame contract is a generalisation from the UA card.** Per-card, the prototype ranges 380-398px wide, radii 0/7/8/9/10/16, borders 1/1.5/2px, padding 20-22px. Only the UA card is literally 394/7px/2px/22px. I applied the single contract uniformly as the brief settled; the per-card values are there for the faction slices to argue with.

## Deviations

Every place a house rule or a hard constraint overrode a design value:

1. **Enso ships as a `public/` asset painted through a CSS mask, not an inline component.** Design/ADR-0049 said component; the asset is 705 KB / 284 paths in a single colour, and inlining it would put that in the main JS bundle. Colour still comes from a token (`background-color`), so the ADR's actual requirement — tintable, theme-safe, no hex — holds. Agreed with the coordinator mid-task; ADR-0049 is being amended to match. `Lotus` (9 KB, gradient fills a mask cannot carry) stays inline.
2. **Padding 22px to 24px** (`--space-xl`). `no-raw-style-values` forbids raw px and there is no 22px step on the scale.
3. **Default-stamp type sizes snapped to the `--text-*` scale**: design 9 / 10 / 12 / 26 / 30, shipped 9 / 10 / 12 / 24 / 32. Same rule.
4. **394px is a flex basis, not a max-width.** Every surface that renders praxis cards uses a wrapping flex row; a hard cap would maroon a card in a wide column. Per the brief's "not a hard floor".
5. **Multiplier chip is a solid ink chip, not the prototype's green-to-blue gradient with white text.** New colour tokens were explicitly out of scope (#838 owns palette), and white on the rainbow's amber stop fails AA.
6. **Score box sits on the card background, not the prototype's lifted white** (`#fff` on `#fffdf9`). No lift token exists and adding one is a colour change.
7. **2px border applied as width only; no border colour changed.** The design says "2px faction border"; recolouring frames is palette work owned by #838 and the faction slices. Everymen keeps neutral `--color-border`.
8. **Snide gets no border** (design: 1.5px acid). Its frame is a torn scrap built from `clipPath` strips — a rectangular border would draw straight through the tears. Left for the Snide slice.
9. **The `na` stamp nests its total mark inside the score box's border**, rather than as two visually separate objects. That is how the prototype's Unaffiliated card draws it (the UA card is the one that puts its mark outside).
10. **Unloaded families were repointed, not loaded.** Playfair Display and Marcellus SC are not faces the design names for UA; adding two webfonts to keep a wrong pairing alive would be the worse fix.
11. **`DefaultPraxisCard`'s card-level rainbow band is retained.** Only the *stamp's* invented band was in scope; card restyling belongs to the faction slices.

## Verification

- `tsc --noEmit` — clean (the 15 `Cannot find module 'node:fs'/'node:url'` lines are the known stale-junction false alarm from PR #675 and are present on `main` too).
- `eslint src --max-warnings=0` — clean.
- `vitest run` — **105 files / 1101 tests passing**, up from 102 / 1086 (11 new: score breakdown + dispatch, faction marks, fonts-loaded).
- `vite build` — clean; `enso.svg` emits to `dist/factionMarks/` and stays out of the JS bundle.

**Visual QA is outstanding.** I cannot screenshot (the Browser pane renderer times out) and there is no dev stack here, so nothing below has been looked at by an eye.

## What to eyeball

- **The interim stamp fall-through, light + dark.** Every non-`na` faction now renders the unaffiliated cream stamp until #840/#841/#842 land. This is intended by the brief, but it is the loudest change: check Singularity and Snide, whose dark frames now hold a cream box.
- **UA, everywhere** (praxis card, task card, comment, sigil, faction hero, faction body, praxis detail, edit praxis, mobile card): Cormorant Garamond + EB Garamond replacing generic serif. This is the single biggest typographic change in the PR.
- **WOW, everywhere**: MedievalSharp headline replacing Lora — chronicle praxis card, faction card, task card, comments.
- **Unaffiliated + Albescent**: Bebas Neue headline replacing Lora; Albescent additionally under its rainbow drift.
- **Mobile praxis cards, all nine, at 390px**: the horizontal score strip is gone and the boxed stamp took its place — confirm it does not overflow or crowd the vote widget.
- **Frames**: 7px radius is new on UA / Everymen / Snide / Ephemerists / Singularity. Everymen and Snide both carry `clipPath` torn edges that interact with a radius; check the corners.
- **Praxes feed and faction-page card grids at a wide viewport**: the 394px flex basis changes wrapping density from the old 280px.


pixieofhugs:
Reviewed and pushed one fix (`b266978`) before merging — **the flattened frame radius, and it was my brief's fault, not this PR's.**

I told you "~394px, radius 7" as a global. That is the README's *template contract*; the prototype draws a different radius per faction:

| Snide | Everymen | UA | Singularity | Ephemerists | chronicle | Default/Albescent | Coven sticker |
|---|---|---|---|---|---|---|---|
| **0** | 2 | 7 | 8 | 8 | 9 | 10 | **16** |

You flagged this yourself — *"only literally true of the UA card… I applied it uniformly as settled"* — and you were right to. A shared 7px is exactly the flattening this rebuild exists to undo: it erases the difference between an evidence slab with hard corners and a 16px sticker. Radius now lives in each archetype, with Snide carrying a comment so the absence reads as intent rather than an oversight someone later "fixes".

The flex basis stays collapsed at 394px, now recorded as a named deviation (the prototype runs 380–398; under flex the row absorbs it).

Three other things from your report, for the record:

1. **`--faction-ua-card-font` resolving through `--font-faction-old`, not `--font-faction-gilt`** — you're right, my brief had the mechanism wrong. Bug real, cause different. Good catch.
2. **`scoreBreakdown()` had no test.** My "its test must still pass untouched" was protecting nothing. Writing 7 cases was the correct response to a brief that assumed a guard existed.
3. **The `TS7006 implicit any` errors in your two new guard files are the known stale-junction false alarm**, one step downstream — `node:fs` fails to resolve, so `readdirSync` loses its types and its callback param goes implicit-any. Worth knowing the false alarm has a second face beyond `TS2307`.

Verified locally with a real junction (`tsc --version` → 5.9.3, not a phantom): 105 files / 1101 tests pass, eslint clean.

Issue #867, opened 2026-07-21
Mobile Snide + Singularity praxis cards still carry the invented chrome deleted on desktop (and Singularity injects a keyframe inline)


pixieofhugs:
Found while building #842, which removed the invented chrome from the **desktop** Snide and Singularity praxis cards. The **mobile** twins still have it, so desktop and mobile now disagree.

## The gap

`frontend/src/components/praxisCard/mobile/SnideMobilePraxisCard.tsx`
- Still renders the **tape strip** and **torn-paper edge**. Per #842 / ADR-0050 these are inventions with no design counterpart; they were deleted on desktop.

`frontend/src/components/praxisCard/mobile/SingularityMobilePraxisCard.tsx`
- Still renders the **header block cursor** (deleted on desktop as an invention).
- **Injects a stylesheet into the component**: `<style>{'@keyframes blink'}</style>`. This is a real hazard independent of the design question — a component-injected keyframe sits outside the `[data-theme="dark"]` cascade, duplicates on every mount, and is invisible to the `index.css` token discipline. Keyframes belong in `index.css` behind a reduced-motion-gated class, like `.sg-scan` / `.sg-cursor` which #842 added there.

## Why #842 didn't fix it

Mobile praxis cards are a separate design surface (`Singularity Mobile Card.dc.html`, #495) with its own `.dc.html`, and #842's scope was the desktop rebuild. Deleting mobile chrome without reading the mobile design would have been guessing.

## Fix

1. Read the mobile `.dc.html` for both factions and confirm the tape / torn edge / cursor genuinely have no counterpart there either (they may be legitimately drawn on mobile — **check before deleting**).
2. Move the `@keyframes blink` into `index.css` behind a reduced-motion-gated class regardless of the outcome of (1). That part is not design-dependent.

## Related

#842 (desktop rebuild), ADR-0050, #495 (mobile design surface), #818 (epic).

pixieofhugs:
**Read the mobile designs (step 1 of this issue). Both design questions come back the opposite way to what the issue assumes — deleting the tape and the cursor would have been wrong.**

The designs are not at the path this issue cites. There is no `Singularity Mobile Card.dc.html`; the mobile surfaces live in the **World Zero Frontend Kit** project at `mobile-system/factions/WZ Mobile - SNIDE.html` and `WZ Mobile - Singularity.html`.

## SNIDE — the tape is real, the torn edge is not

[...]

Issue #826, opened 2026-07-20
There are many flurs. There should be one flur (light and dark mode)


pixieofhugs:
The fleur-de-lis (`TaskCrown`, ADR-0028) marks the **highest-scoring praxis of any task**. Today it renders **many** ways — a fixed rainbow ring but a per-faction disc **and glyph color** passed by every caller (coven sticker-accent, ua frame, wow gold, snide acid, everymen red, singularity bracket, ephemerists gold, default accent, mobile `theme.accent`). Result: a different-colored fleur on every card (dark-center here, white-center there).

**Design resolved (grill 2026-07-24). One canonical, theme-aware flur, rebuilt around a new mask-cutout construction. Bigger than the old 44px.**
Design ref: `3dd9012c-3b10-4d14-bda6-bbc356fb766d` / `Fleur de Lis.dc.html` (Turn 3, "High Score"). Build agents can't reach it — the full spec is below.

## The construction (replaces the ring+disc+glyph approach)
One continuous **diagonal rainbow** fills the whole disc; a **frame** (ivory in light, ink in dark) is painted over it via an SVG **mask**, and the lily + legend are *cut out* of the frame so the **same rainbow reads through the disc and the lily**. The lily is no longer a colored glyph — it's the rainbow showing through.

**Disc gradient** (repoint `--fdl-rainbow` in `index.css:1380` to this — its only consumer is `TaskCrown`; keep the name):
```
linear-gradient(225deg,#a30f16,#c1272d 10%,#e0431f 20%,#ca8a04 34%,#16a34a 48%,#1d6e72 60%,#2563eb 72%,#6d28d9 84%,#be185d 94%,#8a1046)
```
**Frame token** — add theme-aware `--fdl-frame`: `#fffdf9` (light) / `#14100b` (`[data-theme="dark"]`).

**Markup per instance** (a `.disc` div carrying the gradient, an absolutely-positioned SVG painting the frame rect through a mask):
```html
<div class="disc" style="width:96px;height:96px;border-radius:50%;overflow:hidden;background:var(--fdl-rainbow)">
  <svg width="96" height="96" viewBox="0 0 240 240">
    <defs>
      <mask id="fdl-{uid}">
        <circle cx="120" cy="120" r="110" fill="#fff"/>                         <!-- frame area -->
        <g fill="#000" transform="translate(70,64) scale(1.0)"><use href="#fleur-{uid}"/></g>  <!-- lily cutout -->
[...]

PR #920, opened 2026-07-21
Duel Victor badge art: the Fleur-de-Lis seal (#823)

### LINKED ISSUE #823: Duel-winner badge: 'won a duel last era' (needs era-close freeze)
@
Split from #748 (winner-of-a-duel badge). #748 ships the buildable half (currently-ahead + forfeit wins). This tracks the deferred half: a badge for a **finalized** vote-decided duel win ("won a duel last era").

## Blocked by #824

Unbuildable until a duel has a finalized winner. #719 decided duels **resolve at era close** (honoring ADR-0011); #824 builds that freeze — stamping `winner_character_id` + a final-points snapshot on each duel inside `apply_era_reset`. That frozen winner is this badges source of truth.

Once #824 lands: add a badge whose condition reads a characters frozen duel wins from a *prior* era (via the same batched `build_badge_contexts` path — the first cross-era `BadgeContext` fact). Nothing to show until an era actually closes (era 1 is live).

**Blocked by #824.** Design decision: #719. Parent: #459. Sibling: #748.
@

pixieofhugs:
The frontend half of the Duel Victor badge: the Fleur-de-Lis seal from § Turn 2
of `Fleur de Lis.dc.html` (`2a` light / `2b` dark), registered as `duel_victor`
in `BADGE_ART`.

Closes #823.

## Ordering — merge #918 first

This PR is art only. The backend half (#918) ships the `Badge(key="duel_victor",
name="Duel Victor")` this registry row answers to, and it had not merged when
this was built. The two footprints are disjoint (that PR is 100% backend, this
one is 100% frontend), so they do not conflict — but until #918 lands the
backend never emits the key, and `Closes #823` on this PR would close the issue
with half of it open. **Merge #918 first.**

The key string was read straight out of #918's diff, not guessed.

## The one real decision: the ring legend drops below 72px

The design is drawn at 240px, where `DUEL` / `VICTOR` at 13 units is comfortable.
`BadgeArtProps` defaults `size = 18`, and the five live consumers pass **16 or
18** — `RosterTable` (18), `DefaultProfileBody`, `profileSkin`, mobile
`DefaultProfile`, mobile `DefaultPlayers` (16 each).

The legend's cap height is `size × 13/240`. At 18px that is **0.98px** — not
small text, sub-pixel fuzz around the rim, and it would take the lily's curls
down with it.

**Chosen: degrade to the crowned-lily mark below 72px.** Above it, the design's
own composition renders unchanged, legend and all.

- **Why not keep the legend?** It is a smudge at every size anyone currently
  renders. It would read as noise on the rim, not as lettering.
- **Why not bump the render size?** That means editing `RosterTable.tsx` and
  four profile bodies — feature components, out of a style agent's scope, and
  the wrong lever anyway: it would resize *every* badge to suit one.
- **Why 72?** Readable lettering wants roughly 4px of cap height, which puts the
  floor at `240 × 4/13 ≈ 74`. 72 is the nearest round figure and a clean 0.3× of
  the design's 240.
- **The degraded mark invents nothing.** It is the *same* fleur and crown paths
  under a uniform 1.42 scale, with the crown's exact offset from the lily
  preserved, recentred on the disc to reclaim the space the legend freed. No
  path is redrawn, nothing is added, only the legend is subtracted. (Turn 1 of
  the design is the uncrowned base mark — the crown is asserted present at both
  sizes by a test, so this can never drift into being Turn 1.)

Degrading a spectrum by geometry rather than shrinking it is ADR-0039's own
move: `factionFill` takes a `FactionFillShape` for exactly this reason ("a
7-stop linear on a 10px dot reads as mud"), and the spectrum at 16px already
ships as `--faction-default-ring`.

## The gradient is the na spectrum but is NOT `--faction-default-rainbow`

Worth flagging, because the brief asked me to reuse the existing ramp if one
held these stops. **None does.** The design's ramp is 10 stops on a 225° diagonal
(`#a30f16 … #8a1046`); `--faction-default-rainbow` is 7 stops on a 90° horizontal
and its endpoints differ. Substituting it would have flattened the diagonal
sweep and baked in the wrong angle.

So the stops are minted as `--badge-victor-stop-1..10` in `index.css` — the same
argument the neighbouring `--faction-default-chip` / `--faction-default-total-rainbow`
pair already carries in its own comment: the na spectrum comes in ramps cut for
their geometry.

They are **scalar** tokens rather than a gradient token for a hard technical
reason: SVG `<stop stop-color>` resolves a *colour*, and cannot read a CSS
`linear-gradient()`. The glyph composes its own `<linearGradient>` from them.

## Themes

`--badge-victor-frame` is `#fffdf9` light / `#14100b` dark, flipped purely by the
`[data-theme="dark"]` cascade — no ternary, no `useTheme()`. The ten spectrum
stops deliberately do **not** flip: the seal is a cut-out, the frame is the only
painted part, and brightening the rainbow behind an unlit frame would break the
cut. Same posture as `--spectrum-glow-*` ("identity colours, so
theme-independent").

No hex in the TSX. The `white` / `black` in the `<mask>` are luminance channels,
not palette, and are spelled as keywords precisely so the test can assert
"no hex, unconditionally".

## SVG id collisions

The design's `#fleur`, `#crown`, `#sealRing`, `#duel` are document-global, and
a roster renders one badge per row. **Namespaced per instance** via `useId()`,
following the existing `PointsRoundel` precedent — React's ids carry `:`, which
is legal in an id but not in a URL fragment, so it is stripped. A test renders
two seals in **one tree** and asserts every declared id is unique and every
`url(#…)` / `href="#…"` reference is namespaced.

(Rendering twice into two separate roots restarts React's id counter and proves
nothing — the first draft of that test failed for exactly this reason.)

## Copy: no i18n catalog, deliberately

The issue says badge name/description copy belongs in the i18n catalogs
(ADR-0031). **Badges do not work that way.** They emit `key` + `name`, the `name`
IS the display string and is set in `backend/badges.py`, and nothing in
`frontend/src/locales` references `duelist`. #918 matched that existing
mechanism, and so does this. If badge copy *should* be catalogued that is a
separate change covering all four badges — not built here.

## Tests

`frontend/src/components/badges/__tests__/badgeArt.test.tsx` — 8 tests, the
first badge-art tests in the repo:

- `badgeArtFor('duel_victor')` is not `UnknownBadgeArt` (the silent-failure case:
  a key typo on either side falls through to a perfectly reasonable-looking
  medallion ring)
- `duelist` and `duel_victor` stay two distinct glyphs
- frame + first/last spectrum stop come from tokens; no hex anywhere
- per-instance id namespacing, two seals in one tree
- legend absent at 16 and 18; compact placement used
- legend present at 240, design placement unscaled
- the crown is present at **both** sizes, so the mark can never degrade into
  Turn 1's uncrowned base

## What to eyeball

**⚠️ Visual QA is outstanding and cannot be done in this environment**, and it
cannot be done in the running app either, for a reason worth stating plainly:

> **This badge cannot appear anywhere in World Zero until an era closes.** Era 1
> is live. `duel_victor` is granted only from a *frozen* `Duel.winner_character_id`
> in a **prior** era, so in every environment — including prod — there is no
> character who can be wearing it. The art can only be checked by rendering the
> component directly (a scratch route or a temporary `<DuelVictorArt size={240} />`),
> never by finding it on a profile.

When someone does render it:

1. **At 240px, light and dark.** Against `2a` / `2b`. The lily and crown should
   read as rainbow, the disc around them as ivory (light) / near-black (dark).
   `DUEL` on the upper arc, `VICTOR` on the lower, both Cinzel 800.
2. **At 16 and 18px.** Does the crowned-lily silhouette survive? This is the
   size that actually ships and the one I am least able to verify.
3. **The residual mush risk.** Ten stops across ~16px still averages toward
   brown through the middle. I kept the single ramp rather than inventing a
   short one — inventing geometry the design never drew is what #867 had to
   undo. If it reads muddy at roster size, the fix is a short ramp token, and it
   is a colour decision the owner should make rather than one I should guess.
4. **Gradient direction.** Should run top-right → bottom-left. It is applied over
   the disc's bounding box rather than the full 240 plate, so the sweep spans the
   visible disc corner to corner.
5. **The crown's pearls** at small size — three circles at r=3/3.4 on a 240 box
   are ~0.2px at 16px and may vanish into the band.
6. **Dark mode** via the theme toggle, not a re-render: the frame should be the
   only thing that changes.

## Green locally

`tsc --noEmit` clean on the touched files (the `node:fs` / `node:url` errors are
the known stale-junction false alarm, PR #675). `eslint .` clean. `vite build`
succeeds. Full suite **113 files / 1275 tests** pass.


pixieofhugs:
Design update (2026-07-24): the Duel-Victor Fleur-de-Lis is now part of a unified fleur family with the High-Score crown — same mask-cutout construction (diagonal rainbow disc + ivory/ink frame + lily cut out so the rainbow reads through), Turn 2 in design `3dd9012c-3b10-4d14-bda6-bbc356fb766d` / `Fleur de Lis.dc.html`. Duel Victor = base lily **+ a crown glyph** above it **+ a 'DUEL VICTOR' legend** (on the wider `#sealRing` arcs). #826 builds a shared `FleurSeal` primitive (props: legendTop/legendBottom, crown?, size); this badge should **consume that primitive** rather than rebuild the art. Blocked-by/after #826 for the primitive.

PR #936, opened 2026-07-21
fix(praxis): score stamp's votes row goes stale after voting (#912)

### LINKED ISSUE #912: fix(praxis): score stamp's votes row and total go stale after voting (override merges score, not the breakdown fields)
## Problem

Cast or change a vote on a praxis card. The vote control updates, the footer score updates — but the **score stamp's votes row and total do not move** until the page refetches.

## Root cause

`components/vote/useVotedPraxis.ts` merges the client-side vote override into the praxis object, and it patches exactly two fields:

```ts
return {
  ...praxis,
  score: praxis.score + delta.score,
  voter_count: praxis.voter_count + delta.voters,
}
```

The score stamp does not read either of them. `scoreStamp/scoreBreakdown.ts:32` reads the **#819 breakdown fields**:

- `base_points`
- `display_multiplier`
- `metatask_points`
- `points_from_votes`  ← moves on every vote, never merged
- `total`              ← moves on every vote, never merged

So the stamp renders server truth from the last fetch while everything beside it renders the override. `useVotedPraxis`'s own docstring still describes the pre-stamp world — *"PraxisScoreHero's `{base} + {votes}`"* — which is the component the stamp replaced. The override store (#626) predates the stamp (#819/#820, ADR-0047) and was never extended to it.

**Units line up, no conversion needed.** `points_from_votes` is `SUM(Vote.value)` — the raw star sum (`backend/services/vote_tally.py:46`). `delta.score` is `mine - (prior ?? 0)` (`voteOverrides.ts:103`), the same unit. And `total = base × mult + meta + votes` (`praxis_scoring.py:39`), so votes enter the total additively and un-multiplied — the same delta applies to both.

## Fix

In `useVotedPraxis`, add the delta to `points_from_votes` and `total` alongside the existing two fields. Update the stale docstring to name the score stamp instead of `PraxisScoreHero`.

That is the whole fix. Both card dispatchers (`PraxisCard.tsx:34`, `praxisCard/mobile/MobilePraxisCard.tsx:13`) already call this hook before choosing a faction skin, so all nine stamp variants are fixed at once — no per-faction work, no changes to `scoreBreakdown` or any `*ScoreStamp.tsx`.

## Explicitly NOT in scope

- **The praxis detail page.** It has its own merge (`usePraxisDetail.ts:303`) into the `votes` object, which is already correct. Do not touch it.
- **`base_points`, `metatask_points`, `display_multiplier`.** A vote cannot move any of them. Merging them would be wrong.
- Any change to the override store, `scoreBreakdown`, or the stamp components.

## Acceptance

- Voting or re-voting on a praxis card moves the stamp's votes row and total immediately, with no refetch.
- The moved numbers match what the server returns on the next fetch (no double-count when the override clears — the existing `clearVoteOverrides` path already handles retirement).
- A test asserting the stamp's votes row and total reflect an active override. `components/praxisCard/scoreStamp/__tests__/` and the existing `__resetVoteOverrides()` test seam are the places to put it.

pixieofhugs:
## Problem

Cast or change a vote on a praxis card: the vote control and the footer score update, but the **score stamp's votes row stayed on last-fetch server truth** until a refetch.

## What actually changed since the issue was written

ADR-0053 (PR #902) renamed the stamp's score fields. The stamp no longer reads `base_points`/`total`; it reads `points_from_votes` (the votes row) and `score` (the total) via `scoreStamp/scoreBreakdown.ts`. `useVotedPraxis` was **already** merging `score`, so the **total already moved live** — the issue's "total goes stale" claim had rotted. The one remaining bug was the **votes row**: `points_from_votes` was never merged.

## Fix

One line in `useVotedPraxis`:

```ts
points_from_votes: praxis.points_from_votes + delta.score,
```

Units match with no conversion: `delta.score` is `mine - (prior ?? 0)` (raw star-sum delta, `voteOverrides.ts`) and `points_from_votes` is `SUM(Vote.value)` — the same raw star sum. `score` was left as-is (already merged); base / multiplier / metatask (`task_point_value`, `display_multiplier`, `metatask_points`) are untouched — a vote cannot move any of them. Both card dispatchers already call this hook before choosing a faction skin, so all nine stamp variants are fixed at once. The docstring already named the score stamp correctly (no `PraxisScoreHero`), so it was left alone.

## Test

`scoreStamp/__tests__/votedStamp.test.tsx` pins the observable contract at the store-delta + stamp seam: an active override moves **both** the votes row and the total by the same delta, base is untouched, and clearing the override returns to server truth with no double-count.

Harness caveat: the hook reads the override through `useSyncExternalStore`, whose `getServerSnapshot` returns `null` under `renderToStaticMarkup` (nobody votes mid-SSR). I verified empirically that rendering the hook directly can never observe an override in this DOM-less harness, so the test drives the same shared `deltaFrom` the hook's `useVoteOverride` runs, merged exactly as `useVotedPraxis` does, rather than through a live hook render.

## Verification

- `tsc --noEmit`: clean for both changed files (the only errors are the known stale-junction `node:fs`/`node:url` false alarms from PR #675, in pre-existing test files I never touched).
- `eslint` (changed files): clean.
- `vite build`: passes.
- `vitest`: 1339 passed (115 files), including the 3 new cases.
- pytest: N/A (frontend only).
- **Visual QA is outstanding** — no dev stack / no screenshots available in this environment.

## What to eyeball

- Vote on a praxis **feed card**: the stamp's `+ N from votes` row and the total both jump instantly, in step with the footer score.
- **Re-vote** (change your stars) before a refetch: both numbers move by the change, no double-count.
- On the next fetch that returns the same praxis, the stamp matches server truth (the override retires cleanly — no stacked delta).
- Confirm across a few faction skins (Everymen, WOW, Coven, UA, S.N.I.D.E., Singularity, unaffiliated) since all nine share the one hook.

Closes #912

### DIFF
--- a/frontend/src/components/praxisCard/scoreStamp/__tests__/votedStamp.test.tsx
+++ b/frontend/src/components/praxisCard/scoreStamp/__tests__/votedStamp.test.tsx
@@ -1,0 +1,94 @@
+/**
+ * The score stamp's votes row and total track the viewer's own just-cast vote
+ * (#912).
+ *
+ * `useVotedPraxis` merges a vote override into BOTH the total (`score`) and the
+ * votes row (`points_from_votes`) before a card picks its faction skin, so the
+ * stamp's breakdown moves with the footer instead of lagging until the next
+ * fetch. Before #912 only `score` was merged, so the stamp rendered last-fetch
+ * server truth for its votes row while everything beside it showed the override.
+ *
+ * Harness note: the hook reads the override through `useSyncExternalStore`,
+ * whose server snapshot is `null` under `renderToStaticMarkup` (nobody votes
+ * mid-SSR), so rendering the hook directly can never observe an override in this
+ * DOM-less harness. These cases pin the same seam the hook composes instead:
+ * take the store's delta (the shared `deltaFrom` the hook's `useVoteOverride`
+ * also runs), merge it exactly as `useVotedPraxis` does, and assert the stamp.
+ */
+import { describe, it, expect, afterEach } from 'vitest'
+import { renderToStaticMarkup } from 'react-dom/server'
+import '../../../../i18n'
+import type { PraxisCardOut } from '../../../../api/praxis'
+import { recordVote, tallyDelta, __resetVoteOverrides } from '../../../vote/voteOverrides'
+import DefaultScoreStamp from '../DefaultScoreStamp'
+
+const PRAXIS_ID = 7
+
+function praxis(overrides: Record<string, unknown> = {}): PraxisCardOut {
+  return {
+    id: PRAXIS_ID,
+    task_point_value: 12,
+    display_multiplier: 1,
+    metatask_points: 0,
+    points_from_votes: 4,
+    score: 16,
+    voter_count: 3,
+    ...overrides,
+  } as PraxisCardOut
+}
+
+/** The merge `useVotedPraxis` performs — kept byte-for-byte alongside it. */
+function votedPraxis(base: PraxisCardOut): PraxisCardOut {
+  const delta = tallyDelta(base.id)
+  if (!delta) return base
+  return {
+    ...base,
+    score: base.score + delta.score,
+    points_from_votes: base.points_from_votes + delta.score,
+    voter_count: base.voter_count + delta.voters,
+  }
+}
+
+const text = (html: string) => html.replace(/<[^>]*>/g, '')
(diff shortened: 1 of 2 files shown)


pixieofhugs:
Superseded: #912 was already fixed on main by #939 (commit 28d8f75d), which patches `points_from_votes` via the `applyVoteDelta` extraction — the same fix this PR made inline. No longer needed.

Issue #1022, opened 2026-07-28
Task cards v2 · F2 — card contract (base + multiplier + in-progress), mobile unification, na/Default card


pixieofhugs:
## Parent
Epic #1020 · governed by ADR-0055 and ADR-0056

## What to build
The foundation the 8 bespoke cards build on, plus the na/Unaffiliated card itself (it's the fallback and the same file, so it rides here).

**Card contract (ADR-0055).** Change `CardProps` to carry base points and the multiplier factor separately instead of the pre-multiplied product: `{ task, basePoints, multiplier, inProgressCount, onSignup }`. Add a `displayMultiplierFor(task)` helper beside `displayPointsFor` that returns the raw `own_task_modifier`/`other_task_modifier` factor for the viewer (the same factor `computeDisplayPoints` applies) — the raw factor, never `effective / base`. `basePoints` is `task.point_value`. Update every `<TaskCard>` call site (`Tasks.tsx`, `Home.tsx`) to pass the new props. Cards show base points, plus a `×multiplier` badge **only when the factor ≠ 1.0** (so it's absent in era_1).

**Mobile unification (ADR-0056 — reversible experiment).** Rewire the mobile task-browse list (`DefaultTasks.tsx`) to render the shared `<TaskCard>` with the `onSignup`/`signupMsg` state `useTasks()` already exposes, so mobile gains the inline CTA + in-progress + multiplier. **Do NOT delete** the `mobileTaskCard` surface, its dispatcher, `DefaultMobileTaskCard`, or the 9 mobile card files — leave them in place but dormant so the experiment reverts by flipping this one call site back. Deletion is a separate, deferred cleanup gated on the owner's verdict.

**na/Default card.** Rebuild `DefaultTaskCard` as the responsive na/Unaffiliated card to the "Unaffiliated" design (`default-task-card.jsx` in the design project — na ≡ Default is one identity). It serves both form factors via `useFormFactor()` and is the fallback for any faction whose bespoke card hasn't landed.

**Build translation.** Design `theme`/`platform` props → `[data-theme]` cascade + `useFormFactor()`; hardcoded hex → `--faction-*`/theme tokens in `index.css`; ordinal → "Task {task.id}"; in-progress label uniform; reuse the existing na signup i18n key.

## Acceptance criteria
- [ ] `CardProps` is `{ task, basePoints, multiplier, inProgressCount, onSignup }`; `displayPoints` is no longer threaded into cards.
[...]

PR #1025, opened 2026-07-28
#1021: add in_progress_count to TaskOut

### LINKED ISSUE #1021: Task cards v2 · F1 — add in_progress_count to TaskOut
## Parent
Epic #1020

## What to build
Surface a count of how many characters are actively working on a task, so the v2 task cards can show "{n} in progress". Add `in_progress_count` to the task read payload (`TaskOut`), computed from the task's active signups — the same population `GET /tasks/{id}/signups` already exposes, reduced to a count. Thread it through the list and detail reads so every task in the feed carries it.

No new table and no migration — this is a derived read-time count. Keep it efficient for the list endpoint (avoid an N+1 across the feed; a grouped count keyed by task id is fine).

Also add the field to the frontend `TaskOut` TypeScript interface so F2 can consume it.

## Acceptance criteria
- [ ] `TaskOut` (backend schema + frontend interface) carries `in_progress_count: int`.
- [ ] The value equals the number of active/in-progress signups for the task (consistent with the signups endpoint's active set).
- [ ] Populated on both the list read (`GET /tasks`) and the single read (`GET /tasks/{id}`).
- [ ] The list endpoint does not regress into a per-task query (grouped/bulk count).
- [ ] Backend test asserts the count reflects active signups (and excludes non-active states).

## Blocked by
None - can start immediately

pixieofhugs:
## Summary

Adds a derived, read-time `in_progress_count: int` field to `TaskOut`, counting characters actively working on a task, for the upcoming v2 task cards ("{n} in progress").

- `backend/schemas/task.py` — new `in_progress_count: int = 0` field on `TaskOut`. Defaults to 0 because `TaskOut.model_validate()` is also used directly on metatask rows (the praxis seal stack in `services/praxis.py`'s `applied_metatasks_for`), which have no such attribute on the `Task` ORM object — 0 is also the *true* value there, since metatasks can never be signup targets (`evaluate_signup` rejects `TaskType.metatask` before any status check).
- `backend/services/task.py` — new `in_progress_counts_for_tasks(task_ids, session) -> dict[int, int]`: one grouped query (`GROUP BY Praxis.task_id`, filtered to `Praxis.status IN (in_progress, pending)`, joined through `PraxisMember`) mirroring the exact population `list_signups_for_task` already lists for a single task. `build_task_out` and `build_task_out_for_viewer` became `async` and now accept an optional `in_progress_count` — when omitted they self-compute via the same helper scoped to one task (mirrors the existing `crowned_ids`/`applied_metatasks` optional-precompute convention already used by `build_praxis_out`/`build_praxis_card_out`).
- `backend/routers/tasks.py` — `GET /tasks` now does exactly one bulk `in_progress_counts_for_tasks` call for the whole page, then passes each task's count into its `build_task_out_for_viewer` call (no per-task query). `GET /tasks/{id}` leaves it to self-compute (single task, one extra query — same tradeoff the codebase already accepts for `build_praxis_out`'s single-item routes).
- `backend/routers/admin.py` — every existing `build_task_out(task)` call site updated to `await build_task_out(task, session)` (signature change ripple). `GET /admin/tasks/pending` also bulk-precomputes counts for its list rather than defaulting pending tasks to 0 — pending tasks are *usually* signup-free, but `era.allow_praxis_on_pending_task_factions` can carve out a faction allowed to act on a still-pending task, so this reads the real count rather than assuming.
- `frontend/src/api/tasks.ts` — added `in_progress_count: number` to the `TaskOut` interface (typed only; not consumed anywhere yet — that's #1022).

## Counting approach (for review)

```sql
SELECT praxis.task_id, count(praxis_member.id)
FROM praxis JOIN praxis_member ON praxis_member.praxis_id = praxis.id
WHERE praxis.task_id IN (...) AND praxis.status IN ('in_progress', 'pending')
GROUP BY praxis.task_id
```

One row per `PraxisMember` — i.e. one row per character on a praxis — which is what makes it "the same population `GET /tasks/{id}/signups` already exposes, reduced to a count": that endpoint's `list_signups_for_task` already does the identical join+filter, just without the `GROUP BY`/`count`. `PraxisStatus` only has three values (`in_progress`, `pending`, `submitted` — no separate "approved" state on `Praxis`; scoring happens via votes on a `submitted` praxis, not a further status transition), so `submitted` is the one state a correctness test needs to prove excluded.

## Test DB

**I could not run the backend test suite locally.** Per the task's hazard #5 I set up a dedicated `wz1021_test` DB (`TEST_DATABASE_URL=postgresql+asyncpg://worldzero:localdev@localhost:5432/wz1021_test`) and copied `backend/.env` into the worktree, but Postgres (via `docker-compose` / Docker Desktop) could not be brought up in this sandbox: Docker Desktop's backend crashes on startup (`starting services: initializing Inference manager: listening on unix://...dockerInference: remove ...: The file cannot be accessed by the system.`), and I traced this to an OS-level failure deleting/recreating its AF_UNIX-reparse-point runtime socket files — reproducible directly via `Remove-Item`/`[System.IO.File]::Delete` on those files outside Docker entirely, so it isn't a Docker-config issue I can work around. No native Postgres install was available as a fallback.

In lieu of a live pytest run I verified as much as possible statically with the venv's Python: `py_compile` on every touched file, a full `import main` (catches import-time/circular-import errors), `app.openapi()` (validates all route/response-model wiring, confirms `in_progress_count` appears correctly on both `TaskOut` and `PendingTaskOut`), and compiling the new grouped-count query via SQLAlchemy's `.compile(compile_kwargs={"literal_binds": True})` to confirm it produces the intended SQL (shown above). All passed. The new tests are written and committed (`backend/tests/integration/test_tasks.py`) but are **unverified against a live DB** — please rely on this PR's CI `test` job (which provisions its own Postgres service container and isn't subject to this sandbox's Docker limitation) as the authoritative signal, and take a close look at the new tests in review since I couldn't watch them actually pass.

Tests added:
- `test_get_task_in_progress_count_counts_active_signups_only` — 3 characters at `in_progress`/`pending`/`submitted` on one task; asserts `GET /tasks/{id}` reads back 2 (submitted excluded).
- `test_get_task_in_progress_count_zero_with_no_signups` — fresh task reads 0.
- `test_list_tasks_in_progress_count_grouped_per_task` — two tasks with different signup mixes on one `GET /tasks` page, asserting each task's count is independently correct (exercises the grouped/bulk path).
- `test_get_task_response_fields` — extended the existing required-fields assertion to include `in_progress_count`.

Closes #1021

### DIFF
--- a/backend/routers/admin.py
+++ b/backend/routers/admin.py
@@ -40,7 +40,7 @@ from schemas.comment import CommentModerationIn, CommentOut
 from services.praxis import build_praxis_out, moderate_praxis
 from services.vote_tally import crowned_praxis_ids
 from services.comment import build_comment_out, list_flagged_comments, moderate_comment
-from services.task import build_task_out
+from services.task import build_task_out, in_progress_counts_for_tasks
 from services.admin_service import (
     admin_create_character,
     admin_edit_task,
@@ -242,7 +242,7 @@ async def admin_reactivate_task(
     session: AsyncSession = Depends(get_db),
 ) -> TaskOut:
     task = await reactivate_task(task_id, session)
-    return build_task_out(task)
+    return await build_task_out(task, session)
 
 
 # ---------------------------------------------------------------------------
@@ -383,7 +383,7 @@ async def admin_patch_task(
     session: AsyncSession = Depends(get_db),
 ):
     task = await admin_edit_task(task_id, data, session)
-    return build_task_out(task)
+    return await build_task_out(task, session)
 
 
 @router.put("/tasks/{task_id}/status", response_model=TaskOut)
@@ -395,7 +395,7 @@ async def admin_update_task_status(
 ):
     new_status = TaskStatus(data.status)
     task = await update_task_status(task_id, new_status, session)
-    return build_task_out(task)
+    return await build_task_out(task, session)
 
 
 # ---------------------------------------------------------------------------
@@ -413,6 +413,13 @@ async def list_pending_tasks(
     session: AsyncSession = Depends(get_db),
 ):
     rows = await list_pending_tasks_with_proposer(session)
+    # One grouped query for the whole page (#1021) — never a per-task count.
+    # (Pending tasks are ordinarily signup-free, but era config can carve out
+    # factions allowed to act on a still-pending task — era.allow_praxis_on_-
+    # pending_task_factions — so this reads the real count, not an assumed 0.)
+    in_progress_counts = await in_progress_counts_for_tasks(
+        [task.id for task, _ in rows], session
+    )
     return [
         PendingTaskOut(
             id=task.id,
@@ -427,6 +434,7 @@ async def list_pending_tasks(
             metatask_faction_slug=task.metatask_faction_slug,
             is_task_vision_eligible=task.is_task_vision_eligible,
             created_at=task.created_at,
+            in_progress_count=in_progress_counts.get(task.id, 0),
             created_by_name=display_name or "",
         )
         for task, display_name in rows
@@ -440,7 +448,7 @@ async def approve_task(
     session: AsyncSession = Depends(get_db),
 ):
     task = await update_task_status(task_id, TaskStatus.active, session)
-    return build_task_out(task)
+    return await build_task_out(task, session)
 
 
 @router.put("/tasks/{task_id}/retire", response_model=TaskOut)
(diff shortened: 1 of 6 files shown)


pixieofhugs:
Orchestrator note — one correction to the PR body above.

The Summary says the frontend interface got `in_progress_count: number`, but the final commit (`5f2c6606`) made it **optional** — `in_progress_count?: number` — to avoid a mechanical update of every pre-existing `TaskOut` test fixture. The inline comment in `frontend/src/api/tasks.ts` documents this correctly ("Always present on a live backend response… Consumers should treat an absent value as 0"); only the PR body is stale.

Calling it out because it changes what #1022 has to do: the F2 card contract must read it as `task.in_progress_count ?? 0` rather than assuming a number. I'll put that in #1022's brief.

The backend field is genuinely non-optional (`in_progress_count: int = 0`), so the API always emits it — the TS optionality is about test fixtures, not about the wire format.

Issue #1030, opened 2026-07-28
Task details v2 · F2 — shared detail contract, neutral copy, mobile unification, na/Default detail


pixieofhugs:
Part of #1028. **Blocked by #1029.** Blocks C1–C8. One PR, no faction skins — this establishes the contract every skin then copies.

## 1 · Neutral copy block

Add a `detail.*` block to `frontend/src/locales/en/tasks.json` covering the whole anatomy: breadcrumb, faction line, author line (`author · lvl {{level}}`), the three header stats, base / multiplier / total labels, sign-up + slots + level-met, in-progress + continue + drop, submitted, brief heading, roster heading, gallery heading + sort tabs + view-all + empty, comments heading + composer.

Per ADR-0057 (#1029), where the designs disagree, **the domain noun wins**: *praxis* not "submissions"/"accounts"/"reports", *score* not "loved"/"canonical"/"credence"/"witnessed", *sign up* not "acknowledge"/"take the job". Existing `default.*` keys are the closest starting point — several are already neutral and can move across.

**Delete nothing from `tasks.json` in this PR.** The `coven.*` / `snide.*` / `ua.*` / `ephemerists.*` / `everymen.*` / `singularity.*` namespaces are **also consumed by `pages/factionDetail/archetypes/*FactionBody.tsx`, `pages/factionDetail/mobileArchetypes/*FactionPage.tsx`, and `CovenMobileTaskCard.tsx`** — a namespace-level sweep takes the faction pages down with it. Cleanup is Z, key-by-key, after every skin has landed.

## 2 · State contract

`pages/taskDetail/useTaskDetail.ts` currently exposes `factionMultiplier` and the pre-multiplied `modifiedPoints`. The designs draw base and multiplier as separate legible pieces, same as ADR-0055 did for cards. Expose base points and the **raw** factor separately; do not reconstruct a ratio from `effective / base` (the ADR-0053 dead-arithmetic trap). Surface the new author fields and roster levels from #1029.

The multiplier badge renders **only when the factor ≠ 1.0**, so it is invisible at `era_1` across every skin — same as the cards. That is correct, not a bug.

## 3 · Comments

`components/comments/CommentThread.tsx` renders its own `<h3 className="eyebrow">{t('comments.heading')}</h3>`. Archetypes are about to add their own dressed section head, so **add `showHeading?: boolean` (default true)** or every skin gets two headings. Rows and composer are untouched — ADR-0006 stands.

[...]

Issue #1028, opened 2026-07-28
Epic: Faction task details v2 — shared anatomy, neutral copy, 9 responsive skins


pixieofhugs:
Design project `0711d3a7-0074-4a60-8907-270f44168261` (9 `* Task Detail.dc.html` + matching `.jsx`) redesigns every faction's task-detail page. Sibling of #1020 (task cards v2), same shape: one backend read, one contract PR, then one issue per faction skin.

Grilled 2026-07-27. Decisions below are settled — build to them, don't re-litigate.

## What the designs are

Every one of the 9 files carries the same header comment: **one shared anatomy, faction-specific dress only.** Snide's says it outright:

> `// Dress is faction-specific, copy is NOT (task pages carry no faction-specific language).`

Anatomy: header (breadcrumb · faction line · title · author row · Level / In progress / Completed) beside one **action panel** (base points + `×multiplier` + total, and the sign-up / in-progress state block) → full brief → in-progress roster → praxis gallery → comments.

## Settled decisions

1. **One shared neutral copy set.** A single `detail.*` block in `tasks.json`; every archetype reads the same keys. The designs disagree with each other on the neutral wording (`Task Description` vs `The task`; `19 completed praxis` vs `21 submissions`; `Discussion` vs `Comments`) — where they differ, **domain nouns from `CONTEXT.md` win over design wording**: praxis, score, sign up. Dress is the design's authority; copy is not. → ADR-0057 (written in F1).
2. **The author row ships.** All 9 designs draw avatar + name + `author · lvl N` linking to the profile. `TaskOut.created_by` is a bare int with zero frontend uses today. Yes, `seed.py` makes that read "Pixie" on nearly every seeded task — accepted; it is true, and it finally gives `ProposeTask` a payoff surface. → F1.
3. **The in-progress roster is a required slot on every skin.** Snide's design drops it (`// NOTE: no in-progress roster section — the header count is enough`) — that loses. The roster is the only place on a task page you learn a **foe** is working the same task, and shipped `SnideTaskDetail` already has `snide.accomplices`. Each skin draws it in its own dress; the deviation gets named in the PR body per `docs/agents/design-fidelity.md`.
[...]

Issue #1039, opened 2026-07-28
Task details v2 · Z — key sweep, unvendor, and the ADR-0056 verdict


pixieofhugs:
Part of #1028. **Blocked by #1031, #1032, #1033, #1034, #1035, #1036, #1037, #1038** — runs last, after every skin has landed.

## 1 · Sweep dead i18n keys, key by key

The 7 rebuilt archetypes leave ~180 orphaned keys in `frontend/src/locales/en/tasks.json`.

**Do not delete by namespace.** `coven.*`, `snide.*`, `ua.*`, `ephemerists.*`, `everymen.*` and `singularity.*` are **also consumed by** `pages/factionDetail/archetypes/*FactionBody.tsx` and `pages/factionDetail/mobileArchetypes/*FactionPage.tsx`, and `coven.*` additionally by `pages/tasks/mobileArchetypes/cards/CovenMobileTaskCard.tsx`. A namespace sweep takes the faction pages down with it.

Delete only keys with **zero remaining references** across `frontend/src`, verified per key by script, not by eye. Include the dormant `pages/taskDetail/mobileArchetypes/*` files in the reference scan — they are dormant, not deleted, and must keep compiling so the ADR-0056 revert stays a one-line change.

Note that the mobile archetypes still referencing a key is *itself* the reason to keep it, so expect the sweep to remove less than the raw orphan count suggests. That is correct.

## 2 · Unvendor

Delete `.design-sync/task-details-v2/` (`docs/agents/design-fidelity.md`: the epic's last PR removes it).

## 3 · The ADR-0056 verdict

By the time this runs, the owner will have lived with unified task cards (#1022) and unified task details. ADR-0056 is still **Proposed**.

- If accepted: flip it to Accepted and file (do not perform) the follow-up that retires the dormant `mobileTaskCard` and `mobileTaskDetail` surfaces and their ~17 files.
- If rejected: flip the two call sites back — `DefaultTasks.tsx` for cards, `TaskDetail.tsx` for details — and record the outcome in the ADR.

**Ask the owner before doing either.** Do not infer the verdict from silence, and do not delete the dormant files in this issue under any outcome.

pixieofhugs:
Heads-up before anyone starts this: **the ADR-0056 verdict portion of this issue is already done.** Don't redo it.

[...]

PR #1045, opened 2026-07-28
#1029: task author + signup level on the read API, ADR-0057, ADR-0056 amendment

### LINKED ISSUE #1029: Task details v2 · F1 — task author + signup level on the read API, and the epic's docs
Part of #1028. **Blocks every other issue in the epic** — merge this first.

Two unrelated-but-co-shipped halves: a small backend read, and the epic's decision records (put here so they land before any skin is built).

## 1 · Backend read

All 9 task-detail designs draw an author row (avatar, name, `author · lvl 4`, links to the profile) and the Unaffiliated design's roster rows show `lvl N`. Neither is available today.

**`TaskOut`** (`backend/schemas/task.py:9`) — `created_by: int` is a bare id with **zero** frontend consumers (`grep created_by frontend/src` finds only `created_by_id` on praxis). Add, mirroring `PraxisOut`'s existing naming:

- `created_by_display_name: str`
- `created_by_avatar_url: str`
- `created_by_faction_slug: str | None`
- `created_by_level: int`

**`TaskSignupOut`** (`backend/schemas/task.py:62`) — add `level: int`.

Populate in `services/task.py`'s `build_task_out(_for_viewer)` alongside `in_progress_count` (#1021 set the pattern). **`TaskOut` is returned by the browse list too** — eager-load the author relationship (one join); do not let a 50-task list fire 50 author queries. A test asserting the query count on `GET /tasks` is the point of this issue, not a nice-to-have.

Mirror the `in_progress_count` precedent for optionality on the frontend type (`frontend/src/api/tasks.ts`) so pre-existing test fixtures stay valid.

Note for the record: `backend/seed.py:145` seeds every era task with Pixie's character, so on a live board this row reads "Pixie" on nearly every task. That is expected and accepted (epic decision 2) — do **not** add an `is_seeded` / `source` concept to hide it.

### ADR-0057 — Task detail pages carry no faction voice

Write it. Status Accepted, dated 2026-07-27, relates to #1028, tension with ADR-0016.

- **Context:** the 7 shipped task-detail archetypes each invented a vocabulary — `snide.caseFile`, `coven.spellsHeading`, `ephemerists.credenceHeading`, `ua.salonWallHeading`, `singularity.consensusHeading`, `everymen.verdictHeading`. The v2 designs deliberately do the opposite; Snide's file says `Dress is faction-specific, copy is NOT (task pages carry no faction-specific language)`. The designs nonetheless disagree with *each other* on the neutral wording, so building them verbatim would re-fragment the copy into nine near-synonym sets.
- **Decision:** one shared `detail.*` copy set on this surface. Where a design's word differs from the domain noun in `CONTEXT.md`, the domain noun wins. Faction voice remains law on every *other* surface — praxis cards, faction pages, duel skins, collab composers, comment voices are untouched.
- **Consequences:** a task page reads identically across factions and differs only in dress; ~8 faction vocabularies get deleted (see Z); a future faction skin cannot introduce copy, only dress.
- **Alternative rejected:** keep each design's own neutral-ish wording per faction — bland words, same fragmentation, same catalog size.

### ADR-0056 — amend

Add task detail to the collapse. Same experiment, same posture, same one-line revert (`TaskDetail.tsx`'s `formFactor === "mobile"` branch instead of `DefaultTasks.tsx`'s call site); `pages/taskDetail/mobileArchetypes/*` and the `mobileTaskDetail` manifest rows stay registered but dormant. Do **not** write a second ADR — if the owner rejects the unified cards, both surfaces flip back together.

### CONTEXT.md — three edits

**(a)** The **Alias slug** entry is factually wrong. It claims `albescent` and `aged_out` "both render as `ua`", "Encoded in `SLUG_ALIASES` in `utils/factionDispatch.ts`". `SLUG_ALIASES` does not exist anywhere in the repo, `FACTION_ALIASES` is empty, `aged_out` was retired with its faction (#428), and ADR-0046 §3 explicitly refuses an Albescent alias. Replace with:

```
**Alias slug** *(retired — no aliases remain)*:
A slug that inherited another faction's archetype by design rather than rebrand. Only two
ever existed — `albescent → ua` and `aged_out → ua` — and both are retired (#232, #428).
`FACTION_ALIASES` is empty and `pickVariant`'s alias step is a dormant seam.
_Avoid_: describing Albescent as an alias. Falling through to the Default archetype on an
undressed surface is the ordinary partial-registry behaviour every faction has, not aliasing.
```

**(b)** Add an **Albescent** entry, immediately after **Default archetype**:

```
**Albescent**:
A first-class faction — own slug, members, roster, invitation flow and level-8 unlock —
whose *look* is deliberately the Unaffiliated one with extra frills: quiet tells (a slow
rainbow drift, a morphing vote blob) layered over the Default kit's structure, so a member
reads as unaffiliated at a glance and reveals itself only to someone already looking
(ADR-0027 / ADR-0048). Where a design exists, Albescent wears it; where none does yet, it
falls through to Default like any undressed faction. Its faction page stays hidden until an
account has been revealed to the society.
_Avoid_: "Albescent is na", "Albescent is an alias" — below the skin it is its own faction.
```

**(c)** The **Archetype** entry defines an archetype as "its whole shape, layout, ornament, **and copy voice**". As of ADR-0057 that is no longer true on task detail. Add one clause naming the carve-out — don't rewrite the entry.

pixieofhugs:
Closes #1029

Part of #1028 — this is the blocking foundation issue: the read API every task-detail skin needs, plus the epic's decision records.

## 1 · Backend read

**`TaskOut`** gains the denormalised proposer byline, named after the existing `PraxisOut`/`PraxisCardOut` precedent:

- `created_by_display_name`, `created_by_avatar_url`, `created_by_faction_slug`, `created_by_level`

`created_by` stays the bare id (that's what a profile link needs). `created_by_faction_slug` is the author's **member** faction, not the task's `primary_faction_slug` — a `ua` member proposing an `na` task bylines as `ua` (there's a test for exactly that).

**`TaskSignupOut`** gains `level`, and the roster route actually emits it — `GET /tasks/{id}/signups` returns hand-built dicts (it does not use `TaskSignupOut` as its `response_model`), so the schema field alone would have shipped nothing. The level comes from an extra outer join in `list_signups_for_task`'s existing query, so the roster is still one query.

### The N+1

`TaskOut` is what the browse LIST returns, so the author read is precomputed page-wide by a new `services.task.authors_for_tasks(tasks, session)` — one join over `Character` LEFT OUTER JOIN `character_stats`, keyed by **character id** (authors repeat; `seed.py` proposes every era task as one character, so a whole page of seeded tasks collapses to one map entry). `build_task_out(_for_viewer)` take an optional `author: TaskAuthor` exactly as they take `in_progress_count` (#1021's pattern), and self-resolve for single-task callers.

Why a helper and not a `selectinload` of a `Task.author` relationship: `created_by_level` lives on `CharacterStats` for the **current era** (ADR-0042), so an ORM eager-load of `Character` alone could not supply it. On an unseeded DB every level falls back to 0 rather than raising.

`test_list_tasks_author_read_is_not_an_n_plus_1` asserts a 6-task page (every task by a **different** author) costs the same number of SELECTs as a 1-task page, and that exactly one of them is the author join. There was no query-counting helper in the repo, so the test registers a `before_cursor_execute` listener on the `Engine` class and keeps SELECTs only (transaction bookkeeping can't skew it). Mutation-checked: passing `author=None` from the route (i.e. reverting the precompute) makes it fail with `17 == 7`.

Frontend: `frontend/src/api/tasks.ts` only, and the new fields are **optional**, mirroring the `in_progress_count` precedent so pre-existing test fixtures/mocks stay valid.

## 2 · Docs

- **ADR-0057 — Task detail pages carry no faction voice** (new). One shared `detail.*` copy set; where a design's word differs from the domain noun in `CONTEXT.md`, the domain noun wins. Faction voice stays law everywhere else. The 8 shipped vocabularies it retires are quoted verbatim from `tasks.json` (all verified present).
- **ADR-0056 — amended, not duplicated.** Task detail joins the same reversible mobile-collapse experiment; its one-line revert is `TaskDetail.tsx`'s `formFactor === "mobile"` branch, and `pages/taskDetail/mobileArchetypes/*` + the `mobileTaskDetail` manifest rows stay registered but dormant. One verdict covers both surfaces.
- **CONTEXT.md** — three edits: the **Alias slug** entry marked retired (it named a `SLUG_ALIASES` map in `utils/factionDispatch.ts` that exists nowhere in the repo; `FACTION_ALIASES` is genuinely empty); a new first-class **Albescent** entry after **Default archetype**; one added clause on **Archetype** naming the ADR-0057 carve-out (the entry is otherwise untouched).

## Deviations / notes

- The signups route keeps `response_model=list[dict]`. Its dict already disagrees with `TaskSignupOut` on two fields (`praxis_type`/`joined_at` vs the schema's `status`/`signed_up_at`) — pre-existing, and reconciling it would change the response shape, which is out of this issue's scope. Flagged for whoever builds the roster in a later F-issue.
- `PendingTaskOut` (admin) keeps its own `created_by_name` and leaves the new fields at their defaults — admin surface, not part of this epic.
- The new `TaskOut` fields carry defaults (`""`/`None`/`0`) because `TaskOut.model_validate()` is also called on bare metatask ORM rows for the praxis seal stack, which have no author attributes. Same reason `in_progress_count` defaults to 0.

## Testing

From `backend/`, against a dedicated DB (parallel agents share `worldzero_test`):

```
TEST_DATABASE_URL=postgresql+asyncpg://worldzero:localdev@localhost:5432/wz1029_test \
  pytest --cov=. --cov-fail-under=80
```

→ **791 passed**, coverage 89.69%. Frontend: `tsc --noEmit` clean, `eslint src/api/tasks.ts` clean.

## What to eyeball

- **Visual QA is outstanding, and there is nothing to look at yet.** This PR ships no UI: the author row has no consumer until F2 (#1030) draws it. Everything here is verified by tests, `tsc` and `eslint` only.
- ADR-0057's decision — one neutral copy set on task detail — is the constraint every one of the 9 skin issues inherits. Worth reading before the skins start.
- The ADR-0056 amendment binds task detail's mobile verdict to the cards' verdict. If you want them evaluated separately, say so now rather than after 9 skins ship.
- The CONTEXT.md **Albescent** entry and the retired **Alias slug** entry are quoted verbatim from the issue; check they read the way you meant.
- `created_by_faction_slug` = the author's member faction. If the author row on the designs was meant to wear the *task's* faction, that's a one-line change here and a wrong assumption to catch early.
- `created_by_level` is the author's level **now**, not at proposal time. Era reset will change what an old task's byline says.


pixieofhugs:
Drive-by from the task-**cards** session — **not touching this branch**, just flagging a conflict that's easy to resolve the wrong way.

This PR amends `docs/adr/0056-*.md` from a base where it still reads `**Status:** Proposed — under evaluation, not settled law`. That's no longer true on `main`. #1044 (merged) flipped it to **Accepted**:

- The owner did hands-on mobile QA and ruled *"new cards are better on mobile. Get rid of the old ones entirely."* — the exact condition ADR-0056 was gated on.
- The `mobileTaskCard` revert path was then **deleted**: dispatcher + 8 `*MobileTaskCard.tsx` + `cards/shared.tsx` + 2 tests (12 files), the `mobileTaskCard` key removed from `SURFACE_KEYS` and the `FactionManifest` field, and the registration stripped from all 7 factions.
- Consequences was rewritten from the "if accepted / if rejected" fork into what actually happened, and the header scopes the ADR-0035 supersession to task cards.

This PR is currently `DIRTY`, so the conflict will surface — the risk is resolving it by taking this branch's side, which would silently revert the ADR to Proposed and restore prose describing 12 files that no longer exist.

**What a correct resolution keeps:** Status `Accepted`, the verdict + date + reasoning, and the rewritten Consequences — while adding this PR's task-detail scope on top (the title broadening to "task cards (and task detail)" and the `#1028`/`#1029` cross-refs look right, they just need to sit on the Accepted version).

One knock-on worth a thought, entirely your call: the amendment extends ADR-0056 to task detail, but the verdict on `main` was reached by the owner living with the *cards* only. Detail pages haven't been through that evaluation. It may be cleaner to have the ADR read as Accepted-for-cards with task-detail explicitly still under evaluation, rather than inheriting the verdict for a surface nobody has QA'd yet.

Also FYI: #1039 says it owns "the ADR-0056 verdict" — that portion is already done in #1044. I left a note there. Its key-sweep and unvendor parts are untouched and still yours.

PR #1067, opened 2026-07-28
Rasterise the ensō (211 KB → 31 KB), and guard the load-time contract for new surfaces

pixieofhugs:
Follow-up to #1063. Two things: the last heavy asset, and making sure the win survives the next twenty surfaces.

## 1. The ensō: 211 KB → 31 KB, and a paint cost removed

The ensō is UA's signature mark, delivered as a CSS mask so its colour comes from a token (ADR-0049). **That delivery is unchanged — only the asset behind it is.**

The source is 284 brush strokes at 0.12–0.22 opacity under an `feTurbulence` + `feDisplacementMap` filter: 705 KB raw, 211 KB gzipped. It carried two costs:

- **Bytes** — 211 KB gzipped for one decorative mark.
- **Paint** — the displacement filter re-runs over all 284 paths on *every render*, so a task list showing twenty UA cards paid for it twenty times. **This is the one that would have been felt as "the site is slow" after load, and no byte budget would ever have caught it.**

A 512px WebP costs **31 KB** — 6.8× smaller — and bakes the filter in, so the roughening is computed once here instead of on every paint.

**Nothing is lost.** A mask reads only the alpha channel, so the drawing's three orange tones were already being discarded. Flattening RGB to black before encoding is what gets the file this small: the alpha plane is stored losslessly either way, so once RGB is flat, q50 and q95 produce byte-identical output.

512px covers every call site — the largest legible render is the 138px score stamp, and the only larger one is a 420px backdrop watermark at 0.06 opacity. Measured alpha error against the vector:

| Display size | Mean alpha error (0–255) | Coverage ratio |
|---|---|---|
| 420px | 1.08 | 0.996 |
| 138px | 2.18 | 0.961 |
| 96px | 2.27 | 0.967 |
| 38px | 5.34 | 1.044 |
| 16px | 11.57 | 1.149 |

Under 1% where it's legible. The 16px case is ~15% denser — that's a downscaling artefact on a mark that is a few pixels of orange ring, and I judged it not worth a second asset. Say the word if you'd rather have one.

The source SVG moves to `src/components/factionMarks/enso.source.svg` — beside the component that owns it, and out of `public/` so it is no longer served. Verified absent from `dist/`; regeneration steps are in `Enso.tsx`.

## 2. A guard, so this doesn't need doing again

`manifestsStayLazy.test.ts` walks every faction manifest and fails if any surface is wired to a component that is not code-split. The regression it catches looks completely ordinary:

```ts
import CovenTaskDetail from '../pages/taskDetail/archetypes/CovenTaskDetail'
taskDetail: () => CovenTaskDetail,
```

Because `factions/index.ts` imports all eight manifests, a static import there lands in the entry chunk and drags everything that component imports with it. That is precisely how the app came to ship ~184 archetypes to a logged-out visitor.

The byte budget from #1063 would notice *eventually*, once the damage crossed a threshold. This fails on the **first** one and names the faction and the surface. Verified by injecting the regression into `coven.ts` and confirming the failure message.

## 3. The doc

`docs/agents/load-time.md`, with a routing row in `CLAUDE.md`. It records:

- **Why most new surfaces are already free** — register through the manifest and splitting is automatic; this is the headline, because the answer to "how do I keep this fast" is mostly "keep doing what the neighbours do".
- The four ways to break that, including the one that looks like progress: route-splitting alone recovered only 58 KB of 420 because the barrel still held the weight.
- The asset rules this work produced — a mask needs only its alpha, size the raster to the largest *legible* render, and a per-render SVG filter is a paint cost no byte budget catches.
- The `renderToStaticMarkup` constraint that makes `lazyArchetype` necessary instead of `React.lazy`, and the one recurring edit new faction surfaces need (`resolvedArchetype()` in identity assertions).

It deliberately does **not** restate current byte numbers — `bundle-budget.mjs` owns those, and a doc that caches them just goes stale.

## Verification

Lint, `tsc`, **1856/1856 tests**, build and budget green on top of the merged #1063. Browser-verified against the production build: the asset decodes 512×512 with 19.2% inked coverage, and the masked elements resolve to `enso.webp` tinted `rgb(240, 138, 78)` from the UA token, so ADR-0049 still holds. No console errors.


pixieofhugs:
Added a third commit: dropped the one genuinely unused font family (**Cutive Mono**) and extended `fontsLoaded.test.ts` to guard the reverse direction.

**A correction worth recording.** Earlier in this work I identified *five* unused families. Re-checking against current main, four of them — Quicksand, Grenze Gotisch, Poiret One, Spectral — are now live in the Coven and Ephemerists task cards. Acting on that stale reading would have silently downgraded two surfaces to fallback fonts. Only Cutive Mono was genuinely unused.

The removal itself is small: the blocking stylesheet goes 77,472 → 76,609 bytes, 178 → 176 `@font-face` blocks. **The guard is worth more than the removal**, because this failure mode arrives by deletion — the card that wanted the font goes away, the request stays behind, and nothing notices. `fontsLoaded.test.ts` now fails in both directions, reading the tree at run time so it cannot go stale the way my manual reading did.

One implementation note: the new check is a plain text search rather than the existing `namedFamilies` helper, which only reads `font-family` declarations. Most families reach the page via a custom property (`--font-faction-engraved: "Cinzel", serif`), which is not one — the first version flagged fourteen live families for exactly that reason. Being over-permissive here can only keep a font, never drop a live one.

Issue #571, opened 2026-07-15
you should get an update when a player submits on a task you are collaberating on


pixieofhugs:

pixieofhugs:
## Grill result → ready-for-agent

**Approach: derived feed, not a notifications table.** Updates is a computed activity feed (ADR-0036) — `backend/services/activity_feed.py` fans out `FEED_SOURCES` at request time. So this is **one new `FeedSource`**, consistent with the existing 12 event types. No new table, no write-time emit.

### What "collaborating on a task" means
A collab praxis (`PraxisType.collab`) is co-owned; membership is `PraxisMember` rows (`backend/models/praxis.py:142-164`), each with a `has_submitted` bool. "A co-member submitted" = another member flipped `has_submitted = True` (`services/collab_consensus.py:111-124`).

### Scope
1. **New event type `collaborator_submitted`** — add a `FeedSource` entry (`activity_feed.py`, `FEED_SOURCES` ~line 634 + a query builder/row mapper mirroring the existing ones). Query: `PraxisMember` rows where `has_submitted == True`, on collab praxes where the **viewer** is also a member, **excluding the viewer's own membership**. Recipients therefore = every other member.
2. **Add `submitted_at` timestamp to `PraxisMember`** (`models/praxis.py:142-164`) — the feed sorts strictly by `timestamp` (`activity_feed.py:917`) and today the model only has `joined_at`, which is the wrong time. Set `submitted_at` where `has_submitted` is flipped (`collab_consensus.py:113`). Schema change follows repo Strategy A (model edit + baseline, no incremental migration).
3. **Frontend card** — add a `collaborator_submitted` case to `FeedCardRouter.tsx`; reuse the existing friend-completion card style (actor submitted on <task>). Shared by desktop + mobile Updates (`useUpdates.ts`).

### Notes / bounds
- Only `collab` praxes (solo/duel have a single member — no co-members to notify).
- Fires per co-member submission (the moment `has_submitted` flips), not on final consensus seal.
- No self-notification (viewer excluded).

Issue #1418, opened 2026-07-31
Roster F3 — Nudge the crew on the awaiting footer


pixieofhugs:
The design's collab waiting stage has a single **"Nudge the crew"** button in the footer, beside "Edit my write-up". Shipped has per-member nudge buttons in the roster and nothing bulk — `git grep -in "nudge.*crew|nudgeAll|nudgeCrew"` returns nothing.

## Blocked by B1

There is no bulk endpoint, and the reason this is its own issue rather than a line in F1 is that the answer is server-shaped, not visual.

`NUDGE_COOLDOWN` is one day per (sender → recipient → praxis) and a repeat is a **422** (`backend/services/nudge.py:36, 248-254`). A "nudge everyone" press will therefore routinely be *partly* refused — some of the crew inside their window, some outside. That is a response shape decision, and B1 owns it.

Do not start this by fanning out N client calls and swallowing the 422s. The cooldown is server-owned today and surfaced viewer-relatively through `nudged_at`; moving that rule into the frontend would put the same rule in two places.

## The work here

Once B1 lands:

1. Mount the bulk action in the awaiting footer, beside the existing "Edit my write-up" (`waiting/PraxisWaitingSurface.tsx:784-800`).
2. Wire it through `useEditPraxis`, next to the existing per-member `state.nudge`.
3. **Report the partial result honestly.** If four of the crew were nudged and one was already nudged an hour ago, say so. Silence would read as "all five nudged" and the player would have no way to know otherwise.
4. Hide, do not disable, when nobody is nudgeable — the repo convention, and the same rule the per-row button already follows (`CollabRoster.tsx:275-305`, gated on `canNudge`).

Who may press it follows the existing per-member authorisation, which the bulk route must preserve: on a collab the sender must be a member **who has already filed** (`_authorize_collab`, `backend/services/nudge.py:67-81`). That is the same gate `CollabRoster` already applies via `gate.iCast` (`:163`).

## Not in scope

The per-member Nudge button. It is already built, already better than the design — the design's version tracked nudged state in local component state, whereas shipped reads `member.nudged_at` off the wire so the cooldown survives a reload. Leave it alone; the deliberate divergence is documented at `waiting/PraxisWaitingSurface.tsx:50-57`.

PR #1500, opened 2026-08-01
Roster B1 — bulk nudge endpoint, member submitted_at, relaxed rescind guard

### LINKED ISSUE #1415: Roster B1 — bulk nudge endpoint, member submitted_at, and two invite rulings
The roster redesign needs almost nothing from the backend — an audit found all four collaborator states, per-member nudge, the tally inputs, invite/rescind/search and leave are **already on the wire**. `declined` is even emitted today and read by no UI.

This issue collects the only real gaps.

## 1. Bulk nudge — the one genuinely missing endpoint

The design's awaiting stage has a **"Nudge the crew"** button. There is no bulk route: `POST /praxes/{praxis_id}/nudge/{character_id}` (`backend/routers/praxes.py:450`) is the only one, and a repeat inside the cooldown is a **422**.

`NUDGE_COOLDOWN = timedelta(days=1)` (`backend/services/nudge.py:36`) is per (sender → recipient → praxis). So a bulk nudge is **inherently partial-success**: some recipients will be inside their window and some outside, and the response has to say which.

Decide the shape. The repo already has a precedent for partial-success batching in `MediaUploadResultOut` (`backend/schemas/praxis.py:19-37`) — follow it rather than inventing a second idiom.

The alternative is to let the client fan out N calls to the existing route and swallow 422s. That is cheaper but pushes the cooldown rule into the frontend, where today it is server-owned and viewer-relative via `nudged_at`. Prefer the endpoint.

Guards `send_nudge` already enforces, which the bulk route must preserve (`backend/services/nudge.py:201-264`): praxis must be `in_progress`/`pending`; recipient must be a member who has not submitted; on a collab the sender must be a member **who has already filed**; self-nudge is 400.

## 2. `PraxisMemberOut.submitted_at` — one line, no migration

`PraxisMember.submitted_at` exists as a column (`backend/models/praxis.py:158-162`, "when `has_submitted` last flipped True", #571), is maintained by `collab_consensus.py`, and **never reaches the wire**.

Add it to `PraxisMemberOut` (`backend/schemas/praxis.py:40-55`), populate in `_build_member_out` (`backend/services/praxis.py:88-107`), mirror in `frontend/src/api/praxis.ts:26-42`. No migration.

## 3. Rescind permission — a ruling, maybe no code

Any member may **invite** (`backend/services/praxis.py:1731`), but only the **inviter** may rescind (`cancel_invite`, `backend/services/praxis.py:1843-1844`, 403 otherwise).

The design puts a rescind affordance on every pending row. Two ways to make that honest:

- relax `cancel_invite` so any member of a co-owned praxis may rescind any pending invite — consistent with ADR-0013 (`created_by_id` grants no powers), or
- leave the guard and have F1 gate the control on `invite.inviter_id === currentCharacterId`.

**Recommendation: relax the guard.** A collab is co-owned; "only the person who typed the name may untype it" is a rule nobody would guess, and the frontend alternative means a control that appears on some rows and not others with no visible reason.

## 4. Precedence for duplicate invites — a ruling, no code

A character can hold **both** a `declined` row and a later `pending` row: the duplicate guard only blocks a duplicate *pending* invite (`backend/services/praxis.py:1739-1752`), and declining never deletes (`:1794-1798`).

So a four-state roster needs a stated rule. **Recommendation: newest wins**, decided client-side in F1 — a re-invited player reads as `Invited`, not `Declined`.

## Not needed

- `nudged_at` on list-route cards. It is detail-route only by design (`backend/services/praxis.py:94-97`) and this roster is not mounted on a card surface.
- An invite `responded_at` column. Only required if a declined row must show *when* it was declined; it can say "invited {created_at}" instead. Skip unless F1 asks.

pixieofhugs:
Closes #1415

## Summary

Backend audit found the roster redesign needs almost nothing new — this closes the three real gaps it found:

1. **Bulk nudge — `POST /praxes/{praxis_id}/nudge`**. The design's awaiting stage needs a "Nudge the crew" button; only the single-recipient route existed before. The bulk route nudges every eligible member (not the sender, hasn't submitted) in one call, preserving every guard `send_nudge` already enforces (praxis open, collab sender must have filed, self-nudge 400). A recipient still inside their 24h per-(sender, recipient) cooldown is **skipped and reported**, not failed — the call is inherently partial-success, so the response follows the existing `MediaUploadResultOut` batching precedent (`backend/schemas/nudge.py`, new `NudgeResultOut`).
2. **`PraxisMemberOut.submitted_at`**. The column already existed (#571) but never reached the wire. Added to the schema and the service builder — no migration needed — and mirrored onto the frontend `PraxisMemberOut` TS type (marked optional there only because ~30 pre-existing test fixtures predate the field; the server always sends it).
3. **Rescind permission relaxed**. `cancel_invite` previously 403'd anyone but the original inviter. Any member may invite, so any member of a co-owned praxis may now rescind any pending invite too, per ADR-0041 (`created_by_id` grants no special powers).

The fourth item from the issue (duplicate-invite precedence, "newest wins") is an explicitly frontend-only ruling for a future roster UI issue — no backend code needed for it.

## Test plan
- `pytest --cov=. --cov-report=term-missing --cov-fail-under=80` from `/backend`: **964 passed**, coverage **91.92%** (floor 80%).
- New integration coverage in `test_nudge.py` (bulk-nudge success, partial-success skip-not-fail, guard preservation: self-nudge, non-member, wrong praxis state, sender-hasn't-submitted-on-collab) and `test_praxes.py` (`submitted_at` on the wire; any co-owned member can rescind, outsider still 403).
- `npm run typecheck` from `/frontend`: clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FfU8VupEYuaXCn5GRQ32xm)_

### DIFF
--- a/backend/routers/praxes.py
+++ b/backend/routers/praxes.py
@@ -49,7 +49,7 @@ class InviteResponse(BaseModel):
 
 class MetataskApply(BaseModel):
     task_id: int
-from schemas.nudge import NudgeOut
+from schemas.nudge import NudgeOut, NudgeResultOut
 from schemas.vote import VoteCastOut, VoteOut, VoteTallyOut, ViewerStatsOut
 from services.scoring import compute_votes_available
 from services.praxis import (
@@ -82,7 +82,7 @@ from services.praxis_out import (
     build_praxis_out,
 )
 from services.media import process_and_save_media
-from services.nudge import send_nudge
+from services.nudge import send_bulk_nudge, send_nudge
 from services.vote import cast_vote_on_praxis
 
 logger = logging.getLogger(__name__)
@@ -469,7 +469,7 @@ async def cancel_invite_route(
     await cancel_invite(
         praxis_id=praxis_id,
         invite_id=invite_id,
-        inviter_id=character.id,
+        requester_id=character.id,
         session=session,
     )
     return Response(status_code=204)
@@ -499,6 +499,32 @@ async def nudge_route(
         session=session,
     )
     return NudgeOut.model_validate(nudge)
+
+
+@router.post("/{praxis_id}/nudge", response_model=List[NudgeResultOut])
+async def bulk_nudge_route(
+    praxis_id: int,
+    character: Character = Depends(get_current_character),
+    session: AsyncSession = Depends(get_db),
+):
+    """"Nudge the crew" — poke every member still owing this praxis (Roster B1, #1415).
+
+    Recipients are every member other than the sender who has not yet
+    submitted. The response is one entry per eligible recipient, following the
+    ``MediaUploadResultOut`` partial-success idiom (``schemas/praxis.py``):
+    most recipients succeed, but a recipient still inside their per-recipient
+    24h cooldown (``services.nudge.NUDGE_COOLDOWN``) is SKIPPED rather than
+    failing the whole call, and reported back with ``error`` set instead of
+    ``nudge``. Guards about the sender (member-of / has-cast-for-a-collab,
+    praxis still open, participant-of-an-active-duel) are call-wide and raise
+    for the whole request, exactly as the single-recipient route raises them —
+    all in ``services.nudge.send_bulk_nudge`` so this handler stays thin.
+    """
+    return await send_bulk_nudge(
+        praxis_id=praxis_id,
+        from_character_id=character.id,
+        session=session,
+    )
 
 
 @router.post("/{praxis_id}/kick/{member_id}", response_model=PraxisOut)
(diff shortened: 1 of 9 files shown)


pixieofhugs:
Superseded by #1505, which merged the same three gaps for #1415 (bulk nudge with partial-success `NudgeResultOut`, `PraxisMemberOut.submitted_at`, relaxed rescind).

Closing rather than merging — this branch is now CONFLICTING against main, and #1505 is a superset of it:

- same nine files, same design (including `submitted_at` optional on the TS side for the pre-existing fixtures — that is how `frontend/src/api/praxis.ts:39` reads on main)
- plus `frontend/src/api/nudge.ts`, the typed client the roster F3 issue (#1418) consumes
- plus a query-count assertion pinning the bulk route to a constant statement count (13 for a crew of 2, 13 for a crew of 8)
- plus a fix for a stale-collection read in the rescind path

Nothing here is lost. Apologies for the duplicated run — this PR was open eight hours before I dispatched #1415, and I did not re-check open PRs before shaping that batch.

Issue #1567, opened 2026-08-02
The type filter opens an empty picker on a quiet board — hide a facet trigger with no options


pixieofhugs:
https://worldzero.org/updates

pixieofhugs:
## Diagnosed — and the menu is behaving as written. The screenshot did not attach (`<!-- Error while uploading Screenshot_20260802-024504.png -->`), but the cause is in the filter code.

`typeFacetOptions` (`frontend/src/pages/updates/updatesFilters.ts:306-324`) keeps a row only when:

```ts
const rows = candidates.filter(
  (type) => (byType[type] ?? 0) > 0 || selected.includes(type),
)
```

**A type appears only if its count is above zero**, or it is already ticked.

The database was wiped, so almost nothing has happened, so almost every type has a count of zero — and the menu is empty. Not a break: the docstring gives the reason, which is that a selected type must keep its row at zero so it can be un-ticked, and that otherwise the list fills with dead options.

The defect is what happens next: the trigger still renders, so tapping "Type" opens a picker with nothing in it and no explanation.

## Owner ruling, 2026-08-02: hide the trigger when the facet has no options

Not "list every type with a zero count", and not "list them disabled".

This is the repo's standing rule — *hide unusable controls, don't show them disabled* (CLAUDE.md) — and it keeps this facet consistent with its sibling rather than special-casing one axis of the same widget.

Accepted cost: the control appears and disappears as activity comes and goes. On a live board with any traffic it is always present; it is the empty board where it vanishes, which is exactly when it had nothing to offer.

## Where to put it

**In `OptionPicker`, not in `FilterBar`'s facet loop.** The component knows when it has nothing to offer; `FilterBar` (`components/ui/FilterBar/index.tsx:122-123`) maps facets to pickers unconditionally and should not have to reason about each facet's contents.

Return `null` when `facet.options` is empty.

## This very likely fixes the faction facet too — verify

`factionFacet` (#1446) is the sibling built on the same widget and, as far as I can tell, filters by count the same way. On a freshly wiped board it would present the same empty picker. Check it while you are here; if it does, one `null` return fixes both and the issue should say so.

Two things to confirm rather than assume:

[...]

Issue #1365, opened 2026-07-30
F1 — FilterBar: the shared responsive filter component


pixieofhugs:
Part of #1361. **Net-new component, mounted nowhere.** No page file is touched — F2 (#TBD) and F3
(#TBD) do the mounting. Runs in parallel with B1 (#1362) and B2 (#1364).

Design: [World Zero filter redesign](https://claude.ai/design/p/355cf736-de2c-4705-88cf-755d49122dcf).
Both files show the same component; read both. `Task Filters - Eligibility.dc.html` is the fuller one
(it has the four-segment rail); `Praxis Filters - Voting.dc.html` shows the phone layout.

## Scope

One responsive component under `frontend/src/components/ui/FilterBar/`, presentation-only —
it takes values and setters, owns no fetch. **No mobile twin**: this runs with the grain of
kit-hygiene epic #1312, where #1319 collapses a form-factor split and #1313 kills a
modal-vs-sheet duplicate. The `MOBILE_ARCHETYPE_BY_SLUG` rule in CLAUDE.md governs faction-dressed
surfaces; the filter bar is chrome, is not skinned per faction, and stays outside that seam.

## Parts

1. **Segmented rail** — 2 to 4 segments, sliding thumb, spectrum-ringed border. One rail = one value =
   one `ORDER BY`; rails never compose. The design's `seg2`/`seg4` thumb maths is
   `calc(i * (100% - 6px) / n + 3px)`.
2. **Faction multi-select** — popover at tablet and up, bottom sheet on phone. One conditional, not
   two components (#1313 is the precedent). Rows are checkbox + sigil + name; selected sort first.
   **No per-faction counts** — cut by owner ruling, see #1361.
   **Roster = `GET /factions` plus the explicit `na` sentinel**, exactly as `FilterFactionTabs` does
   today with its documented rationale. Not the design's hardcoded 9.
3. **Applied-chip row** — one removable chip per active filter, plus "Clear all". Faction chips carry
   their sigil.
4. **Collapse** — open on desktop, collapsed on phone, with a count badge of applied filters. Search
   and the chip row stay outside the collapse, as in the design.
5. **Empty states** — three, not the design's one. See #1361 ruling 9. The "Clear all filters" button
   the design puts in the empty state is a real improvement; keep it.

## Chrome rules

- **The spectrum ring stays** (#1361 ruling 1): `--faction-default-rainbow` is `na`'s spectrum and
[...]

Issue #1325, opened 2026-07-30
The CSS budget warn line sits below the current size, so the ratchet warns on every build and signals nothing


pixieofhugs:
The CSS half of `npm run budget` has been printing WARN on `main` for a while and nobody is acting on it, which means it has stopped being a ratchet.

Found independently by **three** agents in the 2026-07-30 batch (#1263, #1236, #1285), each reporting it as pre-existing and unrelated to their own diff — which is the tell. When three separate builders all have to say "that warning isn't mine", the warning has become background noise.

## The numbers

`frontend/scripts/bundle-budget.mjs`:

```js
css: { warn: 20_000, fail: 25_000, target: 20_000 },
```

Measured on `main`: **CSS 20.7 KB** — over the 20 KB warn line, under the 25 KB fail line. So every build prints the loud WARN block and exits 0.

## Why this matters more than 700 bytes

The script's own docblock states the mechanism:

> Over WARN prints a loud block and still exits 0 — you learn a change costs weight without being blocked by it. […] **WARN sits just above today's number so any real growth speaks up immediately**, while the build stays green. Ratchet WARN downward each time a chunk moves.

The design depends on WARN sitting *just above* the current number. It is now sitting *below* it. That inverts the whole mechanism: instead of speaking up on real growth, it speaks up on every build regardless, so the signal carries no information and the next 4.3 KB of CSS growth arrives silently under cover of a warning everyone has learned to scroll past.

Note `warn` and `target` are both 20,000 for CSS — unlike JS, where `target` (120 KB) sits well below `warn` (150 KB). So CSS has no headroom band by design; being over warn *is* being over target.

## Two honest options

1. **Shed the ~700 bytes** and put CSS back under its own line. Restores the ratchet as designed, with no decision required.
2. **Raise the line deliberately**, with the reason written into the docblock ledger the way #1045's JS entries are. If 20 KB is no longer the right target after the epic #1179 / #1192 faction dress work, say so on the record and set the new number just above today's.

What is not an option is leaving it: a check that warns unconditionally is a check that has been turned off without anyone deciding to turn it off.

## Not urgent

Nothing is failing and nothing is blocked. This is filed so the mechanism gets repaired before the next CSS-heavy epic lands 4 KB under the noise floor.

pixieofhugs:
Rung-zero verdict from the 2026-07-30 builder-bot run: **DEFER**, not dropped — with a reason that changes the number.

[...]

PR #1586, opened 2026-08-02
The filter sheet clears the tab bar, and an empty facet draws no trigger

### LINKED ISSUE #1566: The filter sheet's Done button is clipped by the tab bar and the home-indicator inset on mobile
https://worldzero.org/updates

pixieofhugs:
Fixes #1566 and #1567 — two faults in the same widget, both reported from a phone. Neither screenshot attached, so both were diagnosed from the code.

## #1566 — the Done button was clipped

`.filter-factions__sheet` is `position: fixed; bottom: 0` and reserved nothing for the two things already living at the bottom of a phone viewport: the **3.5rem fixed tab bar** and the **home-indicator inset**. `ShellContent` computes exactly that clearance for the page region; the sheet did not, so its last child — Done — landed under one of them. There was no `env(safe-area-inset-bottom)` anywhere in `index.css`.

`overflow: auto` also sat on the sheet rather than on the option list, so a long list scrolled the heading and the Done button out of view, and reaching Done meant scrolling past every option. The scroll moves to `__rows`, **scoped to the sheet** so the desktop popover — which has no height cap — keeps growing to fit.

The tab bar's height becomes **`--mobile-tabbar-height`**. It was a bare `3.5rem` in `ShellContent`, and this is its second consumer; two literals would drift the first time the bar changed height.

## #1567 — the type picker opened empty

Both facets drop a zero-count row, so on a quiet board — a fresh account, or a freshly reset one — every row is filtered out and the trigger opened a picker with nothing in it. Working as written, but a dead end.

`OptionPicker` now returns `null` when it has no options: **hide the unusable control rather than disable it** (CLAUDE.md).

It tests the option *list*, not the counts, because a **selected** option keeps its row at zero — a `?types=nudge` deep link on a view with no nudges must stay removable, so its trigger must stay reachable.

The fix lands in `OptionPicker` rather than `FilterBar`'s facet loop: the component knows when it has nothing to offer, and the **faction facet gets it for free**.

## One existing test pinned the old behaviour

`updatesFilterBar` asserted the Type trigger renders on a bare `/updates`. But that harness runs no effects, so the counts never arrive and the facet is *always* empty there — the assertion only passed because the trigger drew unconditionally. It becomes two tests: the trigger appears with a selected type, and is absent on an empty board.

Three new tests in `filterState` cover the widget directly: no trigger when every row is filtered out, a trigger when a selected option keeps its row, and the rest of the bar untouched.

## Blast radius

`OptionPicker` is the shared `FilterBar` picker (#1365/#1446), so both fixes land on **tasks, praxes and updates** at once. Worth a look at all three at mobile width.

## Verified

- **3,868 tests** across 222 files pass
- `tsc --noEmit` clean, `eslint` clean
- `npm run build` succeeds; bundle budget **within limits** — though CSS is at 22.4 KB against a 22.5 KB warn line, so this leaves ~0.1 KB of headroom. That is #1325's territory, not something this PR should move.

### DIFF
--- a/frontend/src/components/layout/ShellContent.tsx
+++ b/frontend/src/components/layout/ShellContent.tsx
@@ -59,7 +59,10 @@ const DESKTOP_REGION = 'flex-1 relative max-w-[min(92vw,1600px)] mx-auto w-full 
 
 const MOBILE_REGION_STYLE: CSSProperties = {
   zIndex: 5,
-  paddingBottom: 'calc(3.5rem + env(safe-area-inset-bottom))',
+  // The tab bar's height is `--mobile-tabbar-height` in index.css, not a literal
+  // here: the filter sheet needs the same clearance (#1566) and two hardcoded
+  // 3.5rems would drift the first time the bar changes height.
+  paddingBottom: 'calc(var(--mobile-tabbar-height) + env(safe-area-inset-bottom))',
 }
 const DESKTOP_REGION_STYLE: CSSProperties = { zIndex: 5 }
 
--- a/frontend/src/components/ui/FilterBar/OptionPicker.tsx
+++ b/frontend/src/components/ui/FilterBar/OptionPicker.tsx
@@ -39,6 +39,16 @@ export default function OptionPicker({ facet }: { facet: FilterFacet }) {
   const isSheet = useFormFactor() === 'mobile'
   const { label, options, selected, onChange, renderOrnament } = facet
 
+  // A facet with nothing to offer draws no trigger (#1567). Both facets hide a
+  // zero-count row, so on a quiet board — a fresh account, or a freshly reset
+  // one — every row is filtered out and tapping the trigger opened an empty
+  // picker. Hide the unusable control rather than disable it (CLAUDE.md).
+  //
+  // A SELECTED option always keeps its row even at zero, so a deep link like
+  // `?types=nudge` on a view with no nudges still renders the trigger and stays
+  // removable. That is why this tests the option list and not the counts.
+  if (options.length === 0) return null
+
   const panel = (
     <div
       className={isSheet ? 'filter-factions__sheet' : 'filter-factions__popover'}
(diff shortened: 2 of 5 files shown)


pixieofhugs:
Closing as **superseded, not rejected** — this was correct work, built concurrently by another session.

Both issues it fixes are already closed:

- **#1567** → PR #1589 (`9b6833af`) — `OptionPicker` returns `null` when a facet has no options.
- **#1566** → PR #1590 (`c82b6236`) — the sheet reserves the tab bar + safe-area strip, and the scroll moves to the option list.

The two arrived at the **same diagnosis and the same two fixes**, independently. Where they differ, what merged is a superset:

| | this PR | merged |
|---|---|---|
| token name | `--mobile-tabbar-height` | `--tab-bar-height` + `--tab-bar-clearance` |
| consumers de-duplicated | 2 (`ShellContent`, the sheet) | **5** — also `DefaultCreateCharacter.tsx`, `DefaultEditCharacter.tsx`, `factionDetail/mobileArchetypes/shared.tsx` |
| scroll moved to `__rows` | yes | yes |
| `WORLD_ZERO_STYLE.md` | — | §5 gained the bottom-edge rule |

Nothing here is lost by closing it.

## One thing this PR got right that is worth recording

Its framing — *"two literals would drift the first time the bar changed height"* — is exactly the argument that justified minting the token rather than inlining the clearance. The merged version found three more literals on the same reasoning.

## And one correction to both

Both PRs inherited #1566's explanation that the sheet's `z-index: 40` beats the tab bar's `10`. **That is backwards.** `ShellContent` sets `position: relative; z-index: 5`, which opens a stacking context, so the sheet composites at 5 against a bar at 10 in the root context — the bar genuinely paints over the sheet. The fix (reserve the strip, don't fight paint order) is right either way, but the mechanism matters for the next person here, and it has a live consequence: the tab bar stays **tappable** behind an open sheet's scrim. Filed as #1591, which needs a portal and is not reachable from CSS.

pixieofhugs:
Reopened this by mistake, and closing it again — everything in it is already on `main`, shipped independently and in better shape.

- The tab-bar clearance landed as a composed `--tab-bar-clearance` token (`index.css:1774`), which folds the bar height and `env(safe-area-inset-bottom)` into one name. This branch spelled that calc out at each of its two call sites, which is exactly the drift the token prevents.
- The empty-facet guard is on `main` at `OptionPicker.tsx:62`, identical.
- The two tests exist on `main` with the same assertions.

Closing as superseded rather than merging. No action needed.

Issue #1636, opened 2026-08-03
Ephemerists surfaces assume a light ground; the plate is dark


pixieofhugs:
Several Ephemerists surfaces were built assuming a light ground. The plate palette is dark. This is the gap, and it is the reason a third of the *Ephemerists Kit* design's JavaScript exists.

Design: https://claude.ai/design/p/9305f548-25ce-4dac-9d35-10867994831f

Foundation for the re-skin epic — do this before the surface queue, or every surface rediscovers it.

## What the design found

Four separate failures, each with its own runtime patch and its own comment:

**1. Grounds do not re-resolve on a theme flip.** *"Some skins paint their ground through a transitioned `background` shorthand, which doesn't re-resolve its token on a theme flip — the plate stays vellum under night ink."* The design keeps a table of six light plate colours and force-pins each to a dark counterpart. The real fix is that a transitioned `background` shorthand cannot carry a themed token — find those declarations and split the colour to a longhand, as the design itself does for `.eph-plate-sheet`.

**2. Links land near 2:1 on the plate.** *"Links inherit one light-mode brown; on the plate's dark grounds that lands near 2:1."* One link ink is used across grounds of very different luminance. The design measures each link against its actual ground and picks the better of two inks. **2:1 is a real WCAG failure, not a preference.**

**3. Light-mode muted ink lands on the dark band.** *"Some skins put light-mode muted ink on the plate's dark band (the profile header)."* `plate-quiet` and `plate-band-quiet` both exist; surfaces are reaching for the wrong one depending on which ground they are actually on.

**4. A translucent veil is being treated as a ground.** *"A 4%-white veil over a dark plate is not a light ground — require real opacity."* Anything computing contrast by walking up for a background colour has to skip near-transparent layers, or it measures against `rgba(255,255,255,0.04)` and concludes the ground is white.

## What to actually build

**Not** the design's approach. It re-measures every node in the document on an interval because a canvas cannot edit tokens. In the real app these are token and stylesheet bugs, and they should be fixed at the source so nothing measures anything at runtime.

Expected shape: correct which token each surface reads, split any transitioned `background` shorthand that needs to carry a themed colour, and give links a ground-appropriate ink through the cascade rather than per-element.

## Guard it

[...]

Issue #1630, opened 2026-08-03
Profile header flair: spectrum rings, Coven sparkles, UA's lotus, Singularity's phosphor ground, SNIDE's ransom slip


pixieofhugs:
Design: https://claude.ai/design/p/2d1855ad-b86d-43dd-9284-3045343e6694 — *Faction Profile Headers*.

Seven bespoke treatments on the profile identity header, one per faction, transcribed from the design's CSS and canvas JS. The shared header changes are filed separately; this is the dressing.

Each of these is independent — they can land one at a time.

## The treatments

**Unaffiliated + Albescent — the progress ring reads in the spectrum.** The level ring's filled arc becomes a rainbow conic gradient instead of a solid `--color-text-primary` arc. The design builds it as `conic-gradient(transparent 0 Ndeg, var(--color-border) Ndeg 360deg)` layered over `conic-gradient(<rainbow stops>)`, preserving the existing fill angle.

**Albescent — a spectrum-stroked sigil, and a drifting frame.** The sigil is `AlbescentSigil` with its strokes painted in the rainbow gradient, in a slot bordered `#60a5fa`. And its spectrum frame **animates**: `background-size: 200% 100%`, `background-position` drifting `0 → 200%` over **9s linear infinite**. Unaffiliated keeps the same frame **static** — that contrast is the point.

**Coven — candle-yellow sparkles.** Eight `CovenSigil` marks scattered across the header ground, twinkling on a 5.5s ease-in-out loop (`opacity .12 → .85`, `scale .8 → 1.08`), each with its own delay. Positions and sizes from the design, as `[left%, top%, size, delay]`:
`[7,18,22,0] [22,72,14,1.4] [41,12,17,2.6] [58,64,12,0.7] [72,28,20,3.4] [86,78,15,1.9] [93,34,11,4.2] [63,88,16,2.2]`
Decorative — `aria-hidden`, `pointer-events: none`, behind the content.

**UA — the lotus, ghosted into the ground.** `Lotus` at **340px**, `opacity .1`, positioned `right: -40px` and vertically centred, coloured `--faction-ua`. Behind the content, `aria-hidden`.

[...]

Issue #1626, opened 2026-08-03
CredentialCard: drop the kicker and bio, grow the portrait, and swap the faction label for a labelled sigil


pixieofhugs:
Design: https://claude.ai/design/p/2d1855ad-b86d-43dd-9284-3045343e6694 — *Faction Profile Cards*.

The canvas mounts the **shipped** `WZ.CredentialCard` nine times, once per faction, and layers CSS overrides on top. So the design is a **delta list against the real component**, not a redesign — each override in its `<style>` block carries a comment saying what it does. Transcribed below, because agents cannot reach the design URL.

## Read this first: four mount contexts, one component

`frontend/src/components/CredentialCard.tsx` is mounted by:

| where | what it is |
|---|---|
| `pages/characterProfile/archetypes/DefaultProfileBody.tsx:340,678` and `archetypes/profileSkin.tsx:302` | **the character-profile header** — the docstrings call it *"① identity + progression (shared CredentialCard as the header)"* |
| `pages/FieldDesk.tsx:202` | the roster card on `/` |
| `pages/CreateCharacter.tsx:164` | the create-character live preview |

Every delta lands on all four. The design draws it as a 266px chip in a rack of nine; check the profile header and the create preview too.

## The deltas

**1. Drop the "CREDENTIAL" kicker.** `CredentialCard.tsx:131` — the second span in the eyebrow row, `t('credential.credential')` / `t('credential.unaffiliated')`. The `@handle` stays.

**2. Portrait ring 96px → 136px.** `CredentialCard.tsx:141`.

**3. Delete the bio slot** (`:210-226`, the two-line clamp) — see the section below; it does not simply vanish.

**4. The footer's faction label becomes the faction sigil.** A 42px circle — transparent ground, 2px border in the faction accent — containing a **28px `FactionSigil`**. The faction name stops being visible text.

**5. Footer level readout to 12px.**

### The bio moves to the profile, it does not disappear

The bio is currently rendered **nowhere else** — grepping `characterProfile` for `.bio` returns only the three `CredentialCard` mounts. So deleting the slot without a replacement would leave the field editable and unrendered.

**Add an "About" block to the profile.** The spine already reserves the slot: `DefaultProfileBody`'s docstring lists *"② about — skipped in v1 (no field)"*. This is that field arriving.

[...]

Issue #901, opened 2026-07-21
WOW redesign 6/6: mobile skins


pixieofhugs:
## Parent
#787

WOW redesign 6/6. The mobile skins.

## Design source

`WoW Faction Kit.dc.html` §10 (Mobile Skins), project
`d0a6fdd7-a2f3-4f61-834a-3a06ab4acf07`. ⚠️ Subagents cannot reach claude.ai —
the orchestrator must paste §10 and the `mobileThemes` array into the build
prompt.

## Read this before scoping

**The kit drew one mobile screen, not a mobile version of each surface.** §10
covers a status bar + header, a quest list, and a bottom nav — which maps to the
**field desk** and the **mobile task card**, and nothing else. #785 asked for a
mobile twin of every surface; the kit delivered one screen and its chrome.

Every other faction registers **9 mobile surfaces**; UA registers 7. WOW today
registers 1 (`mobilePraxisCard`, from #818).

So this slice builds what is drawn and **derives** the rest from it. That is the
same thing every other faction's mobile build did, and the kit's `mobileThemes`
array is a complete theme contract — bezel, page, rim, field, header gradient,
name/sub/plum/gold inks, card background — for both themes. Derive from it; do
not invent a second palette.

## What to build

From §10 directly:
- `frontend/src/pages/fieldDesk/mobileArchetypes/WowFieldDesk.tsx`
- `frontend/src/pages/tasks/mobileArchetypes/cards/WowMobileTaskCard.tsx`

Derived from §10's chrome plus the matching desktop archetype:
- `mobileTaskDetail`, `mobilePraxisDetail`, `mobileFactionPage`, `mobileProfile`

Register each in `frontend/src/factions/wow.ts` as it lands. Follow the Coven
mobile archetypes for structure and file placement.

**Out of scope for this slice:** `mobileEditPraxis` (#836 owns it),
`mobilePraxisCard` (#818 shipped it), `mobileDuelSeal` / `mobileDuelRail` (the
duel design child), `mobileCreateCharacter`, `mobileEditCharacter`,
`mobileFactionsDirectory`, `mobilePlayersDirectory` — those last four keep
defaulting; file a follow-up if they read wrong in use.

## Constraints

- **Single column. No fixed-px inline grids for layout structure** — see
  `docs/spec/SPEC-faction-ui-profile.md` §1a and the #494 form-factor contract.
  Must survive a 375px viewport.
[...]

PR #1655, opened 2026-08-04
The profile header drops its eyebrow and its caption, and the tagline takes the name's place

### LINKED ISSUE #1629: Profile header: drop the eyebrow and affiliation caption, and the tagline takes the name's place
Design: https://claude.ai/design/p/2d1855ad-b86d-43dd-9284-3045343e6694 — *Faction Profile Headers*, which mounts the shipped `WZ.FactionProfileBody` for all nine factions and crops it to the identity header.

Like its sibling, the design is a **delta list expressed as CSS overrides on the real component**, each with a comment saying what it does. Transcribed below; agents cannot reach the design URL.

This issue is the changes that apply to **every** skin. The per-faction flair is filed separately.

## Blocked by the tagline field

The header's centre slot renders `character.tagline`, which does not exist yet.

## The three shared deltas

**1. Drop the "PLAYER · &lt;FACTION&gt;" eyebrow** above the name — the uppercase, wide-tracked line. Gone on every skin.

**2. Drop the "Unaffiliated · faction pending" tagline line** — the display-italic line at content size. Gone on every skin. (Note: the *design's* name for this is confusingly close to the new field; they are unrelated. This is the na-only affiliation caption, and it goes.)

**3. The name leaves the identity column, and the tagline takes its place.**

The design's reasoning, verbatim: *"the name already reads off the credential card — the identity column carries the bio instead."* The credential card is right beside it showing the name large, so repeating it is redundant.

Owner ruled the slot carries the **new `tagline` field**, not `bio` — the design's own treatment (30px, `max-width: 22ch`) is a slogan slot, and bio is long-form.

Treatment, from the design:

- font-size **30px**, line-height **1.16**, `max-width: 22ch`
- **hidden entirely when the tagline is empty** — house rule, hide rather than show a placeholder

**Be deliberate about day one:** no character has a tagline yet, and this delta also removes the name. Until someone writes one, that column is credential card, level ring and progression bar only. That is expected, not a bug.

## Where the work goes

`FactionProfileBody.tsx` is the **dispatcher** (#459, ADR-0033) — it picks a faction's `profileBody` skin and holds no markup. The header itself lives in `archetypes/profileSkin.tsx` and the eight `*ProfileBody` files.

Put these three deltas in the **shared** layer wherever the header is shared, so no archetype needs editing for them. Anything that genuinely differs per faction belongs to the flair issue, not here.

Remember the profile has **two branches** — the desktop grid and the `MobileProfile` phone stack, which `DefaultProfileBody`'s docstring says are deliberately separate components rather than one layout at two sizes. All three deltas need doing in both.

## Not this issue

- **The credential card's own deltas** (kicker, portrait size, bio slot, sigil footer). The design repeats them because it mounts the same card; they are filed on the CredentialCard issue and land there.
- **The Ephemerists `card-*` repoint**, also repeated here. Filed separately as the faction-wide token change.
- **Per-faction flair** — the spectrum rings, Coven's sparkles, UA's lotus, Singularity's phosphor ground, S.N.I.D.E.'s ransom slip, Ephemerists' brass ring label.

## Tests

- the eyebrow and the affiliation caption are absent on every skin
- the tagline renders when present and the slot is **absent** when blank
- the display name does **not** appear in the identity column (it still appears on the card)
- both profile branches, desktop and phone

Harness is `renderToStaticMarkup` — assert on markup, not effects.

pixieofhugs:
Closes #1629

The identity column's display line was the player's own name, printed at 42–72px next to a credential card that already says it large. #1629 gives the slot to `character.tagline` (the field shipped on the wire in #1643) and deletes the two lines that bracketed the name.

## The seam

`FactionProfileBody` × the identity header, over every skin at both widths. The failing test went in first — 53 red across the matrix — and the deltas landed in the **shared** layer, so no archetype needed markup edits for them:

- `archetypes/profileSkin.tsx` — the header the seven faction kits delegate to, one component at both widths.
- `archetypes/DefaultProfileBody.tsx` — na, albescent and any unskinned slug, **twice**: the desktop grid and the `MobileProfile` phone stack are deliberately separate components, not one layout at two sizes.

## The three deltas

1. **The "PLAYER · &lt;FACTION&gt;" eyebrow is gone** on every skin, its spectrum-ring dot with it — the dot was the line's bullet, not a mark of its own. The `ProfileKit.playerEyebrow` knob is deleted, so no kit can grow one back by accident.
2. **The "Unaffiliated · faction pending" caption is gone**, on both na branches. Its three i18n keys (`profile.playerFaction`, `profile.unaffiliated`, `profile.unaffiliatedPending`) had no other reader and go with it.
3. **`TaglineSlot` takes the name's place** — a shared block on #1626 `AboutBlock`'s pattern: hidden entirely when the tagline is blank, dressed by the caller through `style`, never re-implemented per faction. It owns its size and measure; the caller owns `margin` (the phone stack centres it).

**30px is `var(--text-heading)` (32px).** A tagline is read, so §4a's ornament hatch does not cover it and it takes a tier rather than a raw number — which is also what keeps `no-raw-style-values` green.

## Two kit knobs, one retired and one renamed

`nameSize` is **deleted** (7 kits): it was optical compensation for each archetype's display cut, and the slot it sized is gone. `nameExtra` is **renamed `taglineExtra`** (5 kits) rather than deleted — three of the five are ink CORRECTIONS for a header band that is a different stock from the kit's `surface` (#1636, Ephemerists' cornice and Everymen's banner), and dropping them would have put an ink measured for the panel back on the band. Per-faction tagline flair beyond that is #1630.

## One documented exception: WOW on a phone

`WowProfileBody` has its own phone header (#901) and it is the **one identity header with no credential card** — an avatar hoop over an `<h1>` name is the whole identity. Deleting the name there removes no repeat; it leaves a nameless profile. So that branch **keeps its name**, takes the shared tagline slot under it, and keeps its own motto eyebrow ("Of the Court of Whimsy"), which is the pavilion's copy and not the retired `playerEyebrow` slot. The test matrix encodes this carve-out explicitly rather than skipping the slug. Say the word and I will delete the name there too.

## Tests

Extended `__tests__/factionProfileBody.test.tsx` with a `describe.each` over **desktop × mobile**, each looping **all nine slugs** (na, albescent and the seven kits) — a wider matrix than `profileAbout.test.tsx`'s four branches, in the file that already sweeps every skin, so a new bespoke body is covered the moment its manifest row lands:

- every wording the deleted eyebrow knob ever produced is absent (`Player ·`, `PLAYER:`, `Practising ·`, `· the Court`);
- `faction pending` is absent everywhere, not just off na;
- the tagline renders and the slot's `max-width:22ch` fingerprint is present when set — and **absent** for `""` and for a whitespace-only draft;
- the display name appears on the credential card and in the praxis "sealed by" eyebrow, and nowhere else.

Two existing assertions that read the deleted copy were retargeted, not deleted (`factionProfileBody`'s na case, `profileFormFactor`'s desktop-na case). The #783 albescent guard is untouched and green.

## Verification

`tsc --noEmit`, `eslint src .ds-kit e2e`, `vitest run` (4134 tests, 232 files, all green), `vite build`, the initial-load budget (JS 127.9 KB, CSS 22.3 KB — both ok) and `typecheck:design-sync` all pass locally. `api/generated/schema.d.ts` is untouched. **No visual QA** — I have no browser or dev stack.

## What to eyeball

A profile **with no tagline** is expected to show a gap where the name was: credential card, level ring and progression bar, and nothing above the handle line. No character has written one yet, so that is what production will look like on day one. It is the design, not a missing placeholder.

- **Desktop:** na, albescent, coven, ua, snide, ephemerists, singularity, everymen, wow — with and without a tagline. Watch the two kits whose slot ink is a band correction: **Ephemerists** (cornice band) and **Everymen** (banner cream).
- **Phone:** the same nine. Three different headers to check — the na/albescent stack (tagline centred under the card), the six kits through `ProfileSkin`, and **WOW's pavilion**, which is the one that still shows a name and now stacks a 32px tagline under a 32px name. That pairing is the thing most likely to want a per-faction size, and that is #1630's call.
- **S.N.I.D.E.** inherits the old name treatment on its tagline — skew, uppercase, pink drop shadow — at 32px over up to 22 characters. In voice, but worth a look at a long tagline.
- Dark mode on all of the above; nothing here branches on theme.

## Deferred

The profile page now has **no `<h1>`** — the identity `<h1>` was the only one, and the card's name is a `<div>`. `CredentialCard` is under `components/`, outside this issue's file scope, so promoting the card's name to the page heading needs its own issue.


pixieofhugs:
## Two fixes on top of the header delta

**Seam for both:** the rendered profile-body markup — `FactionProfileBody` × faction slug × `useFormFactor()`. Both loops went red on the real defect before either fix existed (17 of 18 slug×form-factor cells had no `<h1>` at all; the WOW-phone About row failed all five of #1626's assertions), then green.

### ① The route lost its only `<h1>` — a11y regression this PR introduced

On `main` the only `<h1>`s on `/characters/:id` were exactly the three identity-column display names this PR deletes (`profileSkin.tsx`, `DefaultProfileBody`'s two branches, plus `WowProfileBody`'s phone header, which survives). The route mounts no `PageTitle`, and `CredentialCard`'s name is a `<div>` — so every profile except WOW-on-a-phone was shipping with **no top level to its document outline**, and a screen-reader user lost the "what page am I on" landmark.

`ProfileNameHeading` in `profileSkin.tsx` restores it **visually hidden**, so the display line the design deliberately removed stays removed — delta 3 is unchanged and still proven (see below).

- **Existing helper, not a new one.** I grepped for `sr-only` / `visually-hidden` / `.visuallyHidden` / a clip-rect utility. `index.css` defines none, but Tailwind's core `sr-only` is in the build and already live in `pages/updates/RequestsQueue.tsx`. Confirmed in the *built* CSS, not just the source: `.sr-only{position:absolute;width:1px;height:1px;…clip:rect(0,0,0,0)…}` is in `dist/assets/index-*.css`. Nothing hand-rolls a clip rect, so `no-raw-style-values` has nothing to trip on.
- **Shared layer only.** Mounted beside `TaglineSlot` — `ProfileSkin` (all seven kits, both widths) and `DefaultProfileBody`'s two branches. No archetype carries its own copy.
- **`components/` untouched.** `CredentialCard`'s name stays a `<div>`: that component mounts four times, including nine-up on the `/` roster and in the create-character preview, so promoting it would mint nine `<h1>`s on one page.
- **WOW's phone header gets nothing.** It has no credential card (#901) — its avatar hoop over a *visible* `<h1>` is the identity — so `MobileProfile` is simply a branch the shared slot does not reach. No conditional needed.

### ② WOW on a phone still rendered no bio — a gap #1626 left, not part of #1629

#1626 mounted `AboutBlock` in `profileSkin.tsx` and both `DefaultProfileBody` branches and counted that as every renderer. `WowProfileBody`'s bespoke `MobileProfile` (#901) routes through neither, so **a WOW player could write a bio in Settings and see it rendered nowhere.** The shared block is now imported there — not re-implemented; hidden-when-empty, `pre-wrap` and the content tier are all the block's own — dressed with `WowSectionHead` and the kit's body font/ink, matching how `DefaultProfileBody` passes its `heading`.

### Tests

Extended the matrices that already exist rather than starting a file.

`__tests__/factionProfileBody.test.tsx` — the 9 slugs × 2 form-factor loop:
- **exactly one `<h1>`, asserted as `toEqual(["Wren Aldercross"])` on the array of heading texts.** Count, not presence: presence alone would pass a WOW-phone double-heading, which is its own defect. The single assertion covers count, content and the WOW regression.
- `nameCount` now strips the `sr-only` heading before counting, so delta 3's "the name is on the credential card and nowhere in the identity column" still asserts about **visible** sites and is unchanged for all 18 cells.

`__tests__/profileAbout.test.tsx` — added `['the WOW pavilion phone stack', 'wow', 'mobile']` to the existing `BRANCHES` table. That is the seam #1626 was written at, and one row buys all five of its assertions: heading + bio present, bio read exactly once, line breaks survive, and the whole block hidden on each of `""` / whitespace-only / absent. Its header comment said "three renderers"; there are four.

### Verification

All six CI frontend steps run locally, all green: eslint · `tsc --noEmit` · `vitest run` (232 files, 4158 tests) · `vite build` · initial-load budget (JS 127.9 KB, CSS 22.3 KB — both `ok`; the `sr-only` rule was already in the bundle, so this adds nothing) · `typecheck:design-sync`.

**Visual QA is outstanding** — I have no browser or dev stack. Everything above is verified through markup, types and the built CSS.

### For `frontend-style`

One nit I deliberately did not decide: `AboutBlock` carries its own `var(--space-2xl)` top/bottom margins, and WOW's phone content column is a flex column with `gap: var(--space-xl)`. Flex-item margins don't collapse, so the About block sits on a wider gap than its neighbours in that one branch. Same shape as `DefaultProfileBody`'s phone stack, so it is consistent rather than novel — but it's a spacing call, not mine.

Issue #1638, opened 2026-08-03
Ephemerists surface dressing: vote plate, runes, credential card


pixieofhugs:
The per-surface half of the Ephemerists re-skin. Everything here sits on the foundation, so **do not start until the masthead, the sigil, the grid and the contrast pass have landed** — the masthead in particular changes what a card's header is.

Design: https://claude.ai/design/p/9305f548-25ce-4dac-9d35-10867994831f

These are independent of each other and can land one at a time. Split them into separate PRs as they are picked up.

## The deltas

**The vote plate loses its scaffolding.** Three removals, all with the same reasoning — the metals *are* the scale, so nothing needs to restate it:
- the dashed track behind the metals
- the numeral pips on each metal
- the italic gloss caption above the vote row

**The vote plate gains a per-tier burst.** The component draws a fixed 10 rays below the top tier; the design redraws the burst as a conic ring whose spoke step is set per metal, so it densifies as the metal improves: lead 60°, copper 45°, silver 36°, gold 30°, platinum 22.5°, each inked in its own metal token. Masked to a ring (`radial-gradient(closest-side, transparent 0 82%, #000 84%)`) and animated on a 3.2s ease-in-out loop. **The animation needs a `prefers-reduced-motion` guard** — the repo's pattern is at `pages/editPraxis/archetypes/shared.tsx:182-185`.

**The vote plate's frame survives its own clip.** *"The stepped clip-path shaves the plate's own brass border off."* Keep the stepped silhouette; draw the frame as an inset ring, which the clip follows.

**"Who voted" becomes the struck metal.** Instead of a name and a Roman numeral, each voter is the metal's own glyph struck in a 28px brass-ringed disc, inked in plate ochre.

> **Flag before building:** the design hides both the metal's name and the numeral, leaving only a `title` attribute. A `title` on a span is not reliably announced by screen readers and is unreachable on touch. Whatever ships must leave the metal's name available to assistive tech — a visually-hidden label, not a tooltip. The visual idea is good; its accessible name needs designing.

**Fluted rules become shifting runes.** *"FlutedRule → shifting runes, on every surface that draws it."* On the task card the neighbouring papyrus band already carries them, so the card's own rule is dropped rather than converted.

**Edit praxis pairs its rune band with the brass rule** — *"the rune band reads as a loose row on its own — pair it with the brass rule, the way the write-up header carries it."*

**The sign-up mark becomes the platinum glyph** rather than the eye.

[...]

Issue #1633, opened 2026-08-03
Two Ephemerists card bugs: a clipped byline and colliding vote bursts


pixieofhugs:
Two live defects on Ephemerists surfaces. The *Ephemerists Kit* design fixes both in passing, but neither is a re-skin — they are broken now, and they should not wait behind the epic.

Design: https://claude.ai/design/p/9305f548-25ce-4dac-9d35-10867994831f

## 1. The praxis card's byline clips a letter behind its own avatar

The design's comment: *"The byline name was clipping a letter behind its own avatar."* Its fix is `overflow: visible` and `text-overflow: clip` on the byline link.

The design also notes an adjacent cause it fixes separately: *"The card's flex-basis is a WIDTH; in a column stack it became a height cap and clipped the byline and vote row."* — `flex: 0 0 auto` on the card.

Those are two different clips. **Work out which one actually bites in the shipped layout before writing either fix** — the design is a canvas stacking cards in a column, which is not necessarily how they sit in the app. If the flex-basis is behaving in production, only the byline fix is real; if it is not, the byline fix may be treating a symptom of the basis bug. Fix the cause, not both.

Reproduce on a praxis card whose author name is long enough to reach the ellipsis.

## 2. The vote metals' bursts and sheens collide

The design's comment, verbatim: *"the metals' bursts and sheens were colliding at the default gap"*. Its fix widens the vote plate: `gap: 22px` with `padding-left` and `padding-right` to match.

Take the spacing change only. The design *also* redraws each metal's burst as a conic ring whose spoke density scales with the tier — that is dressing, and it belongs to the epic's surface queue, not here.

## Do not take literally

The design selects by inline-style substring (`div[style*="praxis-card-basis"]`) because a canvas cannot reach the components. Find the real elements and fix them at source.

Both fixes are `frontend-style` work. Neither should need a component change.

PR #1660, opened 2026-08-04
The Ephemerists vote plate drops its scaffolding for a per-tier burst, and the fluted rules become runes

pixieofhugs:
Part of #1638

Five of the issue's seven bullets. The credential card was already done (#1651, confirmed on `main`), and the **cavetto cornice is deliberately left** — see *What I did not build*.

## The seams

- **The vote plate's markup** (`ephemeristsVote.test.tsx`) — the scaffolding removals, the per-tier spoke step, the pair of nested clips. Each of the three removals fails silently on its own: a plate that keeps the dashed rail still renders, still votes, and still looks deliberate.
- **The stylesheet, partitioned by brace-counting** — the burst's 3.2s loop is perpetual, so its `prefers-reduced-motion` guard is a floor, and a component cannot be asked about a guard that lives in CSS. A dropped `@media` opener balances every brace and still builds. Also verified in the **built** stylesheet, not just the source (27 guards; the burst's `animation` inside one, its ring and mask outside).
- **The source tree** (`ephemeristsPlateSurfaces.test.tsx`) — the rune sweep is grepped, not rendered. `FlutedRule` had six mounts reachable by following imports and a **seventh hand-copied declaration** in `EphemeristsTaskDetail`, so an import-graph sweep converts six surfaces, leaves two mounts on the old drawing, and every test stays green. The check pins zero `FlutedRule` and exactly seven `RuneRule` files.
- **The voters row's accessible name** (`ephemeristsDetail.test.tsx`) — asserting the metal's word reaches *text* is what tells `sr-only` apart from the design's `title`: an attribute is not text content, so a `title`-only build renders identically and says nothing to a screen reader.

## What changed

**The vote plate loses its scaffolding.** The dashed rail, the roman numeral struck on each disc, and the italic gloss caption — three restatements of the row they sat on. `votes:chrome.ephemerists.idle` and `.tag` were deleted with the caption rather than left for a dead-key sweep to puzzle over. Nothing is lost: each disc's `aria-label` names its metal and `VoteSummary` states the cast.

**The per-tier burst is what earns that.** The ray fan becomes a conic ring whose spoke pitch is `METAL_SIGILS[n].burstStep` — 60/45/36/30/22.5 degrees, six spokes at lead to sixteen at platinum — so the burst densifies up the ladder and carries the rank the numeral used to. Each ring is inked in its own metal. Pigment, ring mask and cycle are `.eph-metal-burst` in `index.css`; only geometry arrives inline. `@keyframes eph-metal-ray`, `.eph-metal-rail` and `eph-metal-current` went with the fan.

**#1633's gap arithmetic is re-derived, not left stale.** Its figures were trigonometry over a fan — which spoke came nearest horizontal, since one pointing straight up costs a neighbour nothing. A ring has no nearest-horizontal spoke: it reaches its full radius in every direction, so the projection is just the overhang, `BURST_MARGIN / 2` = 12px, **and it is identical at every tier** because the ring box grows with the disc. Two rings want 12 + 12 = 24px rim-to-rim, which is `--space-xl` exactly rather than the 0.36px-short fit the fan had. `BURST_MARGIN` drops 30 to 24 to make that true. The block padding opens to `--space-lg`, which #1633 explicitly deferred to this issue: the fan's index-0 ray reached 14.5px into 12px of padding and the clip cut it; the ring reaches 12 and 16 clears it by 4.

**The frame survives its own clip.** Confirmed: `clipPath: stepClip(7)` chamfers two corners and a `border` painted at the border box is cut away with them — the plate shipped with two open corners. The brass is now a **ground** with the night sheet laid 1px inside it and stepped a pixel tighter, so the clip carries the rule instead of shaving it. (An inset box-shadow fails identically; the inner chamfer leg is 6, not 7, or the two cut corners land on each other and the frame opens again.)

**"Who voted" becomes the struck metal.** Each voter row ends in the sigil for the metal that voter cast, in a 28px brass-ringed disc inked in plate ochre — the same drawing the plate strikes, out of the same `METAL_SIGILS`, so the two cannot diverge.

**The accessible name is `sr-only`, not `title` — a deliberate pick, not a default.** `GlossedGlyph` + `.eph-gloss` (#1656) is right for a *decodable* glyph whose English is a bonus **already printed elsewhere on screen**; the tooltip is a convenience over a fact the reader can otherwise obtain, and the visually-hidden twin was rejected there because it would be a second copy free to drift. Here the metal's name is the **only** identifying information in the row — nothing prints it — so hiding it behind a pointer affordance leaves a screen-reader user a row that reads a name and stops, and leaves a touch user with no route to it at all. Tailwind's `sr-only` is in the bundle and used elsewhere. `reframeLabel` still produces the word; only its rendering changed.

**Fluted rules become shifting runes, on all seven surfaces.** `FlutedRule` becomes `RuneRule`: the plate's own register of incised signs replaces 48 brass strokes that were the cornice drawn a second time. It shifts on `.epg-glyph`, the register's existing cycle, already gated — so **this adds no CSS and no ungated motion**. Struck in `-brass`, not the register's masthead gold, which on papyrus is a stain rather than a rule. The task **card**'s `JournalRule` is dropped rather than converted (its masthead already marches two registers of the same signs); `ruleBleed` went with it. `EphemeristsTaskDetail`'s hand-copied declaration is collapsed into the shared one.

**Edit praxis pairs its band with the brass rule** (`<RuneRule rule />`) — the only mount that does, because the other six sit under a section head whose filler rule already draws that line.

**The sign-up mark becomes the platinum glyph** on both summonses (task card, task page), read from `METAL_SIGILS` rather than copied. The open eye had no other reader on the card and went with it.

**The rune band draws without a `viewBox`.** `slice` scales by `max(w/vbW, h/vbH)`, and with a fixed height that is 1 — so any column wider than the viewBox blew every sign up (a 16px band of 32px glyphs at the 1200px page cap), which is the exact failure `EphemeristsTaskDetail` documented when it stopped using a fixed sign count. Without a viewBox, user units are CSS pixels and the SVG viewport crops the overspill.

## What I did not build

**The cavetto cornice retirement.** The bullet is compound and has already proved half-stale once (the winged sun disc half is #1656's, and `WingedDiscSign` is a deliberate keep for the metatask seal — I did not touch it). The remaining half is bigger than the issue reads: `Cornice` is mounted on **seven** surfaces, not the two named — `FactionCard`, `EphemeristsFactionHero`, `EphemeristsFeedFrame`, `EphemeristsPraxisCard`, `EphemeristsEditPraxis`, `EphemeristsPraxisDetail`, plus a local copy in `EphemeristsTaskCard` — and it carries `.eph-cornice-glow`, which the praxis record's own docstring calls its one piece of motion. Retiring it is a one-line delete on each mount and then 100% eyeball work across seven surfaces at two themes, and I have no browser. It wants its own PR with a visual pass, and none of the design's stated rationale ("the sigil is the only mark") is about the cornice.

## Bundle budget — over the warn line by 21 bytes, and the headroom was 3

Measured with the script's own `gzipSync(level: 9)`:

| | gzipped CSS |
|---|---|
| `origin/main` | **22,997** |
| this branch | **23,021** |
| warn / fail | 23,000 / 25,000 |

The net cost of the whole burst — a new class, its keyframe, its guard, minus three retired rules — is **+24 bytes**, and `main` was already **3 bytes** under the line. Any CSS at all trips this warn today. I shaved what was free (the repo's `mask` shorthand, dead `var()` fallbacks); the remaining 21 bytes are the ring's mask and its reduced-motion resting state, and I am not cutting either to squeak under. FAIL at 25,000 is untouched.

One free follow-up spotted, not taken here: **autoprefixer already emits `-webkit-mask`**, verified by building without it — so the seven hand-written `-webkit-mask` / `mask` pairs in `index.css` are redundant. It measured at 0 bytes gzipped (gzip already deduplicates them), so it is tidiness rather than budget, and it touches seven unrelated ornament rules.

## Verification

`tsc --noEmit` clean, `eslint src` clean, `vitest run` 233 files / 4192 tests green, `vite build` clean, `typecheck:design-sync` clean (the vote preview's prop shape is unchanged), budget WARN as above.

**Visual QA is outstanding, and it is most of this change.** I have no browser and no dev stack. Five metal tiers, a new animation and seven rune surfaces are almost entirely eyeball work; everything above is markup, source and stylesheet assertions.

## What to eyeball

**The vote plate** (`/praxis/:id` on an Ephemerists task, signed in and able to vote)

1. Each metal in turn, hovering 1 to 5: the conic ring should visibly densify — six spokes at lead, sixteen at platinum — and each ring should be its own metal, not one accent.
2. The plate's **top-left and bottom-right chamfers**: the brass rule should run all the way round the stepped silhouette, with no gap at either cut corner. This is the bug being fixed; check it first.
3. The ring's clearance: 12px past each rim, so neighbouring rings should touch but not overlap, and the top and bottom of each ring should clear the plate edge by ~4px rather than being clipped.
4. That nothing looks *missing* where the rail, the numerals and the caption were — the row should read as a scale on its own.
5. Both themes. The plate is a fixed night surface in both, so it should look identical; the page around it flips.
6. **Reduced motion** (OS setting, or DevTools, Rendering, Emulate `prefers-reduced-motion`): every ring should stay fully drawn at its own pitch, just not breathing. A blank halo there is the failure.

**The struck metal** — the voters panel on the same page: a 28px brass-ringed ochre sigil at the end of each row, and with a screen reader the row should read "*name*, silver".

**The seven rune surfaces** — each should show a row of incised signs in brass where a fluted strip used to be, at the same size on a phone and at the 1200px page cap:

1. `EphemeristsComment` — the comment voice's rule
2. `EphemeristsFactionBody` — desktop faction page
3. `EphemeristsFactionPage` — mobile faction page
4. `EphemeristsFieldDesk` — mobile field desk
5. `EphemeristsPraxisDetail` — under every section head
6. `EphemeristsEditPraxis` — the composer, **and this one alone should carry a brass hairline above the band**
7. `EphemeristsTaskDetail` — twice: the brief's section head and the gallery head (this is the ex-duplicate)

**And the two that are removals** — the task **card** should have *no* rule between its description and its in-progress line (the masthead's registers carry the signs), and both sign-up buttons (task card, task page) should show the platinum sigil instead of the open eye.


pixieofhugs:
**Held by owner ruling, not by anything wrong with this PR.**

CI is green and the work is right. The problem is upstream of it: `origin/main` measures **22,997 bytes** gzipped CSS against a **23,000** warn threshold, so main already had three bytes of headroom before this branch existed. This PR's +24 is a fair price for the burst; there is simply nowhere to put it.

The owner ruled: **shed CSS rather than merge over the line or move the line.** Filed as #1661 and dispatched now.

Nothing to change here. When #1661 lands, rebase and re-run `npm run budget` — the +24 will fit and this merges as-is. Do not shave the ring mask or the reduced-motion resting state to squeak under in the meantime; declining to do that was the right call.

Also noted from this PR and carried into #1661: the seven hand-written `-webkit-mask`/`mask` pairs are genuinely redundant but measure **0 bytes gzipped**, so they are tidiness rather than headroom.

Issue #1727, opened 2026-08-15
Narrow what contrast.spec.ts scans, so it asserts only what it can measure


pixieofhugs:
Owner ruling 2026-08-14 on #1675: option (1) now, **option (3) as the real answer**. This is option (3).

## Blocked by #1675

#1675 makes unmeasurable backdrops report-instead-of-fail, which retires the 26 failures. That is the interim. This issue is the fix that makes the suite meaningful again, and should land after it.

## The problem #1675 diagnoses

`contrastScan.ts` returns `background: null` when it cannot resolve what is behind text -- a gradient with an opaque stop, or an image. World Zero's faction skins are built out of exactly those, so the scanner cannot measure most of what it walks. #1675 lists the evidence that this is a harness limit rather than 26 style bugs: it fails for **every** faction, **both** themes, **both** viewports, it has never passed, and the 2026-08-01 prod visual audit found zero real defects.

## What to build

**Assert contrast only on surfaces that are supposed to be flat**, and exempt decorative fills by construction rather than by a growing exception list.

- Scan: body copy, cards, chrome, form controls, navigation -- anywhere text sits on a solid token.
- Exempt: text over a gradient, an image, or any backdrop the scanner resolves to `null` **because the design intends it to be decorative**.
- The exemption must be **structural, not a list of selectors that rots.** Prefer a marker the surface itself declares (a data attribute or a class the decorative frames already carry) over a hand-maintained array in the spec.

`contrastBaseline.ts` holds the grandfathered-exception list and is the shape any new exemption should follow -- read it before inventing a second mechanism.

## The thing that makes this worth doing properly

Two independent measurements during the 2026-08-14 grill both got the wrong answer by **one compositing layer**, and both were caught only by hand:

- the profile header composites `--color-bg-surface-alt` (a translucent white) over the page -- #1715
- a selected filter row composites `--filter-thumb` over `--filter-well` over the page -- #1579

Both are **flat, resolvable grounds** once you composite correctly. So the scanner should be able to measure them, and a narrowed scan must not exempt them by accident -- translucent-over-solid is exactly the case worth keeping, and it is where the real defects were found. Use `frontend/src/utils/contrast.ts`'s `compositeOver` rather than treating a translucent token as opaque.

## Done when

- The suite asserts contrast on flat surfaces and does not attempt decorative ones.
- Translucent-over-solid grounds are still measured, correctly composited.
[...]

Issue #1715, opened 2026-08-15
The dark muted text tokens sit on the AA floor rather than clearing it


pixieofhugs:
Owner-reported from a mobile profile screenshot, then measured.

## The measurement

The first measurement used the page ground and looked fine. **That was the wrong ground.** The profile draws its panels on `--color-bg-surface-alt`, which in dark is `rgba(255, 255, 255, 0.06)` over `#13121a` -- compositing to **`#212028`**. Against the surface the text actually sits on:

| token | value | on `#13121a` | on `#212028` (the real card) |
|---|---|---|---|
| `--color-text-tertiary` | `#9490c0` | 6.21:1 | **5.38:1** |
| `--color-text-secondary` | `#988c78` | 5.63:1 | **4.88:1** |

Both merely clear the 4.5:1 AA floor -- `secondary` by 0.38. AA is a floor, not a target, and the muted tier is resting on it.

Note `secondary` is the worse of the two even though `tertiary` is the one that looked wrong: it carries `23 / 60 pts this level` on the profile.

## What to change

Lift **both** dark tokens until they clear **AAA 7:1 on `#212028`**. For reference, `--color-text-tertiary` at `#b3b0d6` measures 7.76:1; treat that as a worked example, not a mandate -- keep the hue family.

**Then re-audit the light tokens the same way**: measure against `--color-bg-surface-alt` (`#f0ede6`), not `--color-bg-page` (`#f7f4ee`), and lift them to the same target if they fall short. Light is now the opt-in theme but it is still shipped.

## Do it through the global tokens

Precedent: #669 / #677 lifted dark-mode chrome luminance **through the global text tokens, never per-faction repoints**. Same here. Edit `index.css` and let the cascade carry it.

Related but **not** this issue:
- **#1608** -- a third of the label tier is a local `CSSProperties` object, so 9-10px labels ship regardless of colour. Real, separate, still open.
- **#1675** -- `contrast.spec.ts` failing 26 of 32 E2E tests is a *harness* problem (the scanner cannot resolve gradient backdrops). Do not try to fix it here.

## Watch for

- These tokens are used site-wide. Check a few faction surfaces in both themes -- a faction card that assumed a dimmer muted tone may now read hot.
- Do not introduce per-faction overrides to compensate. If one kit genuinely breaks, say so in the PR rather than patching around the global.
- No `opacity` on text as a substitute -- opacity multiplies against whatever is behind and drifts per surface.

## Done when

[...]

PR #1752, opened 2026-08-15
An unmeasurable backdrop gets reported and ratcheted instead of failing 28 tests

### LINKED ISSUE #1675: contrast.spec.ts fails for every faction, theme and viewport — 26 of the 32 E2E failures
`contrast.spec.ts` accounts for **26 of the 32 failures** in the nightly E2E run. Fixing this one spec retires ~80% of the red.

## What happens

`frontend/e2e/contrast.spec.ts:113` generates one test per faction × theme × viewport — 7 × 2 × 2 = 28 tests. Essentially all of them fail with:

> `ua/light/mobile: text below WCAG AA` — *An UNRESOLVED BACKDROP is a failure, not a skip — text over an opaque-stop gradient or an image cannot be measured, so it must be given a solid backdrop (or the fill hoisted behind an opaque card).*

## Why this reads as a harness problem, not 26 style bugs

- It fails for **every** faction, **both** themes, **both** form factors. Real contrast defects cluster; they do not distribute perfectly across every combination.
- The **prod visual audit on 2026-08-01 found zero contrast defects** on the live site.
- The failure text is the scanner reporting it *could not measure*, not that it measured and got a bad ratio. `finding.background === null` is a distinct state from a low ratio — see the `ratio` handling at `contrast.spec.ts:105`.
- It has never passed. This is not a regression that crept in; the strictness was never satisfiable.

The scanner cannot resolve what is behind text when the backdrop is a gradient with an opaque stop, or an image. World Zero's faction skins are built out of exactly those.

## The decision to make first

**This needs an owner ruling before any code.** Three options, and they are genuinely different products:

1. **Unmeasurable → skip, but reported.** The test passes, and the run prints a list of surfaces it could not measure so the gap stays visible. Keeps the suite honest without asserting something it cannot check. Retires the 26 failures.
2. **Unmeasurable → still a failure, and the styles change.** Every measured surface gets a solid backdrop or the fill is hoisted behind an opaque card. Strongest accessibility guarantee, largest diff, and it constrains the faction skins — which are the point of the design system.
3. **Narrow what is scanned.** Only assert contrast on surfaces that are *supposed* to be flat (body copy, cards, chrome), and exempt decorative gradient fills by construction.

My read: **(1) for now, (3) as the real answer.** (2) fights the design system, and the live site already measures clean by hand.

## Where the code is

- `frontend/e2e/contrast.spec.ts` — the parameterized test, the failure message at ~line 153
- `frontend/e2e/contrastScan.ts` — the scanner that returns `background: null` when it cannot resolve
- `frontend/e2e/contrastBaseline.ts` — the grandfathered-exception list, and the shape any new exemption should take

Related history worth reading before changing the policy: #651 (the contrast family), #669/#677 (dark-mode chrome contrast).

## Explicitly NOT this issue

- The six functional spec failures — sibling issue.
- Changing any faction's colours to satisfy the scanner. If option (2) is chosen, that is its own piece of work with design review, not a byproduct of a test fix.
- Putting e2e on the PR-blocking path.

## Done when

A nightly run where `contrast.spec.ts` contributes **zero** failures, and — whichever option is chosen — the run output still makes visible which surfaces went unchecked. A silent pass is not the goal; an honest one is.

pixieofhugs:
Closes #1675

Option (1) per the owner ruling of 2026-08-14: a backdrop the scanner cannot resolve stops failing the suite and gets **reported** instead, loudly, with a count ratchet so gaining one still fails.

## The seam

`triageFindings(theme, findings)` — the point where one scanner `Finding` becomes pass / report / fail. That decision used to be inline in the spec body, reachable only from a nightly run against Playwright + a live backend + a seeded Postgres. It is now a pure function in `contrastBaseline.ts`, and `src/utils/__tests__/contrastTriage.test.ts` drives it from a PR.

Red-first: making `background === null` fall through to the ratio comparison (i.e. the old behaviour) fails 3 of the 8 new tests, and only those 3.

## What changed

**Three outcomes, not two.**

- a **measured** pairing below AA still fails, unchanged, allowlist and all;
- `background === null` — the scanner *refusing* to measure, because what is behind the glyphs is a gradient with an opaque stop — is collected, printed, and asserted only by count;
- an allowlisted pair that now clears AA still fails. The list only ever shrinks.

**The report prints on every run, before any assertion, pass or fail:**

```
[contrast] wow/dark/desktop: 8 unmeasurable surface(s) (ceiling 8), 42 text node(s) unchecked.
  12 node(s) over linear-gradient(160deg, rgb(55, 46, 37), rgb(36, 30, 24) 60%)
    e.g. div > div > div > div (42px/600) "UA"
  ...
```

**The ratchet** is `UNRESOLVED_SURFACE_CEILING` in `contrastBaseline.ts` — distinct `backdropCss` values per faction/theme/viewport. A combination with no entry has a ceiling of 0, so adding a faction or a viewport forces you to record what it costs in coverage.

Why a count and not an allowlist of surfaces: that is what the file *used* to do, and it rotted. The gradient CSS embeds its colour stops, so every palette edit minted a fresh key — by 2026-08-14 **not one of the 56 grandfathered UNRESOLVED entries matched anything the sweep still saw.** They are deleted here. A count survives a restop and still catches the thing that matters, which is a new region of the app going unchecked.

## Where the ceiling numbers come from — and the thing this PR does NOT fix

I cannot run this suite: a subagent worktree has no Playwright browsers and no seeded Postgres. So instead of guessing, I read the numbers off the failing nightly itself, run `31779247838` (2026-08-14), and cross-checked every one against Playwright's own `Received length:` line. All 28 tests reconcile exactly, and both retry blocks agree.

That same parse turned up something the issue does not say, and it matters for the acceptance criteria:

| | findings | distinct |
|---|---|---|
| unmeasurable backdrops | 728 | 21 surfaces |
| **genuinely measured, below AA** | **168** | **23 colour pairs** |

**Every one of the 28 tests carries between 2 and 15 measured sub-AA pairs.** The issue reads the `UNRESOLVED BACKDROP is a failure, not a skip` wording as evidence that the failures are all unresolved-backdrop, but that sentence was a **constant in the assertion message**, printed for every failure including pure ratio ones.

So this PR removes ~81% of the noise (728 of 896 findings) but **will not turn contrast.spec.ts green.** The residual 23 pairs are genuine measured contrast debt, and the ruling is explicit that a measured low ratio must still fail. They are also not random: the top of the list is ghosted body text —

```
22x  light | rgba(26, 18, 9, 0.45)   on rgb(255, 253, 249) @4.5
22x  dark  | rgba(240, 230, 208, 0.4) on rgb(28, 27, 36)   @4.5
14x  light | rgba(107, 96, 80, 0.75) on rgb(255, 253, 249) @4.5
14x  dark  | rgba(155, 144, 128, 0.75) on rgb(28, 27, 36)  @4.5
```

— i.e. `--color-text-secondary` / `--color-text-tertiary`, which **#1715 is lifting in this same batch.** Worth re-reading the next nightly before filing anything: a chunk of the 23 may retire on their own. Separately, none of the 185 measured baseline entries matched anything in this run either, and zero "now measures" reports fired, so that list is stale too — out of scope here, and it does not affect this change.

## Verification

Everything `.github/workflows/test.yml` names for the `frontend` job, locally, all green: `lint`, `typecheck`, `typecheck:design-sync`, `test` (237 files / 4258 tests), `build`, `budget` (JS 129.0 KB, CSS 21.5 KB, both ok). `npm run typecheck` only includes `src`, so I also ran `tsc --noEmit` over the three `e2e/` modules directly — clean. No backend files touched, so `api-schema` and the backend job are unaffected.

**The only real verification is the next nightly.** I did not run the e2e suite and am not claiming it passes.

## What to eyeball

- **The ceiling numbers** in `contrastBaseline.ts`. They are exact for run `31779247838`, but that log truncates each gradient to 80 characters. Two surfaces agreeing for 80 characters would have counted as one and left a ceiling one low. If the first nightly reports a higher count for some combination, bump *that* entry to the number the run prints — please don't raise them speculatively.
- **Deleting the 56 UNRESOLVED entries.** The argument is that they were provably matching nothing; if you would rather keep them as a historical record, say so and I will restore them as a comment.
- **Count vs. set for the ratchet.** A count will not notice one gradient surface being swapped for a different one. That is the trade for surviving a palette edit — and #1727 (option 3, narrowing what is scanned) is the real answer either way.
- **`console.log` for the report.** It lands in the nightly job log. If you want it as an artifact instead, that is a different mechanism and I would rather do it deliberately.
- **The ceiling assertion is last**, after the measured-failure assertions, so while the 23 measured pairs are still red the ceiling never actually gets exercised. The report still prints on every run regardless — that was the non-negotiable part.

### DIFF
--- a/frontend/e2e/contrast.spec.ts
+++ b/frontend/e2e/contrast.spec.ts
@@ -30,9 +30,11 @@ import { scanPageForContrast, type Finding } from './contrastScan'
  * red from the day the guard landed. See contrastBaseline.ts Part D.
  *
  * REGENERATING THE BASELINE. Set `CONTRAST_BASELINE_OUT=<path>` and run the
- * suite; every failure is appended there as a ready-to-paste entry. The list
- * is machine-produced on purpose — hand-typed ratios would be wrong within a
- * week, which is the whole thesis of this issue.
+ * suite; every MEASURED failure is appended there as a ready-to-paste entry.
+ * The list is machine-produced on purpose — hand-typed ratios would be wrong
+ * within a week, which is the whole thesis of this issue. Unmeasurable
+ * backdrops are not baseline entries any more, so nothing is emitted for them;
+ * their ceiling comes off the report this prints on every run.
  */
 
 const API = process.env.E2E_API_URL ?? 'http://localhost:8000'
--- a/frontend/e2e/contrast.spec.ts
+++ b/frontend/e2e/contrast.spec.ts
@@ -1,7 +1,7 @@
 import { test, expect, type Page } from '@playwright/test'
 import { appendFileSync } from 'node:fs'
 
-import { RENDERED_BASELINE, baselineKey, unresolvedKey, type BaselineEntry } from './contrastBaseline'
+import { UNRESOLVED_SURFACE_CEILING, baselineKey, triageFindings } from './contrastBaseline'
 import { scanPageForContrast, type Finding } from './contrastScan'
 
 /**
@@ -21,6 +21,13 @@ import { scanPageForContrast, type Finding } from './contrastScan'
  *
  * Nightly, not per-PR: `.github/workflows/e2e.yml` already stands up Postgres
  * + backend + seed + Playwright. No new CI job.
+ *
+ * THREE OUTCOMES, NOT TWO (#1675). A measured pairing below AA fails. A
+ * backdrop the scanner REFUSES to measure — a gradient with an opaque stop, an
+ * image — is reported instead, and only its COUNT is asserted, against
+ * `UNRESOLVED_SURFACE_CEILING`. Failing on those made the spec unsatisfiable:
+ * the faction skins are built out of exactly those fills, so all 28 tests were
+ * red from the day the guard landed. See contrastBaseline.ts Part D.
  *
  * REGENERATING THE BASELINE. Set `CONTRAST_BASELINE_OUT=<path>` and run the
  * suite; every failure is appended there as a ready-to-paste entry. The list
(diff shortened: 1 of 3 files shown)


pixieofhugs:

**Closing as superseded by #1749**, which merged as `a78dcbcb`.

Two sessions worked #1675 concurrently without knowing about each other. #1749 is the more complete answer and it landed first, so this one goes.

## Worth recording: the two agreed independently

Both PRs parsed the same nightly run (`31779247838`) from scratch and reached the same conclusion, which is much stronger evidence than either alone:

- **The issue's headline premise was wrong.** The even spread across every faction, theme and viewport is not proof of a harness bug — it is shared chrome failing identically everywhere.
- **Both counted 23 distinct measured sub-AA defects.** Both found that every one of the 28 tests carries at least one, so option (1) alone retires zero failures and the issue's "Done when: zero failures" was never reachable through it.

This PR also found the `An UNRESOLVED BACKDROP is a failure, not a skip` sentence the issue quotes as evidence is a **constant in the assertion message**, printed for every failure including pure ratio ones — so it was never evidence about what failed. That is why the original diagnosis went the way it did.

#1749 acted on the same finding by fixing 19 of the 23; this PR stopped at the policy half. #1749 is the better outcome.

## The one thing worth salvaging

This PR extracted the triage into a **pure function** (`triageFindings(theme, findings)`) specifically so it could be exercised in the repo's DOM-less vitest harness, and left 8 unit tests behind. #1749 states plainly that it could not unit-test its scanner change for exactly that reason.

So `contrast.spec.ts` currently has logic whose only verification is a nightly run. Extracting the merged triage behind a pure seam would give it real coverage in seconds rather than hours. Not urgent, and not worth reverting anything for — noted here so the idea is not lost with this branch.

Branch left for inspection rather than deleted, in case that extraction is picked up.

Issue #1743, opened 2026-08-15
Retire the debounced PUT: the room becomes the only way a praxis body is written


pixieofhugs:
**Blocked by #1742** — the composer must be writing into the room before the old save path can be removed, or there is nowhere to persist.

Part of the ADR-0073 epic (Google-Docs-style co-editing of a praxis). Design and rejected alternatives: `docs/adr/0073-a-praxis-is-written-in-a-room-and-the-room-is-not-the-record.md`. Self-contained below.

This is the issue that makes the room the *only* way a praxis body is written.

## The rule

**One write path, for every praxis type.** Rooms serve solo, collab and duel parts alike, and the debounced PUT is **deleted, not kept as a fallback**.

A fallback is the second write path wearing a different hat: it would need a reconciliation rule for a PUT landing behind a room's document, and every future rule about saving would have to be stated in both places. That is the defect class #1692 shipped three production bugs from — a rule stated twice and never reconciled.

## Scope

**`body_text` becomes a derived column.** The server flattens the room's document to `praxis.body_text` on debounce and on freeze. It stays markdown, so no read surface changes — praxis detail, the feed, search, `react-markdown` and the OpenAPI contract are all untouched by this epic, deliberately.

**Delete the client save machinery** in `frontend/src/pages/editPraxis/useComposerDraft.ts`: the 2s debounce, `flushEdits`' cancel-then-write ordering, `hasUnsavedEdits`, the last-persisted refs, and the `isDirty()`/`needsTitle()` seam. All of it exists to make single-writer autosave safe (#360 / #1081 / #1164) and none of it has meaning once the room owns persistence. Delete the tests that only covered that ordering; keep any that assert composer state the room still produces.

Note the knock-on: `hasUnsavedEdits` exists specifically because a PUT hard-resets every member's `has_submitted` (ADR-0012), so Submit had to avoid firing one. With the PUT gone, that whole workaround goes with it — and the freeze issue that follows replaces the rule properly.

**Offline authoring moves to `y-indexeddb`**, syncing on reconnect. This is strictly better than what it replaces: today's PUT simply fails when the network is down.

**Update the API surface.** `PUT /praxes/{id}` loses title/body (or goes entirely, if nothing else uses it — check before deleting). Regenerate the OpenAPI types.

## Out of scope

Freeze-on-pending and the `pullBack` reopen path (next issue). Presence.

## Verification

Backend, in the existing pytest-against-Postgres suite:

[...]

Issue #590, opened 2026-07-16
Collab consensus: promote pending to a real status, unify withdraw/unsubmit


pixieofhugs:
## Problem Statement

When a collab is mid-consensus (some members have submitted, not all), the only way to tell that apart from a fresh, untouched draft is to check whether `submit_proposed_at` is set — there's no first-class status for it. That makes "is this collab pending?" a derived fact scattered across every consumer instead of a queryable state.

Separately, there's no way for a member to pull back just their own submission while the group is still waiting on others: the only existing undo, `withdraw_praxis` (`POST /praxes/{id}/withdraw`), explicitly 422s unless the praxis is already fully `submitted`, and even then resets every member's `has_submitted`, not just the caller's. The naming has also drifted — the codebase already uses "withdraw," "reopen," and (informally) "unsubmit" for what is conceptually the same "undo my submission" action.

## Solution

Add `PraxisStatus.pending` as a real third enum value (`in_progress` → `pending` → `submitted`), set by `collab_consensus.on_submit` whenever a submit doesn't complete consensus. Rename `withdraw_praxis`/`POST /praxes/{id}/withdraw` to `unsubmit_praxis`/`POST /praxes/{id}/unsubmit`, and widen it to handle two cases instead of one: reopening a fully `submitted` collab (today's behavior, unchanged — resets every member) and, new, pulling back an individual member's own cast while the collab is still `pending` (resets only the caller; drops the praxis back to `in_progress` if that was the last remaining cast, stays `pending` otherwise).

`submit`/`unsubmit`/`has_submitted` remain the only code-level verbs for this concept — "cast," "file," "seal," etc. are per-faction display copy layered on top in a separate frontend issue, not backend vocabulary.

This repo is pre-launch with a disposable DB (`docs/agents/db-migrations.md`, Strategy A — the single baseline migration rebuilds the schema from the live models via `Base.metadata.create_all`), so adding the enum value is a model edit + a dev DB reset, not an `ALTER TYPE` migration.

## Commits

1. **Add the enum value.** `PraxisStatus.pending = "pending"` in `backend/models/praxis.py`. Nothing sets or reads it yet — pure no-op commit. Reset local/worktree dev DBs after (`backend/scripts/reset_db.sh`).
[...]

PR #1803, opened 2026-08-15
Freeze on pending, pullBack to reopen, and discard the document on publish

### LINKED ISSUE #1745: Freeze on pending, pullBack to reopen, and discard the document on publish
**Blocked by #1743** — the room must own persistence before the consensus rule can be rewritten around it.

Part of the ADR-0073 epic (Google-Docs-style co-editing of a praxis). Design and rejected alternatives: `docs/adr/0073-a-praxis-is-written-in-a-room-and-the-room-is-not-the-record.md`. Self-contained below.

## The problem this solves

ADR-0012 says any member's edit hard-resets everyone's `has_submitted` — "an edit means we're not done." That rule keys on an edit being a **discrete event**. A CRDT has no discrete event; text simply moves. Left alone, either one stray keystroke silently destroys a 10-day countdown, or the rule quietly stops meaning anything.

## The rule

**Submitting freezes the working text.** ADR-0012 survives verbatim, because the room never has to infer what an edit is.

- **Drafting** — the room is open; every member co-writes freely.
- **Pending publish** — the document is sealed **read-only for everyone**, including whoever submitted. There is nothing to reset, because nothing can change.
- **Reopening is `pullBack()`** (#590), which already exists and already resets consensus. ADR-0059 established that re-entry after submitting is not a raw write; this makes it the only door.

"Freeze", not "lock" — a lock suggests one member holds it against the others. A freeze binds everyone equally.

## Scope

**Server enforces the freeze.** A pending praxis rejects document updates; read-only is not a client-side courtesy. The composer must not be able to talk a frozen room into accepting text.

**The composer shows it.** CodeMirror goes read-only, and the waiting surface (`frontend/src/pages/editPraxis/waiting/PraxisWaitingSurface.tsx`, ADR-0059) explains *why* the text is sealed and offers the `pullBack` re-entry. This is a copy change as much as a behaviour one — a member who cannot type needs to know it is deliberate, not broken. Copy goes in the i18n catalogs (`frontend/src/locales/en/`), never inline.

**Discard the document on publish.** On publish, flatten to `body_text` and **delete the room's stored document**. `pullBack` re-seeds a fresh one through the same single server-side seed from #1740.

This is a privacy decision, not just a storage one: a CRDT retains **tombstones**, so text a player typed and deleted stays in the document's history. Praxes are permanent; their drafts are not, and text a player removed should not outlive the draft they removed it from.

**Confirm the era rule still reads from config.** The pending window is `era.collab_auto_submit_days` — read `era.*`, never hardcode.

## Out of scope

Presence. Any change to the lazy-on-access timeout itself (ADR-0012's flip-on-read helper stays exactly as it is).

## Verification

Backend, in the existing pytest-against-Postgres suite:

- a pending praxis rejects a document update from every member, submitter included
- `pullBack` reopens the room and clears `has_submitted` per ADR-0012
- publishing deletes the stored document, and `body_text` is intact after it is gone
- re-opening a published praxis seeds a fresh document from `body_text` **once**, with no duplication
- a member who deleted text before publishing leaves no trace of it in anything retained

Frontend: static-render tests that the composer and waiting surface render the read-only state and the re-entry affordance.

pixieofhugs:
## The seam

**The room's ASGI app, plus the ADR-0012 consensus transitions.** The freeze is a
server rule or it is nothing — a client-only read-only editor is a courtesy the
composer can be talked out of. The failing tests went in at that seam first
(`backend/tests/integration/test_praxis_room.py`, which drives real `pycrdt`
clients at the mounted app), and four of the five were red on the real defect
before any of this was written.

## What now resets `has_submitted`

The question the issue asked me to close, answered plainly.

**Before this PR:** media add/remove (`routers/praxes.py`, two call sites), kick,
leave, and `pullBack`. #1743 removed the *text* trigger and left nothing in its
place, so a co-author could rewrite a pending collab's body and the 10-day
countdown would run on regardless.

**After:** the same set, minus the hole — because the body cannot change while a
praxis is pending at all. `pullBack` on a pending collab is now ADR-0012's hard
reset (`on_member_edit`), and it is the only door.

**That is a behaviour change to #590 and it is deliberate.** The partial
pull-back — the caller's submission clears, everyone else's stands, the praxis
stays `pending` — only ever worked because typing was a *second* door: a member
who pulled back could still write, and their first keystroke ran the reset the
partial version skipped. Freeze the document and that door is gone, so a partial
pull-back hands a member a write-up they still cannot write in. Worse, it never
had anything at all to offer the **holdout**: they never submitted, so they had
nothing of their own to take back and got a 422 — while being the one member the
countdown actually threatens. `on_member_unsubmit` is deleted, the uncoded-error
ratchet drops by one, and `test_unsubmit_pending_clears_only_caller` is rewritten
as `test_unsubmit_pending_reopens_the_whole_group` with the holdout doing the
reopening.

## How the freeze is enforced

`YRoom.on_message` — the library's own "skip this message" hook, so nothing is
reimplemented. On a frozen praxis, `SYNC_STEP2` and `SYNC_UPDATE` are dropped
before the room applies them. Two things still pass, and both must:

- `SYNC_STEP1`, a client asking what the server holds. A sealed write-up still
  has to be readable, and refusing this leaves the composer on an empty document.
- Awareness, which carries no document state and which `y-websocket` uses as its
  keepalive — dropping it takes the socket down rather than the edit.

The predicate is `status != in_progress`, stated as "not drafting" so a status
added later arrives frozen. It is pushed onto live rooms by the consensus
transitions and re-read from the database whenever a room opens, so a push that
never happened costs one reconnect rather than a permanently wrong room.

## How I verified the tombstone deletion

Not by reading the code — by asserting the premise first, because the test is
worthless if the secret was never stored.
`test_text_a_member_deleted_does_not_outlive_the_draft` types a phrase, deletes
it, and **asserts the retained rows still contain those bytes** before publishing.
Then it publishes and asserts (a) no rows survive for the praxis, (b) no row
anywhere in `praxis_room_update` contains the bytes, and (c) `body_text` is the
right text without them.

**One finding worth your attention: squashing looked like it already solved this,
and half does.** I probed it directly rather than assuming — folding the history
re-encodes it and Yjs garbage-collects deleted content on the way out, so a
squashed row is clean. But that only fires above `_SQUASH_UPDATES_ABOVE` (200), so
the raw tail beneath the threshold holds every recent retraction verbatim, and a
player who cuts a sentence and then submits is exactly the case that never reaches
a squash. Deleting the rows is the only complete answer. This is written into
ADR-0073 so the next person does not re-derive it.

The **client** keeps a copy too. `y-indexeddb` holds the same document, tombstones
included, so `discardRoomStore` deletes it when a praxis publishes. That is also a
correctness fix, not only a privacy one: the server destroys its document and
`pullBack` seeds a **new** one, and the two were built independently — merged,
the praxis holds its body twice. That is ADR-0073's duplication footgun arriving
from the client side, and it is invisible when it happens because both copies say
the same thing.

## The era rule

Unchanged and confirmed. `era.collab_auto_submit_days` has exactly one reader,
`settle_if_window_lapsed`, which takes `era: EraConfig = CURRENT_ERA`; the value
lives in `eras/era_1.py`; no service body imports `CURRENT_ERA`. ADR-0012's
flip-on-read helper is untouched, as the issue required.

## What to eyeball

- **The pullBack semantics change above.** It is the one judgement call in this
  PR that goes beyond the issue's letter, and it is load-bearing. If you want the
  partial pull-back kept, the freeze needs a different shape and I should redo it.
- **The copy.** Five catalog lines, all in `frontend/src/locales/en/forms.json`:
  a new `composer.bodyFrozen` / `bodyFrozenAction` pair, and rewrites of
  `collab.awaitingBody`, `collab.awaitingEditDescription` and
  `collab.holdoutClockCaption` — the last two were left **asserting the opposite
  of the new rule** ("saving any edit cancels the publish window"). "Freeze", never
  "lock", throughout.
- **Where the re-entry sits in the composer.** I put the "this is frozen" line and
  the reopen button in the write-up region, next to the editor that is refusing
  the member, rather than in the footer — the footer's pull-back button belongs to
  whoever has already cast, and the member most likely to be reading this is the
  holdout, who has not. The waiting surface's own re-entry is unchanged; only its
  copy now says *why* the text is sealed.
- **Whether the freeze should also apply to a `submitted` praxis.** It does, under
  the "not drafting" predicate. That closes a second hole: `authorize` only checks
  membership, so before this a member could connect to a published praxis's room
  and rewrite `body_text` with no consensus reset at all.

## Outstanding

**Visual QA and live convergence are not done.** I have no browser and no dev
stack here. Everything below was run locally and passes, but none of it watches
two real tabs converge, and the harness cannot see the thing the freeze actually
reconfigures:

- Backend `pytest --cov=. --cov-fail-under=80` against a dedicated `wz1745_test`:
  **1253 passed, 93.73%**.
- Frontend `lint`, `typecheck`, `typecheck:design-sync`, `test` (**4429 passed**),
  `build`, `budget` (within budget) — the full list read off
  `.github/workflows/test.yml`.
- `regen_api_client.py` leaves `openapi.json` and `schema.d.ts` unchanged; no
  route signature or schema moved.

**The known gap:** `renderToStaticMarkup` runs no effects and CodeMirror builds
its own DOM, so `EditorView.editable` — the compartment the freeze reconfigures —
is not observable in the frontend harness. The new static tests assert the
observable proxies (the frozen line, the re-entry button, the toolbar's absence);
the rule itself is proved server-side. Someone should still open two tabs on a
pending collab and confirm the editor visibly refuses.

Closes #1745


pixieofhugs:

**Held, not merged — CI is green and the review found one defect plus one decision that is the owner's.**

## Blocking: `awaitingEditConfirm` still describes the retired semantics

Three strings in `frontend/src/locales/en/forms.json` were updated for the freeze (`awaitingBody`, `awaitingEditDescription`, `holdoutClockCaption`). A fourth in the same block was not, and it is the **confirm dialog** — the last thing a player reads before acting:

> `awaitingEditConfirm`: "Take your part back out to edit it? **Only your own submission is re-opened** — but the moment you save a change, the publish countdown is cancelled and everyone else's is cleared too."

Both halves are now false. Reopening clears every member's submission **immediately**, not "the moment you save a change", and it is not "only your own". The dialog now understates the consequence of the action it is confirming.

`awaitingEditTitle` ("Take your part back out?") reads per-member too, and `awaitingEditAction` ("Edit my write-up") is at least ambiguous. Worth a pass over the whole `awaiting*` group rather than the one string.

## The owner's call: this changes #590's shipped behaviour

`unsubmit_praxis` on a `pending` collab was a **partial** pull-back — the caller's own submission only, others' standing, and a 422 for a caller who had not submitted. This PR makes it ADR-0012's hard reset for any member, and deletes `collab_consensus.on_member_unsubmit` outright.

The reasoning in the PR is sound: the partial version only worked because typing was a *second* door that ran the reset, and the freeze removes it — leaving the holdout, the one member the countdown actually threatens, with a 422 and no way to stop it. #1745's own text also asserts pullBack "already resets consensus" and is "the only door", which was true of `submitted` but not of `pending`; this makes the code match the issue's stated intent.

But it is still a user-visible capability being removed from a shipped feature, and that is a product decision rather than a build one. Flagged for @pixieofhugs.

**Nothing else in the PR is in question** — the server-side freeze in `YRoom.on_message` (dropping `SYNC_STEP2`/`SYNC_UPDATE` while letting `SYNC_STEP1` and awareness through, so a sealed write-up stays readable), the document discard on publish, and the tombstone finding all read correctly.

The tombstone work is worth calling out as done properly: the agent probed rather than assumed, found that squashing *does* scrub deleted text via Yjs GC on re-encode but **only above the 200-update threshold**, so the raw tail always holds recent retractions — and a player who cuts a sentence then submits never reaches a squash at all. The privacy test asserts the secret is on disk *before* publishing, so it cannot pass vacuously. Deleting the client's `y-indexeddb` copy is also a correctness fix, not only a privacy one: a surviving local copy merging into a re-seeded room duplicates the body invisibly.

Issue #1667, opened 2026-08-10
101 endpoints, 14 unreachable: delete the dead routes and the four-way task-status duplicate


pixieofhugs:
`backend/openapi.json` declares **101 endpoints**. **14 of them are unreachable from the app**, and four of those are the same operation written four times. This issue deletes them.

## How "unreachable" was measured

Not a path grep — those over-report *and* under-report. The chain that was actually followed:

`openapi.json` route → path literal in `frontend/src/api/*.ts` → wrapper fn → **a reference to that wrapper from outside `frontend/src/api/`**

Three false positives were found and cleared this way, and they are worth knowing before you re-run anything:

1. **Wrappers nobody calls.** `approveTask`/`retireTask` are exported from `frontend/src/api/admin.ts` and do reference their paths — so a path-literal grep marks the routes USED. Nothing imports them.
2. **Callback references, not calls.** Matching `\bfnName\s*\(` misses `createCachedResource(getFactions, SESSION_TTL_MS)` in `frontend/src/hooks/useFactions.ts:29`. That wrongly killed `GET /factions`, which is one of only two cached resources in the app. Match the bare identifier, not a call paren.
3. **One page skips the `api/` layer entirely.** `frontend/src/pages/Contact.tsx:24` calls `apiPost('/contact', …)` directly — the only file in the frontend that does. It made `POST /contact` look dead. It is not.

Four further endpoints are unreachable **by design** and are NOT in scope: `GET /auth/google`, `GET /auth/google/callback` (browser navigations, not fetches), `GET /health` (Render's healthcheck), `POST /admin/cli-token` (the CLI seam).

## The headline duplicate: 4 routes for 1 job

`PUT /admin/tasks/{task_id}/status` already performs every task state transition. Three more routes exist that each just pass a different enum to the same service call:

| Route | Body | Reachable |
|---|---|---|
| `PUT /admin/tasks/{id}/status` | `update_task_status(task_id, TaskStatus(data.status), session)` | **yes** |
| `PUT /admin/tasks/{id}/approve` | `update_task_status(task_id, TaskStatus.active, session)` | no |
| `PUT /admin/tasks/{id}/retire` | `update_task_status(task_id, TaskStatus.retired, session)` | no |
| `POST /admin/tasks/{id}/reactivate` | `reactivate_task` — retired→active only | no |

See `backend/routers/admin.py:537-554` and `:330-337`.

[...]

Issue #1090, opened 2026-07-28
Replace the duel rail with the aside duel card (three readings, none on declined)


pixieofhugs:
Part of #1085. Blocked by the new-format issue.

## What

The duel panel on praxis detail becomes the design's compact aside card. `DuelCrossLink` and the twelve `*DuelRail` files are deleted.

## Why the rail can go

The rail carried the whole duel lifecycle — cast roster, next-step line, stakes tiles, and a branch per `DuelStatus`. #1071 moved the **run-up** onto the composer's waiting surface: `active` (you cast, rival hasn't) is narrated there, with the neutral pull-back and the sealed-rival panel, and the author's own `pending`/`active` side now redirects to the composer rather than rendering here.

So detail narrates outcomes only, and the roster / next-step / stakes tiles go with the rail.

## The card

Label, your row, the rival's row (name, score, cross-link), and **one footer verdict line**. Same card, same height, in every state — only the footer changes, plus the rival's row dimming to an em-dash where they have no score.

**Three readings:**

- **live** (`settled`) — "<name> leads by 2.6 · live". The winner floats with the votes until era close (ADR-0011/0052) — never phrase it as decided.
- **won by default** (forfeited) — rival row dimmed, score `—`, footer "Won by default". Their total is *absent*, not losing: once someone forfeits, the tally stops deciding it. The thrown side is back to `in_progress`, so its link 404s — keep it a plain link, which degrades gracefully.
- **final** (`resolved`) — both scores stand, footer "Final · <name> won — frozen at era close", or the no-contest line when the duel never became votable.

**No card at all on `declined`.** The duel link drops and the praxis is a normal `type=solo` (ADR-0011), so the page shows nothing. Note this is a deliberate change: the shipped rail *does* draw a declined state.

Copy for all of it already exists in `praxis.json` under `duelCrossLink` — `wonByDefault`, `final`, `finalNoContest`, `finalStanding`.

## Delete

`pages/praxisDetail/DuelCrossLink.tsx`, the twelve `pages/praxisDetail/duelRails/*`, the `duelRail` / `mobileDuelRail` manifest lines on six factions, and whatever in `components/duel/shared.tsx` becomes unreferenced. **Check first** — `shared.tsx` also serves the duel seal confirms, which stay.

## Check

A settled duel, a forfeited duel (from both sides), a resolved duel, a no-contest, and a declined challenge — the last rendering no duel card at all. Mobile stacks the card above the proof.

PR #1806, opened 2026-08-15
D3's era reset points at the script that replaced the route

### LINKED ISSUE #957: Duel: render the resolved state (D3) — expose frozen fields + rail resolved branch
Part of #953. **Blocked by #954** (flip its D3 `test.fail()` green).

A `resolved` duel (frozen at era close) renders wrong. Backend + frontend both need a fix.

## Backend
`DuelDetailOut` (`schemas/duel.py:48-56`) omits the frozen-outcome fields that already exist on the model: `winner_character_id`, `challenger_final_points`, `opponent_final_points` (`models/duel.py`). Expose them so the UI can render a correct resolved state.

## Frontend
- `api/duel.ts:7` `DuelStatus` type is missing `'resolved'` (backend serializes it — `schemas/duel.py:52`). Add it.
- `DuelCrossLink.tsx` has no `resolved` branch: a resolved duel falls through to the `settled` block (`:329-369`) and shows a **live** `points_from_votes` tally (`:357-360`) instead of the frozen final points. `NextStepLine` (`components/duel/shared.tsx:351-368`) also has no `resolved` case. Add a `resolved` branch that renders the frozen tally + winner.

## Test note
The e2e coverage for this lives in #954's isolated, era-reset-triggering resolved spec (runs last — global reset is destructive). This fix flips it green.

## Done when
- `DuelDetailOut` serializes the frozen fields; FE type includes `resolved`; the rail renders a resolved duel with frozen points + winner.
- #954's D3 step passes.

pixieofhugs:
`frontend/e2e/duel-zzz-resolved.spec.ts` called `PUT /admin/era/reset` and told the fix author, in HAZARD 2, that they "only [have] to supply admin auth". That premise died in #1667: the route is deleted and pinned deleted in `backend/tests/test_deleted_routes_stay_deleted.py::DELETED_OPERATIONS`, so no auth will ever make that request work. The rollover's only entry point is `backend/scripts/era_reset.py`.

## What changed
- **HAZARD 2 rewritten** — names the script, cites the deleted-routes guard, and says re-adding the route is the one fix that is out of bounds. A dev-only HTTP seam (what #957 originally asked for) is still an option, but it is a *new* route and owes that guard an argument first.
- **The call is now the real seam** — `resetEraViaScript()` shells out to `scripts/era_reset.py pixie --yes` (`@pixie` is `seed.py`'s admin; the script refuses a non-admin because `Era.started_by` records who opened the era).
- **Guarded on the isolated database** — refuses unless `DATABASE_URL` matches `_e2e`, the same rule `scripts/reset_e2e_db.py` enforces on itself. An ad-hoc `npm run e2e` points at the dev DB, and this is the most destructive operation in the system with no undo.
- **`run-e2e.sh` exports `E2E_PYTHON`** so the spec gets the same interpreter (venv-aware) the rest of the pipeline uses. `DATABASE_URL` was already exported, and pydantic-settings lets it win over `backend/.env`, so the script hits the same DB the backend under test does.
- Stale framing fixed: #957 is closed — the `resolved` rendering shipped (#957, #1090) — so the header, the test title, and the fixme reason now say what is actually missing.

## Still `test.fixme`
The shell-out has never been run green, and on the ad-hoc flow it cannot run at all. Finishing it is: run the spec under `bash frontend/e2e/run-e2e.sh`, confirm the rollover drives the duel to `resolved`, delete the `test.fixme()`.

## Verification
`tsc --noEmit` and `eslint` clean on the spec; `playwright test --list` collects it (so the ESM `import.meta.url` path resolution transpiles). The suite itself was not run — it needs Postgres and a seeded e2e DB. E2E is nightly and not PR-blocking either way.


pixieofhugs:

**Review pass — no blocking findings.** All three required checks green (`test`, `api-schema`, `frontend`).

Note for the record: `gh pr checks` briefly reported `test` as failing. It was **cancelled**, not failed — a duplicate run superseded by concurrency — and the sibling run had already passed on the same commit. Re-run rather than merged around.

Verified rather than assumed:

- **`scripts/era_reset.py pixie --yes` matches the script's own documented usage** — its module docstring line 4 is literally `python scripts/era_reset.py pixie --yes`, and `argparse` takes `username` positionally with `--yes` as the confirmation gate. `seed.py::bootstrap_admin` does create the Pixie admin account + admin role, and `reset_era()` exits non-zero on a non-admin username because `Era.started_by` records the operator.
- **`BACKEND_DIR` resolves correctly.** `new URL('../../backend', import.meta.url)` from `frontend/e2e/` lands on `<root>/backend`, which is the cwd the script needs since it imports backend modules.
- **The `_e2e` guard holds both ways.** `/_e2e(\?|$)/` matches a bare `…/worldzero_e2e` and a query-suffixed one, and `run-e2e.sh` exports `DATABASE_URL` in the same shell that invokes Playwright, so the value is inherited. A bare `npx playwright test` leaves it unset and the guard fires with a message naming the right entry point — which is the correct failure, not a silent skip.
- **`execFileSync` with `stdio: 'inherit'` throws on a non-zero exit**, so a failed rollover fails the test rather than passing a no-op.

The spec remains `test.fixme()`, so this ships nothing executable — which is what the issue asked for. It corrects a hazard note that was telling the next author to "supply admin auth" for a route deleted in #1667 and pinned deleted by `test_deleted_routes_stay_deleted.py`.

Issue #1676, opened 2026-08-11
Six functional E2E specs fail: smoke, guest, collaboration and duel


pixieofhugs:
Six of the 32 nightly E2E failures are functional specs rather than the contrast sweep. These are the ones that would actually catch a regression, so they are worth more than their count suggests.

Failing (run 31466972630):

- `smoke.spec.ts:5` — authed root shows the FieldDesk, not the marketing Home
- `guest.spec.ts:11` — the dev-login button logs a bot in
- `collaboration.spec.ts` — collaborator's active-tasks sidebar
- `duel.spec.ts:60` — challenge → accept → both seal → settled rail

## Already ruled out — do not spend time re-checking

**Stale copy and missing headings.** Both asserted strings exist and both render inside genuine `<h1>` elements:

| Assertion | Copy | Element |
|---|---|---|
| `/whose shoes today/i` | `common.json:125` | `FieldDesk.tsx:181` |
| `/who are you becoming/i` | `forms.json:8` | `CreateCharacter.tsx:76` |

So this is not the "shared header lost its `<h1>`" failure mode that nearly shipped on 2026-08-03.

## The strongest lead

`FieldDesk.tsx:184` renders the heading as a **ternary**:

```tsx
{drawerEmpty ? t('fieldDesk.headingFirstLife') : t('fieldDesk.heading')}
```

`fieldDesk.headingFirstLife` is `"Nothing in the drawer yet."`. So `"Whose shoes today?"` appears **only when the account already has lives**. If the seeded fixture's character is not present when `smoke.spec.ts` runs, the page is rendering the other branch and the assertion is simply looking at the wrong one.

Worth checking in this order:

1. `frontend/e2e/auth.setup.ts` posts `/auth/dev-login?name=E2E%20Bot`. The route creates a character only when `resolve_active_character` returns None **and** `name` is passed (`backend/routers/auth.py`). Confirm a character actually exists for that account by the time the spec runs — and that the transaction committed.
2. `guest.spec.ts` deliberately uses a bot with **no** character and expects a redirect to character creation. It gets past `getByRole('button', {name: /sign up here/i})).toHaveCount(0)` and then fails on the heading — so the login appears to happen and the destination is wrong. Establish which page it actually landed on; the run uploads `test-results/**/error-context.md` as an artifact.
[...]

Issue #1560, opened 2026-08-02
The roster advertises a second life before the gate — padlock card, and a footer that says it outright


pixieofhugs:
Reported 2026-08-02 from a fresh account: the character roster shows a padlocked **"A second self awaits — Reach Lvl 4 on a life you already carry. 4 levels to go."** card.

**Owner ruling: multiple characters should not be advertised before the gate opens.** A player discovers a second life at level 4; until then they should have no idea it exists.

## Hiding the card is not enough — the page announces it three more times

`pages/FieldDesk.tsx:157` renders the footer **unconditionally**:

> *"A life is a pair of shoes. Lace up an old one, **or cut a new pair from whole cloth**."*

That sentence advertises exactly the thing being concealed. So does the heading — *"Whose shoes today?"* presumes a choice — and the header chip, *"N life in play"*, implies a count that could be higher.

Remove only the padlock and a pre-gate player still lands on a page that asks which of their lives they want, offers to cut them a new one, and shows exactly one card. That reads as broken, which is a louder hint than the padlock was.

## Owner ruling on the fix: don't route there before the gate

The roster should not be reachable while the player has one life and no way to make another. A chooser with one option and no way to add one is a dead end, and it is the only fix that leaks nothing — there is no page to read hints off.

This also follows the repo's own rule: *hide unusable controls, don't show them disabled* (CLAUDE.md). The whole route is currently the disabled version of a control.

## What makes this bigger than it looks — the roster IS the desktop home

`pages/FieldDesk.tsx`'s docstring: *"FieldDesk roster — the authenticated account home (#274)… Rendered at `/` when authed."*

- **Desktop `/`** → this roster.
- **Mobile `/`** → a carried life gets the per-faction mobile home instead (#500), dispatched through `mobileFieldDesk`. So this bug is mostly a desktop condition.

So "don't route there" means **`/` needs to be something else for a pre-gate desktop player** — which is already being built.

### Interlock with #1552 / #1557

**#1557** replaces the desktop `/` with the new home: character card in the sidebar, "Continue where you left off", and a tabbed browse. The design also puts a **`Characters`** pill on the character card, and `components/CharacterSwitcherSheet.tsx` already exists — so the switcher becomes a **sheet opened from the card**, not a page at `/`.

**Land #1557 first.** Once `/` is the new home, this issue is the smaller remainder: make the roster reachable only when it has a choice to offer, and delete the locked dossier and its gate copy.

[...]

PR #1815, opened 2026-08-15
The FieldDesk's <h1> is the page's, not the roster's (#1794)

### LINKED ISSUE #1794: The FieldDesk renders no h1 when the roster gate is shut
The FieldDesk is the authenticated root (`/`). When the roster gate is shut it renders **no `<h1>` at all** — the page's only headings are `<h2>`.

## Evidence

Not inferred from the source — read off the accessibility snapshot Playwright captured for the seeded bot on run 31779247838 (`test-results/smoke-.../error-context.md`, uploaded as the `playwright-report` artifact):

```yaml
- main:
  - heading "Continue where you left off" [level=2]
  - link "Take a Picture of \"Yourself\" Draft - +10 pts Solo"
  - heading "Tasks you can sign up for" [level=2]
  - button "Tasks" [pressed]
  ...
```

No `level=1` node anywhere in `main`, `navigation` or the sidebar.

## Why

`FieldDesk.tsx` puts the `<h1>` **inside** the roster block:

```tsx
<h1 style={headingStyle}>
  {drawerEmpty ? t('fieldDesk.headingFirstLife') : t('fieldDesk.heading')}
</h1>
```

#1560 gated that whole block behind `rosterOffersAChoice(lives, gateClosed, canCreate)`. For an account carrying one life below the second-character level, the gate is shut, the block does not render, and the heading goes with it. `frontend/src/pages/fieldDesk/__tests__/fieldDeskRosterGate.test.tsx:161` pins that state:

```ts
expect(rosterOffersAChoice([life()], true, false)).toBe(false)
```

So this is not a regression in the gate — the gate is behaving as designed. The `<h1>` was simply living inside something conditional.

## Why it matters

That is the state most players are in: one character, below `second_character_level_required`. So the app's signed-in home page has no top-level heading for the common case. A screen-reader user landing on `/` gets a document whose outline starts at level 2, and heading-navigation has nothing to jump to.

This is the same shape as the defect caught on 2026-08-03, where a profile nearly shipped with no `<h1>`.

## What to decide

The page needs a level-1 heading that does not depend on the gate. What it should *say* is the open question, and it is a copy question as much as a markup one:

- `fieldDesk.heading` ("Whose shoes today?") presumes a choice of lives, which is exactly what the gate is withholding — so it cannot simply be hoisted out.
- `fieldDesk.headingFirstLife` ("Nothing in the drawer yet.") is wrong too; the account has a life.
- The gated page's actual subject is the current life and what to do with it. `Continue where you left off` / `Tasks you can sign up for` are its real sections and are already `<h2>`, which is the correct level **if** something sits above them.

Likely a new string. Route through `frontend/src/locales/en/` per ADR-0031 conventions rather than hardcoding.

## Done when

- `/` renders exactly one `<h1>` in every roster-gate state (gate open, gate shut, no lives at all).
- The existing `fieldDeskRosterGate` / `fieldDeskDispatch` suites still pass, and one of them asserts the `<h1>` exists in the **gated** state so this cannot silently regress again.
- Mobile and desktop skins both carry it (`useFormFactor` dispatch, #494) — the mobile FieldDesk is a separate file and was not checked here.

## Not this issue

- Changing the #1560 gate itself. The gate is correct; only the heading's placement inside it is wrong.
- The `smoke.spec.ts` assertion that surfaced this. That is handled in #1676, which re-anchors the spec to `Tasks you can sign up for` — deliberately **not** to an `<h1>`, so the two changes do not collide.

pixieofhugs:
`/` is the authenticated home, and for the commonest account there is — one life, below the era's `second_character_level_required` — it rendered **no `<h1>` at all**. The heading was drawn inside the block #1560 gates, so hiding the roster took the page's only level-1 heading with it.

## The change

`frontend/src/pages/FieldDesk.tsx` only. The `<h1>` and its rainbow rule move **out** of `{showRoster && …}`; the top-bar pill above it and the roster row below it stay **in**, so the gate-open drawing is the same nodes in the same order it ships today: pill, title, rule, roster row. The gate itself (`rosterOffersAChoice`) is untouched — #1560 is correct, only the heading's placement inside it was wrong.

What the heading says:

| state | heading |
|---|---|
| gate open, lives in the drawer | `fieldDesk.heading` — "Whose shoes today?" (unchanged) |
| gate open, drawer empty | `fieldDesk.headingFirstLife` (unchanged) |
| gate shut | **the carried life's display name** |

No new string. Per the owner ruling of 2026-08-15, the gated page's subject is the current life, so the heading names it — data already on the page rather than a new catalogue entry, so ADR-0031 has nothing to route.

The "no character resolvable" row is not a guess: **a shut gate implies a carried life.** `rosterOffersAChoice` returns `true` whenever `!carriesALife`, and `carriesALife` and `active` are the same `user?.character`. So there is always a life to name; the `&& active` in the expression is what lets the type-checker see that, and a blank `display_name` falls to `@username` rather than to an empty heading — an empty `<h1>` would pass a naive "is there an h1" check and fail the actual requirement.

The rule (`rainbowUnderline`) travels with the title because it is the *title's* rule, not the roster's — that keeps the gate-open render identical and stops the gated heading floating ruleless.

## Seam tested at

The **desktop `FieldDesk` render** under `renderToStaticMarkup` — the same seam `fieldDeskRosterGate.test.tsx` already drives, with state pushed through the mocked `useAuth` rather than through an effect (the harness is `node`, no DOM, effects never run, so `lives` stays `null` in every rendered case).

Four new cases in `fieldDeskRosterGate.test.tsx` assert the **text of every `<h1>` in document order**, so "exactly one" is asserted, not just "at least one": gated → `['Mollusk']`, gated with a blank display name → `['@molly']`, gate open → `['Whose shoes today?']`, no life at all → `['Nothing in the drawer yet.']`. The first two go red on `main` (`expected [] to deeply equal [ 'Mollusk' ]`) — verified by stashing the component and re-running.

## Verification

`npm run lint`, `npm run typecheck`, `npm run typecheck:design-sync`, `npm test` (248 files / 4,436 passed), `npm run build`, `npm run budget` (JS 130.0 KB, CSS 21.3 KB — unchanged, within budget) all pass locally. No API contract touched, so `api-schema` has nothing to regenerate.

**Visual QA is outstanding** — no browser or dev stack here.

## Mobile: already compliant, with one exception that is NOT this issue

The issue flagged the mobile skins as unchecked. Seven of the eight archetypes under `frontend/src/pages/fieldDesk/mobileArchetypes/` render their own `<h1>` unconditionally and never mention the roster gate: `Default`, `Coven`, `Ephemerists`, `Everymen`, `Singularity`, `Snide`, `Ua`. **They need no change** — the gate is a desktop-only concern and the mobile branch returns before it.

`WowFieldDesk.tsx` is the exception: it has **no `<h1>`**. Its identity block is `WowPavilionHeader` (`frontend/src/components/factionMarks/wowMobile.tsx`), which draws the name in a plain `<div>`. That is a separate defect from this one — it has nothing to do with the roster gate, it is in a shared component whose second consumer (`frontend/src/pages/factionDetail/mobileArchetypes/WowFactionPage.tsx`) also has no `<h1>`, and promoting that `div` would change the outline of both pages at once. Left alone here; worth its own issue.

`frontend/e2e/smoke.spec.ts` is untouched — its `/whose shoes today/i` assertion belongs to #1676, which is deliberately re-anchoring to `Tasks you can sign up for`. This change neither fixes nor worsens it.

## What to eyeball

- `/` on **desktop**, one life, below `second_character_level_required` (gate shut): the character's display name as the page title, with the rainbow rule under it, then "Continue where you left off" / the browse below. Check the space between the rule and the sections below reads as deliberate — the gated page is now title-then-content with no roster in between, which is a layout `frontend-style` may want an opinion on.
- `/` on **desktop** with the gate open (or two lives): should be pixel-identical to today — pill, "Whose shoes today?", rule, credential cards.
- `/` on **desktop** for a brand-new account with no character: "Nothing in the drawer yet." plus the "Begin a new self" dossier, unchanged.
- A **phone** at `/`: the mobile skins are unchanged; the WOW one still has no `<h1>` (see above).

Closes #1794

### DIFF
--- a/frontend/src/pages/FieldDesk.tsx
+++ b/frontend/src/pages/FieldDesk.tsx
@@ -60,10 +60,15 @@ import DefaultFieldDesk from './fieldDesk/mobileArchetypes/DefaultFieldDesk'
  * #1560 was written when `/` was ONLY the roster, so it asked for the route to
  * be gated. Since #1557 landed, `/` also carries "continue where you left off"
  * and the browse — a pre-gate player still needs those. So the seam is the
- * ROSTER SECTION (chip, heading, rule, cards, dossier, closing line), which is
- * hidden as one piece unless it has a choice to offer. `rosterOffersAChoice`
- * below is that rule, and the create-your-first-life path is inside it: an
- * account with zero lives ALWAYS sees the roster, or signup dead-ends.
+ * ROSTER SECTION (chip, cards, dossier, closing line), which is hidden as one
+ * piece unless it has a choice to offer. `rosterOffersAChoice` below is that
+ * rule, and the create-your-first-life path is inside it: an account with zero
+ * lives ALWAYS sees the roster, or signup dead-ends.
+ *
+ * The page TITLE is deliberately not in that piece (#1794): `/` is the
+ * authenticated home and must have a level-1 heading in every gate state, so
+ * the `<h1>` and its rule sit outside and the heading changes what it says
+ * instead of disappearing.
  *
  * The Albescent letter (#395) is deliberately OUTSIDE the gate. It converts an
  * existing life rather than adding one, so it is a real control for a
@@ -145,6 +150,19 @@ export default function FieldDesk() {
   // does, the carried life is the best proof there is: an account playing
   // someone has at least one. Keeps the heading from swapping under the viewer.
   const drawerEmpty = lives === null ? !carriesALife : lives.length === 0
+  // The page's <h1>, which is the PAGE's and not the roster's (#1794): it used
+  // to be drawn inside the gated block, so the commonest state of all — one
+  // life, gate shut — served `/` with an outline starting at level 2.
+  //
+  // With the roster hidden the page is no longer asking whose shoes; its subject
+  // is the life being carried, so the heading names it (owner ruling
+  // 2026-08-15). No new string — the name is data already on the page. A shut
+  // gate implies `carriesALife`, so `active` is always there to name; the `&&`
+  // is what the type-checker needs to see it, and a blank display name falls to
+  // the handle rather than to an empty heading.
+  const heading = !showRoster && active
+    ? active.display_name || `@${active.username}`
+    : drawerEmpty ? t('fieldDesk.headingFirstLife') : t('fieldDesk.heading')
 
   // Phone + a carried life → the mobile-native home skin for that faction.
   if (formFactor === 'mobile' && homeState) {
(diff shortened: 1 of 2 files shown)


pixieofhugs:

**Review pass — no blocking findings.** CI green on all three required checks (`test`, `api-schema`, `frontend`) across both runs.

Checked:

- **Gate-open drawing is unchanged.** Node order is still pill → title → rule → roster; the `<>` fragment collapsed into two sibling `showRoster &&` blocks with the title between them. The rule travels with the title, which is correct — it is the title's rule, not the roster's.
- **The heading's fallback chain is sound.** A shut gate implies `carriesALife` (`rosterOffersAChoice` returns `true` whenever `!carriesALife`), and `carriesALife` and `active` are the same `user?.character`, so the gated branch always has a name. The `&& active` is type-checker-only, as the comment says. `|| @username` covers a blank `display_name` — worth keeping, since an empty `<h1>` passes a naive existence check while failing the requirement.
- **The unreachable fourth combination is genuinely unreachable**, not merely unhandled: gate shut with no active character would require `carriesALife` false, which opens the gate.
- **The tests pin the right thing.** `h1s()` returns every `<h1>` in document order and all four cases assert with `toEqual([...])`, so "exactly one" is asserted rather than "at least one". Red-first was demonstrated on the two gated cases.

**Nit, not blocking:** `heading` is computed above the `formFactor === 'mobile'` early return, so it is evaluated and discarded on the mobile path. No behaviour change and moving it would split the constant from the comment explaining it.

**Owed to a human:** the gated page is now title → rule → `--space-4xl` gap → "Continue where you left off", with no roster between them. That gap was tuned for a page that had a roster under the rule. Flagged in the PR's own "What to eyeball" and left to `frontend-style`.

Issue #1754, opened 2026-08-15
The Ephemerists plate needs its own --label-ink; the global lift made it worse


pixieofhugs:
> *This was generated by AI during /builder-bot.*

Surfaced by **#1715** (PR #1753), which lifted the neutral muted tokens to AAA. That issue said outright: *"Do not introduce per-faction overrides to compensate. If one kit genuinely breaks, say so in the PR rather than patching around the global."* This is that report.

## The one real casualty

The lift raises contrast almost everywhere — light paper gains ~2 points, dark gains everywhere. But the **three always-dark sheets lose ~1 point**, because both muted inks moved *lighter* in the light cascade and those sheets draw light-cascade tokens on a dark ground:

| sheet | before | after |
|---|---|---|
| S.N.I.D.E. | 3.18 | 2.19 |
| Singularity | 3.29 | 2.27 |
| **Ephemerists plate** | **2.92** | **2.01** |

**S.N.I.D.E. and Singularity cover themselves** — a flipping slip stock and chassis-level repoints respectively, so their labels do not read the global.

**The Ephemerists plate does not.** Verified on `origin/main`: `.eph-plate-sheet` (`index.css:3878`) repoints

```css
--link-ink:       var(--faction-ephemerists-plate-nile);
--link-ink-hover: var(--faction-ephemerists-plate-ink);
```

and **no `--label-ink`**. So its label tier inherits the global tertiary and follows it wherever it goes.

## This is not a new failure — it is an old one getting worse

2.92:1 was already far under the 4.5:1 AA floor. The lift did not break a passing surface; it deepened a failing one. That matters for prioritisation: this is not a #1715 regression to revert, it is a pre-existing gap #1715 made visible and measurable.

Note also that the `--label-ink` block's own recorded figures had **already expired unnoticed** — they read "vellum 4.36 / plate 4.35", which predates #1627 taking that plate to a night ground, where it actually measured 2.92. PR #1753 corrects those numbers. A hand-kept figure in a comment rotted for the whole interval.

## What this is

The per-faction label ink that **#1307 measured and deferred**, and that **#1608** cites as a dissolved precondition — `--label-ink` currently has exactly one declaration (`index.css:89`) and no faction repoints it, which is why nothing has fired before now.

[...]

Issue #1627, opened 2026-08-03
Ephemerists' card tokens move from vellum to the blue-and-bronze plate


pixieofhugs:
Ephemerists' card surface is aged vellum. It should be the blue-and-bronze plate.

Design: https://claude.ai/design/p/9305f548-25ce-4dac-9d35-10867994831f — *Ephemerists Kit*, which mounts every Ephemerists-skinned surface from the real components. Its own prose: *"Every Ephemerists-skinned surface in the kit, mounted from the real components."*

## The change

Four lines. The kit design states them verbatim at `:179-182`, and the *Faction Profile Cards* design repeats the identical four — which is how this was found:

```css
--faction-ephemerists-card-bg:     var(--faction-ephemerists-plate-band);
--faction-ephemerists-card-text:   var(--faction-ephemerists-plate-band-ink);
--faction-ephemerists-card-accent: var(--faction-ephemerists-plate-brass-light);
--faction-ephemerists-card-muted:  var(--faction-ephemerists-plate-band-quiet);
```

Today those resolve to the vellum family — `index.css:241` is `var(--eph-vellum)`, commented *"Ephemerists: aged manuscript stock"*, plus `:249`, `:270`, `:284`. All four `plate-*` tokens already exist.

## Blast radius — this is the point, not a side effect

`--faction-ephemerists-card-*` is read **generically**, through `factionCssVar(slug, 'card-*')`. Repointing it re-skins Ephemerists **everywhere at once**:

`components/collab/CollabRoster.tsx` · `collab/CollabSuccess.tsx` · `comments/CommentThread.tsx` · `factionCard/FactionCard.tsx` · `feed/FactionFeedFrame.tsx` · `praxisCard/` · plus `components/vote/VoteShell.tsx:106`, which names `--faction-ephemerists-card-accent` **directly** rather than through the helper — check that one by eye, it will not show up in a `factionCssVar` grep.

That reach is intended: the whole faction is being re-skinned. But it means this is **not** a card change, and it must not be bundled into one.

## Measure before declaring it done

The plate palette is a dark band with brass; the vellum is light stock with rubric ink. Every pairing measured against `card-bg` is now measured against a different ground.

- Re-run the faction contrast suite — `frontend/src/__tests__/factionContrast.test.ts` is the guard, and any Ephemerists pair asserted there needs its expectation re-derived, not merely re-run.
[...]

Issue #1413, opened 2026-07-31
Translucent --color-bg-surface makes .sidebar-card fail AA over Singularity's dark praxis-detail page


pixieofhugs:
Found while building #1302 (PR #1411), out of its scope and a **different defect class** — this is not a danger/warning token problem, it is a translucency problem.

## The mechanism

`.sidebar-card` grounds on `--color-bg-surface`, which is **translucent**: `rgba(255,255,255,0.72)` in light. Over a light page that resolves to near-white and everything reads fine. Over **Singularity's praxis-detail page, which is always dark**, the same 72% white over a dark ground resolves to a **mid-grey** — and every ink measured against the assumed near-white surface loses most of its contrast.

Measured on that resolved grey:

| ink | ratio |
|---|---|
| `--color-danger` | **2.54:1** |
| `--color-warning` | **2.65:1** |
| `--color-text-tertiary` | **3.11:1** |

All three fail AA (4.5:1 for body text, 3:1 for large/non-text).

## Why it is worth its own issue

**It affects every neutral-chrome card mounted on that page, not just the ones carrying danger or warning.** The tokens are correct against the surface they were measured on; the surface is the thing that changes underfoot. Fixing the inks would be treating the symptom, and would break them everywhere they are currently right — the same lever #1169 already rejected.

It is also invisible to the existing guard: `factionContrast.test.ts` asserts a foreground against a **named token**, and no assertion can name "72% white composited over whatever page this happens to be mounted on". A translucent surface has no single value to measure.

## Worth deciding, not guessing

Two shapes, and they are not equivalent:

- **Make `.sidebar-card` opaque on dark grounds** — give it a solid surface token so its inks have a fixed ground again. Loses the frosted effect wherever that is deliberate.
- **Give Singularity's detail page its own card surface** — the per-sheet mechanism #694 / #1168 / #1411 already established. Consistent with how this codebase has solved ink-on-wrong-ground three times, but this is a *ground* problem rather than an ink one, so the analogy is not exact.

Whichever lands, the guard question needs answering too: **a translucent surface token is currently unassertable**, and that is why this survived. Consider whether `contrast.ts` should refuse to measure against a token with an alpha channel, so the gap is loud instead of silent.

## Related

#1302 / PR #1411 (per-sheet inks for the praxis card's admin marks — deliberately did NOT touch this), #1169 (why deepening a global ink is the wrong lever), ADR-0061.

pixieofhugs:
[...]

PR #1818, opened 2026-08-15
A night-plate ink is not for the app's page, and the second Ephemerists root owed the label seam

### LINKED ISSUE #1793: faction-ephemerists-plate-quiet is documented as AA on all three grounds and measures 2.64:1
`--faction-ephemerists-plate-quiet` carries a comment asserting it clears AA on all three of its grounds. It measures **2.64:1** on one of them.

## The measurement

Nightly `workflow_dispatch` run [31867605613](https://github.com/pixieofhugs/WorldZeroPlayground/actions/runs/31867605613) (2026-08-15), `ephemerists / light / desktop`:

```
2.64:1 (needs 4.5:1) - rgb(162, 151, 122) on rgb(247, 244, 238)
  div.wz-faction-grid > div > div > p.content-text (18px/400)
  "No surveys open."
  "Nothing sealed to the codex yet."
```

`#a2977a` on `#f7f4ee`. Both empty-state strings on the Ephemerists faction page.

## The false comment

`frontend/src/index.css:1601`:

```css
--faction-ephemerists-plate-quiet: #a2977a; /* quiet ink - AA on all 3 grounds */
```

The same hex is also `--faction-ephemerists-plate-muted` at `index.css:1562` (`/* brief + in-progress */`). Check whether that one is used as text anywhere before assuming only the `-quiet` declaration needs to move — a hand-written contrast claim that was wrong once is worth re-deriving for its twin rather than trusting.

**Fix the comment as part of the fix.** A stale contrast assertion in `index.css` is how the next person concludes this token is safe without measuring. This is the same failure mode as `contrastBaseline.ts`'s header claiming 269 entries when the file held 241 (noted in #1780).

## Not a type-scale violation

Unlike #1766's items 3 and 4, this text is at 18px/400 — on the content floor, not under it (`WORLD_ZERO_STYLE` §4). The size is correct; only the ink is wrong. Do not "fix" it by changing the size.

## Watch for

`--color-bg-surface-alt` and friends are translucent and composite over the page — measuring `#a2977a` against the wrong ground is exactly the mistake #1727 documents twice (#1715, #1579). `#f7f4ee` above is what the scanner resolved after compositing; use `frontend/src/utils/contrast.ts`'s `compositeOver` if you re-derive it by hand, and confirm against the Ephemerists page's other two grounds before picking a replacement.

## Done when

Both empty-state strings clear 4.5:1, the `index.css` comment states what is actually true for every ground the token is used on, and `contrast.spec.ts` reports no measured failure at `div.wz-faction-grid ... p.content-text`.

## Related

- **#1766**, and the sibling sky-numeral issue — the other measured failures from the same run.
- **#1727** — composite correctly when re-deriving.

### LINKED ISSUE #1800: The Ephemerists composer root has the same --label-ink gap #1754 just fixed on the plate sheet

**#1754 (PR #1799) fixed one of two Ephemerists roots.** This is the other one, and it is one line.

`.eph-plate-sheet` in `frontend/src/index.css` now sets `--label-ink: var(--faction-ephemerists-plate-quiet)`, taking the label tier from 2.01:1 to 5.98:1 on the plate ground.

`frontend/src/pages/editPraxis/archetypes/EphemeristsEditPraxis.tsx`'s `dress.pageStyle` — mounted by `ComposerPage` and `PraxisWaitingSurface` — is the **second** Ephemerists root and has the identical gap. Unset, its label tier inherits the global tertiary onto the same night ground and reads the same failing number.

**#1636 already had to set the *link* seam on both roots for exactly this reason.** Two roots, one column, every seam owed to both.

## Why it was not done in #1754

The #1754 agent found it and was explicitly scoped out of `editPraxis/**` — #1742 and #1743 were live in that directory at the time. It is not an oversight; it is a deferred one-liner.

## The change

Add the same declaration `.eph-plate-sheet` now carries to `EphemeristsEditPraxis.tsx`'s `dress.pageStyle`. **Same value.** Re-measure with `frontend/src/utils/contrast.ts` (`parseColor`, `compositeOver`, `contrastRatio`) against the composer's actual composited ground rather than assuming it matches the plate's — the numbers in this issue are the *plate's*, not the composer's, and a ratio in prose expires when either token moves.

Confirm both roots agree afterwards, so the next seam does not have to rediscover that there are two.

## Also spotted, pre-existing, NOT part of this issue

`--faction-ephemerists-plate-cta-bg` (brass) fails every candidate label ink, including today's neutral: **3.58:1 light, 1.15:1 dark**. Nothing is broken today because every CTA site sets `CTA_INK` explicitly — but it is a ground no label tier can ever sit on, and nothing records that. Worth a comment at the token so the next person does not try. File separately if it deserves its own ticket.

## Related

- #1754 / PR #1799 — the plate sheet fix and its measurements
- #1636 — the precedent: the link seam needed setting on both roots
- #1715 — the AAA lift that made this measurable
- #1307 — where the per-faction `--label-ink` mint was measured and deferred

---

**Agent:** `frontend-style`

pixieofhugs:
Closes #1793
Closes #1800

One token, `--faction-ephemerists-plate-quiet`, so one PR.

## The seam I tested at

`frontend/src/utils/__tests__/factionContrast.test.ts` — the token-value manifest and its
source-assertion blocks. Both defects are the shape that file already names as the one a
ratio row cannot see: *an ink measured against a ground it is not actually mounted on*
(#1413/#694). So the guards assert on the source and on the luminance bounds, not on a
pairing the manifest could have listed. The loop went red on three tests before any fix:
the two faction-page empty states and the composer's missing `--label-ink`.

## Re-derived measurements

Every number below was measured in this branch with `frontend/src/utils/contrast.ts`
(`parseColor` → `compositeOver` → `contrastRatio`) against tokens resolved out of
`index.css` through the test's own `cssVars` resolver — not copied from either issue.
No ground here turned out to be translucent, so no compositing stack was needed; the
resolver was still run over each one rather than assumed.

### #1793 — `-plate-quiet` (`#a2977a`), on every ground it is painted on

| colour | ground | ratio | where I measured it |
|---|---|---|---|
| `#a2977a` | `#f7f4ee` app page, light | **2.64:1 FAIL** | `EphemeristsFactionBody` tasks + praxis empty states — outside any `CARD`, root is a bare `.wz-faction-grid` |
| `#a2977a` | `#13121a` app page, dark | 6.42:1 | same two nodes, other cascade — which is why this only ever reported in light |
| `#a2977a` | `#101320` `-plate-page` | 6.38:1 | task/praxis detail column ground |
| `#a2977a` | `#171a26` `-plate-bg` | 5.98:1 | the card sheet and `.eph-plate-sheet` |
| `#a2977a` | `#1d2130` `-plate-inner` | 5.52:1 | the action panel's cells |
| `#a2977a` | `#12151f` `-plate-disc` | 6.28:1 | the medallion field |
| `#a2977a` | `#0a0c15` `-plate-band` | 6.73:1 | the cornice masthead |
| `#554b3c` | `#f7f4ee` app page, light | **7.78:1** | the ink the two empty states now take |
| `#b9ac97` | `#13121a` app page, dark | **8.33:1** | same, dark |

**The token does not move, and that is the finding.** Clearing 4.5:1 on the register's
loosest ground, `-plate-inner`, puts a *floor* under the ink's relative luminance of
**0.2453**. Clearing 4.5:1 on the light page ground puts a *ceiling* on it of **0.1626**.
A contrast ratio is monotonic in luminance, so the intervals not meeting means the set of
satisfying sRGB values is empty — whatever the hue. Same shape of answer the Ephemerists
register itself reached one axis over (`-plate-bg`'s comment, on two *themes* rather than
two *page grounds*). The two empty states take the app's own second tier instead, which is
what the `Kicker` beside them has read since #1675 and for the same stated reason.

That inequality is asserted, not described — `no value of \`-plate-quiet\` could have
cleared both registers` recomputes both bounds from the live tokens, so if a repaint ever
makes the split unnecessary the test says so rather than fossilising it.

### The `-plate-muted` twin — audited, sound, left alone

Same hex, second name. Painted as text in exactly one file (`EphemeristsTaskCard.tsx`,
three sites) on exactly two grounds: `-plate-bg` **5.98:1** and `-plate-disc` **6.28:1**.
Both are the register's own night stocks, so unlike `-plate-quiet` it is never off-register.
Its comment now records that, and records that the two names are two roles that converged —
so a future repaint of one does not assume the other follows.

### #1800 — the label seam on the second Ephemerists root

| colour | ground | before | after |
|---|---|---|---|
| `--label-ink` unset → `#4d4868` | `#171a26` `-plate-bg`, light | **2.01:1** | **5.98:1** |
| `--label-ink` unset → `#4d4868` | `#1d2130` `-plate-inner`, light | **1.86:1** | **5.52:1** |
| `--label-ink` unset → `#b3b0d6` | `#171a26` `-plate-bg`, dark | 8.33:1 | 5.98:1 |
| `--label-ink` unset → `#b3b0d6` | `#1d2130` `-plate-inner`, dark | 7.69:1 | 5.52:1 |

Dark spends ~2.3 points to hold one ink across a register that declares nothing under
`[data-theme="dark"]`, and stays well over the floor — the same trade #1754 made on the
sheet, deliberately matched so both roots read one quiet.

**The composer's page ground is NOT the plate's, and I checked rather than assumed.**
`ComposerPage` renders a bare `<div style={style}>` with no background, so what is under
the sheet here is the app's `--color-bg-page` — where `-plate-quiet` is the 2.64:1 of
#1793, not the 6.38:1 of `-plate-page` that `.eph-plate-sheet` paints for itself. It is
safe anyway: every `.label-caption` / `.label-heading` this seam cascades to renders inside
`ComposerSheet` (`-plate-bg`) or a `panelStyle` cell (`-plate-inner`), and the one element
on the bare page — the breadcrumb — takes `breadcrumbInk` explicitly. Recorded in the code
so the first label mounted directly on that page is caught. **Both roots now agree**, which
the test asserts as one `it`, reading `.eph-plate-sheet` and `dress.pageStyle` together.

### The brass CTA bar — a comment, not a change (as instructed)

I agree with #1800's reading, and re-derived it:

| colour | ground | ratio |
|---|---|---|
| `#4d4868` app tertiary, light | `#c9a24b` `-plate-cta-bg` | 3.58:1 |
| `#b3b0d6` app tertiary, dark | `#c9a24b` | 1.15:1 |
| `#a2977a` `-plate-quiet` | `#c9a24b` | 1.21:1 |
| `#12151f` `-plate-cta-ink` | `#c9a24b` | **7.59:1** |

No quiet ink can sit on a mid-luminance brass in either cascade — the bar's own dark partner
is the only thing that works, and every CTA site already sets `CTA_INK` explicitly. Left as
a comment at the token recording that, so the next person does not repoint the seam for one bar.

## Also corrected, same disease

`EphemeristsFactionBody`'s file header was quoting **eight** ratios measured on the papyrus
plate; #1627 turned the register night and nothing asked the comment to keep up
(`ink` 11.3 → 13.91, `quiet` 5.6 → 5.98, `caption` 4.8 → 7.22, `nile` 5.0 → 7.00,
`band-ink` 12.4 → 14.00, `gold` 9.4 → 14.00, `band-quiet` 8.6 → 8.97). It also asserted the
page beneath was `.eph-backdrop` — #1675 already found that nothing applies it here, which
is precisely the assumption both of these bugs grew out of.

## Verification

`npm run lint`, `npm run typecheck`, `npm run typecheck:design-sync`, `npm test`
(248 files / 4,439 passing), `npm run build` and `npm run budget` (JS 130.0 KB, CSS 21.3 KB,
both inside warn) all green locally. The built stylesheet was grepped, not just the source —
`--faction-ephemerists-plate-quiet`, `-plate-muted`, `-plate-cta-bg` and the whole
`.eph-plate-sheet` body all resolve in `dist/assets/index-*.css`, so no comment edit ate a
brace. Nothing here touches a route, a `response_model` or a Pydantic schema, so `api-schema`
has nothing to regenerate. No `contrastBaseline.ts` entry existed for this pair (it was a new
nightly finding, not allowlisted), so there is none to delete.

**Visual QA is outstanding** — I have no browser and did not run a dev server.

## What to eyeball

1. **Ephemerists faction page, LIGHT, desktop** — the ④ Tasks and ⑤ Praxis empty states
   ("No surveys open.", "Nothing sealed to the codex yet."). This is the fix: they should
   now read as legible warm-grey prose, matching the italic kicker directly above each of
   them, rather than washing out. Needs a faction with zero open tasks and zero recent praxis.
2. **Ephemerists faction page, DARK, desktop** — same two strings. They were passing before
   at 6.42:1 and are now 8.33:1, so this is a check that the tone shift is acceptable, not
   that it is legible.
3. **Ephemerists faction page, LIGHT, mobile** — the same two strings once the grid stacks;
   the ground does not change but the column width does.
4. **Ephemerists faction page, both themes** — the ③ Road block, the keeper spotlight and
   the roster empty state. These deliberately did NOT change: they are inside a `CARD` on the
   night plate and keep the faction's quiet ink. If any of them shifted tone, the split is wrong.
5. **Ephemerists praxis composer, LIGHT, desktop** — every small label on the sheet: the
   title counter, the section captions, the collab/duel chips. They were near-invisible at
   2.01:1 and should now be the plate's warm quiet.
6. **Ephemerists praxis composer, LIGHT, desktop, in a panel cell** — the labels inside the
   action panels specifically (1.86:1 before), which is the tightest of the pair.
7. **Ephemerists praxis composer, DARK, desktop** — the same labels. This is the direction
   that *loses* ~2.3 ratio points by design; the check is that they still read as a quiet
   tier and not as body copy.
8. **Ephemerists praxis WAITING surface (post-cast), LIGHT and DARK** — same root, different
   stage, and the half that is easy to forget. The nudge receipt and the pull-back description
   under the primary button.
9. **Ephemerists praxis composer + waiting surface, mobile, both themes** — the sheet narrows
   and more labels wrap; nothing should change colour between form factors.
10. **A NON-Ephemerists composer and the praxis-detail steward bar, either theme** — the
    control. `.card-on-page` pins `--label-ink` back to the global neutral, and none of this
    should have leaked off the Ephemerists column.

## Two things I noticed and did not fix

- `frontend/src/pages/editPraxis/archetypes/controls.tsx:1004` and `:1023` hardcode
  `color: "var(--color-text-tertiary)"` on `.label-caption` (the `bodyConnecting` and
  `bodyFrozen` notices). That defeats the seam this PR just set: on the Ephemerists sheet
  they read **2.01:1** in light, and they will do the same on S.N.I.D.E. and Singularity.
  It is the exact pattern `factionContrast.test.ts` warns about at its `.label-caption`
  assertion, one layer out where the CSS guard cannot see it. Cross-faction and not this
  issue — worth its own ticket.
- `--faction-ephemerists-plate-muted` and `-plate-quiet` are the same hex under two names.
  Both are sound and each is measured, so collapsing them is a judgement call about roles
  rather than a fix; I recorded the duplication at both declarations instead of deciding it.

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -1567,7 +1567,19 @@
      gold, which is likewise never painted as text. */
   --faction-ephemerists-plate-bg: #171a26;
   --faction-ephemerists-plate-ink: #f0e6c8;
-  --faction-ephemerists-plate-muted: #a2977a; /* brief + in-progress */
+  /* Brief + in-progress, on the task card. THE SAME HEX AS `-plate-quiet` below,
+     which is a fact worth stating rather than leaving to be rediscovered: #1793
+     re-derived this twin when it caught that one's comment lying, and this one
+     is sound. It is painted as text in exactly one file
+     (`EphemeristsTaskCard.tsx`) on exactly two grounds — the plate 5.98:1 and
+     the medallion disc 6.28:1 — both of which are the register's own night
+     stocks, so it is never off-register the way `-plate-quiet` was.
+
+     Two names for one value is normally what this file warns about; they are
+     kept apart because they are two roles that happen to have converged, and
+     collapsing them would decide for the card what the page decided. If a
+     repaint ever walks one, walk it knowing the other does not follow. */
+  --faction-ephemerists-plate-muted: #a2977a;
   --faction-ephemerists-plate-caption: #c9a24b; /* small-caps labels */
   --faction-ephemerists-plate-brass: #c9a24b; /* rules, borders — never an ink */
   --faction-ephemerists-plate-brass-light: #e6c877;
@@ -1585,7 +1597,17 @@
   --faction-ephemerists-plate-disc: #12151f; /* the medallion's field */
   --faction-ephemerists-plate-ochre: #d9744c; /* the fifth tally stroke */
   /* The CTA bar swaps ends against the light plate it replaced: a BRASS bar
-     carrying dark ink, where the papyrus had a dark bar and gold ink. */
+     carrying dark ink, where the papyrus had a dark bar and gold ink.
+
+     NO LABEL TIER CAN SIT ON THIS BAR (noted while measuring #1800; nothing is
+     broken, this is a record so nobody tries). It is a mid-luminance brass, and
+     the label seam is a QUIET ink by definition — the app's tertiary reads
+     3.58:1 on it in light and 1.15:1 in dark, and `-plate-quiet`, the ink both
+     Ephemerists roots point `--label-ink` at, is 1.21:1. There is no quiet
+     neutral this bar can carry, in either cascade. What works is the dark
+     partner the bar was drawn with: `-plate-cta-ink`, 7.59:1. Every CTA site
+     sets that explicitly today, which is why the seam's value never lands here
+     — keep it that way rather than repointing the seam for one bar. */
   --faction-ephemerists-plate-cta-bg: #c9a24b;
   --faction-ephemerists-plate-cta-ink: #12151f;
   /* The gold wash behind the cartouche and the ratio chip is GONE, not dimmed —
@@ -1606,7 +1628,26 @@
      collapsed onto it rather than onto its twin. */
   --faction-ephemerists-plate-page: #101320; /* the page ground under the plate */
   --faction-ephemerists-plate-inner: #1d2130; /* the action panel's two cells */
-  --faction-ephemerists-plate-quiet: #a2977a; /* quiet ink — AA on all 3 grounds */
+  /* Quiet ink for the register's THREE NIGHT GROUNDS, and only those: the page
+     6.38:1, the plate 5.98, the panel cell 5.52 — one reading each, since the
+     register declares nothing under dark (#1627). It also clears the two grounds
+     the family added later, the medallion disc 6.28 and the cornice band 6.73,
+     though the band has its own second tier in `-plate-band-quiet`.
+
+     WHAT IT IS NOT AA ON (#1793). The APP's page ground: 2.64:1 on #f7f4ee in
+     light. This line used to read "AA on all 3 grounds" full stop, and a bare
+     `.wz-faction-grid` is not one of those three — `EphemeristsFactionBody`'s
+     two empty states were painted in it on the app's own near-white page and the
+     nightly sweep measured them there. They now take `--color-text-secondary`,
+     because NO value of this token could have served both: clearing the panel
+     cell (#1d2130) needs relative luminance >= 0.1937 and clearing the light
+     page needs <= 0.1611, and the intervals do not meet. Same shape of finding
+     as the `-plate-bg` block above, one axis over — there it was two themes,
+     here it is two page grounds.
+
+     The rule for the next reader: this ink belongs to a surface this register
+     paints. Off the register, the ground is the app's and so is the ink. */
+  --faction-ephemerists-plate-quiet: #a2977a;
   --faction-ephemerists-plate-nile: #5fb3a6; /* links, affirmations, the eye */
   --faction-ephemerists-plate-rule: rgba(201, 162, 75, 0.12); /* journal ruling */
   --faction-ephemerists-plate-line: rgba(201, 162, 75, 0.28); /* hairline borders */
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -1640,8 +1640,8 @@
      two empty states were painted in it on the app's own near-white page and the
      nightly sweep measured them there. They now take `--color-text-secondary`,
      because NO value of this token could have served both: clearing the panel
-     cell (#1d2130) needs relative luminance >= 0.1937 and clearing the light
-     page needs <= 0.1611, and the intervals do not meet. Same shape of finding
+     cell (#1d2130) needs relative luminance >= 0.2453 and clearing the light
+     page needs <= 0.1626, and the intervals do not meet. Same shape of finding
      as the `-plate-bg` block above, one axis over — there it was two themes,
      here it is two page grounds.
 
--- a/frontend/src/pages/editPraxis/archetypes/EphemeristsEditPraxis.tsx
+++ b/frontend/src/pages/editPraxis/archetypes/EphemeristsEditPraxis.tsx
@@ -447,12 +447,33 @@ export default function EphemeristsEditPraxis({ state }: Props) {
        renders into (hover 1.16:1). The detail column declares the identical
        pair through `.eph-plate-sheet`; this is the same seam reached from the
        one Ephemerists root that is not inside that column. NILE is the plate's
-       declared link hue and is measured on all three of its grounds. */
+       declared link hue and is measured on all three of its grounds.
+
+       AND THE LABEL SEAM, for the identical reason, one issue later (#1754 fixed
+       the sheet, #1800 this root — the #1754 agent found this one and was scoped
+       out of `editPraxis/**` at the time). Unset, `.label-caption` and
+       `.label-heading` inherit the global tertiary: on the sheet's `PLATE` that
+       is 2.01:1 in light and in a panel cell's `INNER` 1.86:1. QUIET is
+       `-plate-quiet`, the same ink `.eph-plate-sheet` names, so both Ephemerists
+       roots land on one value — 5.98 on the sheet, 5.52 in a cell, in both
+       cascades.
+
+       THE COMPOSER'S PAGE GROUND IS NOT THE PLATE'S, which is worth saying
+       because the plate's numbers do not transfer wholesale. `ComposerPage`
+       renders a bare `<div style={style}>` with no background, so what is under
+       the sheet here is the APP's `--color-bg-page` — near-white in light, where
+       QUIET reads 2.64:1 (the #1793 defect, on the faction page). Every label
+       this seam reaches renders inside `ComposerSheet` or a panel, and the one
+       thing on the bare page, the breadcrumb, takes `breadcrumbInk` explicitly.
+       A `.label-caption` mounted directly on this page would need the app's own
+       tier instead; there is none today, and the guard in
+       `factionContrast.test.ts` records why. */
     pageStyle: {
       fontFamily: DECO,
       color: INK,
       ["--link-ink" as string]: NILE,
       ["--link-ink-hover" as string]: INK,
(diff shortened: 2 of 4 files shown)


pixieofhugs:

**Review pass — no blocking findings, and the headline finding is correct.** I re-derived the two load-bearing numbers independently rather than taking them on trust.

## The impossibility proof holds

The claim is that no value of `--faction-ephemerists-plate-quiet` can clear 4.5:1 on both `-plate-inner` (`#1d2130`) and `#f7f4ee`. Recomputed from the WCAG relative-luminance definition:

- `#1d2130` → L = 0.01563. As the lighter of the pair, the ink needs L ≥ 4.5 × (0.01563 + 0.05) − 0.05 = **0.2453**.
- `#f7f4ee` → L = 0.9065. As the darker of the pair, the ink needs L ≤ (0.9065 + 0.05) / 4.5 − 0.05 = **0.1626**.

Floor above ceiling, and contrast is monotonic in luminance, so the satisfying set is empty for every hue. Both figures match the PR's to four decimals. **The issue as scoped could not be built, and saying so is the right outcome.**

## The diagnosis is right, not just the arithmetic

The decisive fact is that `--color-bg-page` is `#f7f4ee` in light — **exactly the ground the nightly scanner resolved**. So the two empty states genuinely sit on the app's page and never on a plate, and repointing them rather than moving the token is the root-cause fix. `--color-text-secondary` (`#554b3c` light) on that page recomputes to **7.78:1**, matching the PR.

That also makes this consistent with #1675 rather than a new rule: that issue already moved this page's section headings and kickers off plate inks for the same reason. The two empty states were simply left behind.

## Consequences checked

- The token value did not move, so `ephemeristsPlateSurfaces.test.tsx:103` and the pinned manifest rows in `factionContrast.test.ts` correctly needed no edits. Verified none were touched.
- The `--label-ink` addition to `EphemeristsEditPraxis`'s `dress.pageStyle` is the #1800 half and is asserted by a new guard.
- The false `/* AA on all 3 grounds */` comment is gone, replaced by the five grounds the token *is* measured on plus the one it is barred from. That was the actual disease — a hand-written contrast claim nobody re-derived.

## Follow-up filed, not folded in

The agent found `controls.tsx:1004` / `:1023` hardcoding `color: "var(--color-text-tertiary)"` on `.label-caption`, which defeats the seam this PR sets — 2.01:1, and cross-faction (S.N.I.D.E. and Singularity too). Correctly left out of scope; filed as #1819.

**Visual QA outstanding** — no browser in the agent's environment.

Issue #1740, opened 2026-08-15
A praxis room: mount the CRDT server, seed it once, and guard both auth doors


pixieofhugs:
Foundation issue for the ADR-0073 epic (Google-Docs-style co-editing of a praxis). **Nothing else in the epic can start until this lands.** Backend only — no frontend changes here.

Design and rejected alternatives: `docs/adr/0073-a-praxis-is-written-in-a-room-and-the-room-is-not-the-record.md`. This issue is self-contained; every decision it depends on is restated below.

## What a room is

The live editing space for one praxis: a server-held CRDT document plus the members currently connected. Every praxis gets one — solo, collab and duel parts alike.

## Scope

**Mount the room server.** `pycrdt` + `pycrdt-websocket`, as an `ASGIServer` route inside the existing FastAPI app. One process, one deploy, no second service.

**Seed exactly once, server-side.** When a room first opens, the server seeds the document from `praxis.body_text`. Clients must never seed. Two clients independently building a doc from the same text merge into *two copies of that text* — this is the standard Yjs duplication footgun and the single most important rule in this issue. It needs a test.

**Persist to Postgres.** `pycrdt-websocket`'s bundled stores are SQLite/file only. Subclass `BaseYStore` with `read`/`write` against Postgres. Include a squash policy — the store appends every update forever otherwise.

**Authorization, both doors:**

- *At connect* — `on_connect(msg, scope)` decodes the existing `access_token` httpOnly cookie, then runs the existing member check (`_require_member`, `services/praxis.py`) against the praxis id in the path. Returning `True` from `on_connect` rejects the socket before accept.
- *At connect* — **validate `Origin` by hand** against the same list `CORS_ORIGINS` feeds. `CORSMiddleware` (`main.py`) does **not** apply to WebSockets, so the allowlist protecting every REST route silently would not protect this one. Without this, a cookie-authed socket is open to cross-site WebSocket hijacking.
- *At revoke* — kick and leave **close the socket**. `on_connect` only checks membership at connect time; a removed member otherwise keeps writing until they close the tab. A gate needs both doors.

Auth transport is settled: the JWT is an httpOnly cookie, host-only on `api.worldzero.org`, and `worldzero.org` shares its registrable domain — so `SameSite=Lax` treats a `wss://api.worldzero.org` handshake as same-site and sends it. No ticket endpoint, no token in a URL.

[...]

Issue #1390, opened 2026-07-30
Three effects still dep on the whole user object — swap to user?.character?.id


pixieofhugs:
## What

The refetch-storm amplifier: five hooks dep on the **whole `user` object**, so every `useAuth().refetch()` (a new object each time) refires them all. One star cast = **5 requests to move one number**: `castVote` + `/auth/me` + requests-feed(limit 100) + global-feed(5) + `listPraxes(member)` — and each feed request is a ~35-query server-side fan-out. #1349 counts the 16 trigger sites; the `[user]`-object deps are the **multiplier**, and each is a one-line fix.

## The five hooks (audit-latency.md §6)

- `hooks/useMyActiveTasks.ts:28`
- `hooks/usePendingRequests.ts:29`
- `components/layout/Sidebar.tsx:80` (global feed)
- `pages/editPraxis/useEditPraxis.ts:643` (refires the full getPraxis→getTask chain — if the EditPraxis waterfall issue from this audit lands first, its dep fix covers this file; coordinate)
- `pages/CharacterProfile.tsx:81`

Fix: swap `[user]` → `[user?.character?.id]` — the pattern the codebase already uses deliberately (`useTaskDetail.ts:227-230`).

## Also bundled (same effect-hygiene family)

- **TaskDetail double `listRelationships`**: friend + foe as two requests differing only in `type` (`useTaskDetail.ts:181,193`) — one unfiltered call returns both (`routers/relationships.py:21`; CharacterProfile already calls it unfiltered). 2→1 requests.
- **Gate `useFieldDeskHome`'s fetching on form factor** (`pages/fieldDesk/useFieldDeskHome.ts:40-46`, consumed only by the mobile branch, `FieldDesk.tsx:71-74`): desktop `/` currently fires 2 requests that duplicate the sidebar's (`listPraxes(member_id)`, `getActivityFeed(requests,100)`) plus 1 nobody renders (`getVotesReceived`). Keep the hooks-rules constraint in mind — the hook still runs; its **fetching** is what gets gated (`useFormFactor()` is available). Note: #1344's compound endpoint is the deeper fix for the duplication; this is the cheap independent win.

## Cross-links

#1349 (trigger-site count — this is its amplifier), #1344 (sidebar wave / compound endpoint).

## Acceptance criteria

- A star cast on a praxis page with the sidebar mounted issues ≤2 requests (the POST + at most one deliberate refresh), verified in the network tab.
[...]

Issue #1232, opened 2026-07-29
factionContrast: add composer pairs for all eight skins in ONE PR


pixieofhugs:
Deferred deliberately during epic #1179, by the SNIDE agent, for a good reason worth preserving:

> Adding SNIDE's composer pairs there is the right long-term move, but it's a shared registry and seven agents are in flight — restating that row is the exact shape that reddened `main` before.

That refers to the failure where a shared test registry is asserted **both ways** (listed = registered, unlisted = Default), so parallel PRs that each *restate* the row drop their siblings: every branch green, `main` red only after squash.

## The work

Epic #1179 rebuilt all eight edit-praxis composers, each now drawing faction inks on faction grounds. None of those pairs are in `ARCHETYPE_PAIRS` in `frontend/src/utils/__tests__/factionContrast.test.ts`.

Add composer rows for **all eight at once, in a single PR** — na, UA, WOW, SNIDE, Ephemerists, Singularity, Everymen, Coven. Not one per faction; that reintroduces exactly the hazard above.

## Known readings to encode

Measurements the skin agents took while building, worth asserting rather than rediscovering:

- **WOW** — `--faction-wow-card-muted` is 4.77:1 on `--faction-wow-card-bg` but **4.25:1 on `--faction-wow-chronicle-panel`** (cream-only ink). `--faction-wow-stamp-chip-text` against `-chip-bg` is **4.14:1 in dark** — that pair is measured against `--faction-wow-duel-ribbon`, not against each other.
- **Everymen** — `--faction-everymen-sheet-accent` pays only 4.49:1 on `--everymen-paper`; red labels sit on `-sheet-panel` (4.95 light / 5.45 dark).
- **Singularity** — tightest are `dim` on the chrome bar 4.52/5.35, `dim` on the readout well 4.57/5.10, `blue-bright` on that well 4.68/9.50.
- **Coven** — `--faction-coven-cast-*` worst reading 5.04:1.
- **SNIDE** — the `PTS` caption in bright acid measures 1.35:1 on the composer stock; the light half uses a deep acid (4.7:1 on the slip field, 5.2:1 on the stock).

Verify each before asserting — these were taken at build time and the tokens may have moved.

pixieofhugs:
[...]

PR #1820, opened 2026-08-15
Presence: faction-coloured carets in the composer and a live dot on the roster

### LINKED ISSUE #1744: Presence: faction-coloured carets in the composer and a live dot on the roster
**Blocked by #1742** — carets need the CodeMirror binding to draw into.

Part of the ADR-0073 epic (Google-Docs-style co-editing of a praxis). Design and rejected alternatives: `docs/adr/0073-a-praxis-is-written-in-a-room-and-the-room-is-not-the-record.md`. Self-contained below.

This is the part players actually recognise as Google Docs — and the last issue in the epic.

## What presence is

Who is in a room right now, and where their cursor sits. **Decoration, never authorization**: presence state is self-reported by each client and relayed, so a client can claim to be any character. Low stakes — only fellow members of that praxis ever see it — but the caret label must never gate anything. Membership is the edit key, checked server-side (#1740).

## Scope

**Carets and selections in faction colour.** Awareness carries the character's name and faction slug; `y-codemirror.next` draws each co-author's caret and selection in *their own faction's* hue via `factionCssVar()`. No new colour decisions, no hardcoded hex, and it reads as this game rather than as a generic editor. Each faction has its own card archetype and this should not flatten that — a caret is chrome, so it follows the collaborator's identity, not the viewer's.

**A live "here now" dot in `CollabRoster`** (`frontend/src/components/collab/CollabRoster.tsx`), which already lists members. Presence is a state on an existing row, not a new list.

**Render nothing on a solo praxis.** Every praxis has a room (ADR-0073), so a solo author is alone in one; presence chrome must not appear for an audience of one. Duel parts are two separate `type=solo` praxes (ADR-0011), so they are solo for this purpose too.

**Dark mode** via the `[data-theme="dark"]` cascade — no `dark ? a : b` ternaries. Check caret contrast against every faction ground in both themes; a caret is a thin 1–2px mark and is the easiest thing in the epic to lose against a dark surface.

## Out of scope

Follow-mode (click a collaborator to track their viewport) — considered and dropped as a lot of surface across eight skins for something crews may rarely use. File it separately if anyone asks for it.

## Verification

- Static-render tests for the roster dot's presence/absence, which the `node`-environment harness *can* cover.
- One **nightly Playwright two-context spec** (`frontend/e2e/`, not PR-blocking by design — `.github/workflows/e2e.yml`): two browser contexts in one collab, each sees the other's caret, and the caret disappears on disconnect.
- Manual contrast pass across the eight faction grounds, light and dark.

pixieofhugs:
Presence in a praxis room: co-authors' carets and selections in their own faction's hue, and a live "here now" dot on the roster. Last issue in the ADR-0073 epic.

`Closes #1744`

## The seam I tested at

**`frontend/src/pages/editPraxis/roomPresence.ts`** — a new pure module holding the awareness→`characterId[]` derivation, the slug→paint, and the sanitizing proxy. It is exercised against a real `y-protocols` `Awareness`, with remote state relayed through the real `encodeAwarenessUpdate`/`applyAwarenessUpdate` encoder rather than by poking `awareness.states`, because JSON-through-the-socket is exactly the step that turns a *claimed* object into whatever the sender wanted. The roster dot is tested at the existing `renderToStaticMarkup` seam in `CollabRoster.test.tsx`.

Both loops went red on the real defect first (module missing; then "M1 is here now" absent from the markup).

## Security: the wire carries a slug, the receiver derives the colour

`y-codemirror.next` builds its caret as `style="background-color: ${color}; border-color: ${color}"` — **the remote's own `user.color` string, interpolated into a style attribute on your page.** Awareness is self-reported and relayed, so a co-member could send `red; position: fixed; inset: 0; background-image: url(…)` and take the viewport plus a load beacon. Membership is server-checked at connect (#1740), so the blast radius is people you invited — but that is not a reason to render an attacker-controlled declaration.

So the sender publishes `factionSlug` and the **receiver** derives `color`/`colorLight` through `factionCssVar()`, whose unknown-slug fallback resolves to `default`. The reachable set of colours is closed at nine whatever a remote sends. `paintedAwareness()` is a `Proxy` over `getStates()` **only** — `setLocalStateField`, `getLocalState`, `doc.clientID` and the `on`/`off` the caret plugin subscribes with all reach the real `Awareness` bound and untouched, and the `WebsocketProvider` keeps the raw object so our derived `var()` strings never go on the wire. The remote `name` is clipped in the same pass (not an injection — the library uses `textContent` — but unbounded text in an absolutely-positioned `nowrap` label). `presence` reaches no authorization branch anywhere.

Two related library facts the tests pin: `colorLight` **must** be supplied, because the library's `color + '33'` fallback produces the invalid `var(--faction-coven)33` and paints no selection at all; and `yCollab` skips `clientid === awareness.doc.clientID`, so **a solo author alone in their room draws nothing by construction** — no `praxisType` gate needed on the caret side.

## The roster dot

`presentCharacterIds?: readonly number[]` is a **prop, not a derivation**. `CollabRoster` is mounted in 11 places and only the composer mount sits inside `PraxisRoomProvider`; the other ten get "nothing known" for free, and **absent means no dot markup at all** — an empty dot column on a public detail page would read as a crew that had left.

It is a bare filled badge on the **avatar's corner**, not a chip: connected is ephemeral awareness that vanishes with a tab, while `StatusPill` is persistent workflow state, and only one of them means the publish is waiting on somebody. Ink is `card-credit`, not `card-accent` — the row already spends the accent on *done*. The badge carries no word, so it has `role="img"` + `aria-label`/`title`. The one new string lives in `editPraxis.composer` (the shared, voiceless block beside `collaboratorsLabel`), not `editPraxis.collab`, per the owner's shared-copy ruling.

Only a `row.member` can ever light up, so an invited or declined row cannot — pinned with `presentCharacterIds={[1, 2, 107]}` against the invite whose `invitee_id` is 107.

## Files

- **new** `frontend/src/pages/editPraxis/roomPresence.ts`, `frontend/src/pages/editPraxis/__tests__/roomPresence.test.ts` (12 tests)
- `frontend/src/pages/editPraxis/praxisRoom.tsx` — `awareness` lives beside `body`/`meta` so its identity is stable; `present` is derived behind a **value** guard (awareness `change` fires on every cursor move, so without it a co-author holding a key re-renders the roster per keystroke); the local identity is published from a **second** effect keyed on the viewer, so a character switch re-publishes without tearing the socket down. Deps are narrowed to `id`/`display_name`/`faction_slug` per #1390.
- `frontend/src/components/collab/CollabRoster.tsx` + its test — the prop and the badge
- `frontend/src/pages/editPraxis/archetypes/controls.tsx` — `yCollab(ytext, paintedAwareness(awareness))`, `presentCharacterIds={room?.present}`
- `frontend/src/locales/en/forms.json` — one key
- `frontend/e2e/collaboration.spec.ts` — a `presence` describe block

No backend change and no new API endpoint consumed. `PraxisMemberOut` carries no faction slug, which is why the slug rides awareness — so nothing regenerates `openapi.json` / `schema.d.ts`.

## Verification

Every step in the CI `frontend` job, run locally: `npm run lint` ✅ · `npm run typecheck` ✅ · `npm run typecheck:design-sync` ✅ · `npm test` ✅ (249 files, 4447 passed) · `npm run build` ✅ · `npm run budget` ✅ (JS 130.0 KB, CSS 21.3 KB, both within budget).

**The Playwright spec has never executed.** `npx playwright test --list` collects both new tests (`collaboration.spec.ts:402` and `:445`); nothing here ran them, and they are nightly-only by design (`e2e.yml`), outside PR CI.

**A manual contrast pass across the faction grounds in both themes is owed to a human.** A caret is a 1–2px mark and is the easiest thing in this epic to lose against a dark surface; I have no browser and cannot measure it.

## For `frontend-style`

`.cm-ySelectionInfo` — the hover label — keeps the library's hardcoded **white** ink on the remote's faction hue. It is measured against the *remote's* colour, so no single ink is right for all nine, and the `-on-accent` tier does not cover them (#1232 deleted Ephemerists'). Per the owner's ruling this PR accepts white rather than minting a tier as a byproduct; it is marked with a `ponytail:` comment in `roomPresence.ts` naming the upgrade path (mint the tier in `index.css`, override the label's `color` in `BODY_EDITOR_BASE_THEME`). It is a hover-only tooltip.

## What to eyeball

1. **Two browsers, one collab.** Open the same collab composer as two members. Each should see the other's caret in the *other's* faction colour, a tinted selection band when they select text, and their name on hover. Close one tab: the caret and the dot should both go.
2. **Caret contrast on every faction ground, light and dark** — eight composer skins, and a 1–2px mark. This is the pass that most needs a human.
3. **The selection band.** `color-mix(… 22%, transparent)` over each skin's field — legible text through it, visible band.
4. **Dot vs pill.** On a row where somebody is connected *and* has cast, the corner badge and the status pill must not read as two versions of the same statement.
5. **A solo praxis draws nothing** — no caret, no roster.
6. **No jitter.** Type a paragraph while a co-author types: the caret must not reflow the line, and the roster must not flicker.

### DIFF
--- a/frontend/e2e/collaboration.spec.ts
+++ b/frontend/e2e/collaboration.spec.ts
@@ -378,3 +378,89 @@ test.describe('collaboration UI (clicked buttons)', () => {
     }
   })
 })
+
+/**
+ * Presence (#1744) — the only assertions in this repo that can see a caret.
+ *
+ * Everything else about presence is unit-testable: the derivation, the paint
+ * and the sanitizing proxy in `roomPresence.test.ts`, the roster dot in
+ * `CollabRoster.test.tsx`. What no unit test can reach is the actual claim —
+ * that two real browsers in one room draw each other, and that closing one
+ * takes its caret away. That needs two contexts, a live WebSocket and the
+ * CodeMirror plugin, so it lives here (nightly, `.github/workflows/e2e.yml`).
+ *
+ * Locators are the library's own class names: `.cm-ySelectionCaret` is the
+ * widget span, `.cm-ySelectionInfo` the hover label carrying the name. Each
+ * page sees only the OTHER player's caret — `y-codemirror.next` skips the local
+ * client id, which is also why a solo author alone in their room draws nothing.
+ *
+ * A block rather than a new file: `login` / `seedCollabDraft` are module-local
+ * helpers, and exporting them to reuse them elsewhere would be a wider change
+ * than the tests are worth.
+ */
+test.describe('presence: carets and the roster dot', () => {
+  test('P1: each member sees the other caret, and it goes when they leave', async ({
+    browser,
+  }) => {
+    const seed = await seedCollabDraft(browser, 'presence')
+    try {
+      const aPage = await seed.alice.ctx.newPage()
+      const bPage = await seed.bob.ctx.newPage()
+      await aPage.goto(`/praxis/${seed.praxisId}/edit`)
+      await bPage.goto(`/praxis/${seed.praxisId}/edit`)
+
+      // A caret exists only once its owner's cursor is IN the document, so both
+      // have to put a selection there. `fill` auto-waits on the editor becoming
+      // editable, which is the room finishing its first sync.
+      await aPage.locator('.cm-content').first().fill('Alice is typing')
+      await bPage.locator('.cm-content').first().fill('Bob is typing')
+
+      // Alice sees exactly one remote caret: Bob's, labelled with his name.
+      await expect(aPage.locator('.cm-ySelectionCaret')).toHaveCount(1)
+      await expect(aPage.locator('.cm-ySelectionInfo')).toHaveText(seed.bob.name)
+      await expect(bPage.locator('.cm-ySelectionInfo')).toHaveText(seed.alice.name)
+
+      // The roster's live dot names the same fact in words, on Bob's row.
+      await expect(
+        aPage.getByRole('img', { name: `${seed.bob.name} is here now` }),
+      ).toBeVisible()
+
+      // Presence is ephemeral: closing the tab closes the socket, and the caret
+      // must go with it. This is the half that distinguishes "he's not here"
+      // from the persistent workflow state beside it.
+      await bPage.close()
+      await expect(aPage.locator('.cm-ySelectionCaret')).toHaveCount(0)
+      await expect(
+        aPage.getByRole('img', { name: `${seed.bob.name} is here now` }),
+      ).toHaveCount(0)
+    } finally {
+      await seed.alice.ctx.close()
+      await seed.bob.ctx.close()
+    }
+  })
+
+  // Presence chrome must not appear for an audience of one (ADR-0073). Nothing
+  // gates on the praxis type: the plugin skips the local client, so a solo
+  // author is alone in a room that draws nobody.
+  test('P2: a solo author alone in their room draws no caret', async ({ browser }) => {
+    const s = pairSeq++
+    const solo = await login(browser, `p-${RUN}-${s}`, `P${s}-${RUN}`, 0)
+    const task = await pickOpenTask(solo)
+    try {
+      const created = await solo.ctx.request.post(`${API}/praxes`, {
+        data: { task_id: task.id, type: 'solo', title: `Solo ${s}`, body_text: 'draft' },
+      })
+      expect(created.ok(), `solo create failed: ${await created.text()}`).toBeTruthy()
+      const praxis = await created.json()
+
+      const page = await solo.ctx.newPage()
+      await page.goto(`/praxis/${praxis.id}/edit`)
+      await page.locator('.cm-content').first().fill('Just me in here')
+      await expect(page.locator('.cm-ySelectionCaret')).toHaveCount(0)
+      // And no roster at all, which `CollabRoster` already gates on positively.
+      await expect(page.getByRole('img', { name: /is here now/ })).toHaveCount(0)
+    } finally {
+      await solo.ctx.close()
+    }
+  })
+})
--- a/frontend/src/components/collab/CollabRoster.tsx
+++ b/frontend/src/components/collab/CollabRoster.tsx
@@ -60,6 +60,13 @@ export type { CollabGate, CollabState } from './collabGate'
 const AVATAR_SIZE = 34
 /** The design's 6px status dot. */
 const DOT_SIZE = 6
+/**
+ * The "here now" badge on the avatar's corner (#1744), and the ring that lifts
+ * it off the monogram's edge. Bigger than DOT_SIZE because it stands alone on a
+ * 34px circle rather than inside a pill that is already a word wide.
+ */
+const PRESENCE_DOT_SIZE = 10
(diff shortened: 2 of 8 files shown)


pixieofhugs:

**Review pass — no blocking findings.** I traced the security argument to its root rather than accepting the summary, since this PR is the first place an *untrusted* string reaches `factionCssVar()`.

## The injection defence holds, verified end to end

1. **The threat is real and correctly characterised.** `y-codemirror.next/src/y-remote-selections.js:88` builds the caret as `pair.create('style', \`background-color: ${this.color}; border-color: ${this.color}\`)`, and `:206`/`:217`/`:226`/`:236` do the same with `colorLight`. Both come from `state.user` at `:194-195` — the remote's object.
2. **The set is genuinely closed.** `factionCssVar()` routes through `resolveCssKey()`, which is `CSS_KEY[slug ?? ""] ?? "default"` — a **lookup table**, not string interpolation. An arbitrary slug cannot reach the `--faction-*` name. This is the load-bearing fact and it checks out.
3. **The wholesale `user` replacement is the right shape.** `paintStates` spreads `state` then overwrites `PRESENCE_FIELD` entirely, so a remote-supplied `color`/`colorLight` cannot ride along beside the derived pair. Spreading rather than rebuilding is also correct: `cursor` must survive verbatim to position the caret.
4. **Nothing else from `state` reaches the DOM.** Grepped the library for `style`/`setAttribute`/`innerHTML`/`cssText` — the only interpolations are the four `colorLight` bands, the caret's two, and `name` via `dom.text()` (textContent). So preserving unknown top-level keys through the spread is safe.
5. **Proxying `getStates()` alone, and keeping the raw `Awareness` on the provider, is the correct split.** The protocol encoder reads `awareness.states` directly, so painting the broadcast view would put our derived `var()` strings on the wire as somebody's identity. The docblock says exactly this.

## Nit, not blocking, and not this PR's bug

`resolveCssKey` uses `CSS_KEY[slug] ?? "default"`, which reaches the prototype chain: a remote sending `factionSlug: "constructor"` yields `Object`, and the var name becomes `--faction-function Object() { [native code] }`. **Not exploitable** — no prototype member stringifies to anything containing `;`, so it is a malformed declaration the parser drops, and the caret simply loses its colour. But `factionCssVar` has until now only ever seen server-supplied slugs, and this PR is what makes it reachable from the wire. A `typeof key === "string"` guard (or `Object.hasOwn`) in `factions.ts` would close it. Worth a follow-up ticket rather than a change here.

## Also verified

- `presentCharacterIds` validates `Number.isInteger(id) && id > 0` before trusting a remote's id.
- `samePresence` earns its place — the awareness `change` event fires per cursor move, and without the value guard a co-author holding a key re-renders the roster and every composer archetype under it.
- `name.slice(0, 64)` is correctly described as a layout ceiling, not a security one.
- The `ponytail:` comment on the white `.cm-ySelectionInfo` ink names both the ceiling and the upgrade path, per the ruling I gave.

**Outstanding and stated in the PR, correctly:** the Playwright two-context spec has never executed (collected only), and the manual contrast pass across the faction grounds in both themes is owed to a human — a caret is a 1–2px mark.

Issue #1792, opened 2026-08-15
The sky score numeral fails AA for four faction hues, not one


pixieofhugs:
The sky/meadow score numeral fails AA for **four** faction hues, not one. #1766 item 1 was scoped on the belief that it was one, and its "mint no new token" ruling rests on that belief.

## The measurement

Nightly `workflow_dispatch` run [31867605613](https://github.com/pixieofhugs/WorldZeroPlayground/actions/runs/31867605613) (2026-08-15) — the **first** run to execute #1779's report-and-ratchet mechanism. Every one of these is a *measured* failure, not an unresolved backdrop.

All four share one selector, so this is one component painted with four different hues:

`div.rounded-2xl.overflow-hidden > div > a.font-display > span.font-body`, 18px/700

| Token | Hex | Ground | Ratio | Needs | Combos |
|---|---|---|---|---|---|
| `--faction-ua` | `#c24a18` | sky `#0f0e15` | 3.92 | 4.5 | **all 14 light** |
| `--faction-singularity` | `#2563eb` | sky `#0f0e15` | 3.71 | 4.5 | singularity/light/desktop |
| `--faction-everymen` | `#c1272d` | sky `#0f0e15` | 3.29 | 4.5 | everymen/light/desktop |
| `--faction-ephemerists` | `#1d6e72` | sky `#0f0e15` | 3.22 | 4.5 | ephemerists/light/desktop |

The UA row hits every light combination because a UA character appears in the shared routes (the leaderboard), so its numeral paints regardless of the viewer's faction. The other three appear only on their own `/factions/<slug>` surface.

## Why #1766 saw only one

#1766 was measured on run 31779247838, **before** #1762 deleted the 56 `ratio: null` allowlist entries and before #1779 landed. Those three hues were being filtered out of the report, not clearing it. The data changed under the issue; nobody did anything wrong.

## What this changes

#1766 states:

> Only UA's hue fails on `--sky-bg`; the other six clear it, so this is one faction's hue, not a systemic sky problem.

That is now false — it is **four of seven**, at one site, painted by `factionCssVar(character.faction_slug)` in `Constellation.tsx` and `Meadow.tsx`.

[...]

Issue #1783, opened 2026-08-15
Button and card-meta label size is class-owned, so 20 inline restatements cannot be swept individually


pixieofhugs:
> *This was generated by AI during /builder-bot.*

Surfaced by **#1608** (PR #1782), which swept 188 label-tier sites from the two smallest steps and deliberately left 25 behind. **20 of those 25 are this one finding.**

## What is true

`frontend/src/index.css` still declares, as class values:

- `.btn-primary` / `.btn-outline` — `--text-sm` (**9px**), uppercase, letter-spaced
- `.card-meta` — `--text-xs` (**8px**), uppercase, tracked

Twenty of the sites #1608 left are **inline restatements of those exact class values** — a component hand-rolling the button's own font size next to the class rather than instead of it.

## Why #1608 was right not to touch them

Raising a restatement individually **desynchronises that button from every other button on the site.** The inline value and the class value are the same number by intent; changing one of them silently forks the tier. So this cannot be swept per-site — it has to move at the class, and moving the class moves **every button and every card meta line at once.** That is the design call #1608's ruling explicitly scoped out.

`pages/praxisDetail/shared.tsx:611` was left for a sharper version of the same reason: a hand-rolled confirm button stands **beside** a `.btn-outline` cancel. Raising one of a visible pair is worse than raising neither.

## What this issue is

**A ruling, then a two-part change.**

1. **Decide whether button and card-meta chrome joins the label-tier floor.** #1307 moved `.label-heading` to `--text-md` (11px) and `.label-caption` to `--text-lg` (12px). Buttons at 9px uppercase-tracked carry the same four stacked legibility costs the tier work exists to remove — and unlike a caption, a button label is text a player must read *to act*. The counter-argument is that button chrome is a different tier by intent, and #769 already ruled seal buttons stay `--text-sm` as button chrome.

2. **If it moves:** change the class in `index.css`, then delete the 20 inline restatements rather than editing them — they should never have restated the class. That is a deletion, not a swap, and it removes the fork permanently.

**Do not** raise the inline sites while leaving the class where it is. That is the one outcome worse than doing nothing.

## Related, and probably the same pass

[...]

Issue #1449, opened 2026-07-31
Flagged and failed badges read the same ink on faction cards — mint -card-alarm beside -card-notice


pixieofhugs:
Follow-up to #1302, which fixed the *readability* of admin badges on faction card sheets but collapsed two meanings into one ink doing it.

## The collapse

PR #1411 repointed the card's admin badges from global `--color-danger` / `--color-warning` to `notice: factionCssVar(slug, "card-notice")` (`frontend/src/components/praxisCard/shared.tsx:98`). That was right — global alarm ink on faction paper was the bug.

But `card-notice` is **one** token, so **a flagged badge and a failed badge now render identically**, and so do the adjacent hide-versus-fail controls. `praxisCard/shared.tsx:82` carries a `ponytail:` comment naming the collapse as deliberate-for-now, and `git grep card-alarm` on `origin/main` returns nothing — so no second ink was ever minted.

## Owner ruling, 2026-07-30: mint the second ink

An admin scanning a moderation feed is doing exactly the at-a-glance triage colour exists for, and this is the only place in the app where two alarm states sit adjacent to each other. The counter-argument — that the badges are textual so WCAG 1.4.1 already holds — was considered and overruled: compliance was never the problem, misclicking hide-versus-fail at speed is.

## The work

Mint `--faction-<slug>-card-alarm` across **all eight faction keys in both cascades**, then have the badge dispatch pick `card-alarm` for the destructive state and keep `card-notice` for the cautionary one.

Follow `-card-notice` exactly — it already exists for all eight keys in both themes, so this is a registry row and roughly sixteen CSS lines, not a doctrine change. No new mechanism.

**Measure against `-card-bg`.** That is the ground `-card-notice` and `-card-credit` are measured on, and the reason the card plate's family cannot be reused elsewhere (see the sibling issue for praxis detail, where S.N.I.D.E.'s `card-notice` reads 1.31:1 against the detail wall). `frontend/src/__tests__/factionContrast.test.ts` is the guard — add the new pairs there or the token family grows an unmeasured member.

## Not this issue

The praxis-detail half of #1302 — roughly fifteen sites in `pages/praxisDetail/shared.tsx` still painting global danger ink on faction sheets. That needs a per-**page** notice family and is filed separately.

[...]

Issue #1341, opened 2026-07-30
The feed row's actor name fails AA on all seven chassis — four need a -quiet mint, three are done


pixieofhugs:
The actor's name in a feed row — "Dave submitted a praxis" — paints `--faction-{key}` at 18px/700, which puts it under the **4.5:1** body-text floor (700 weight at 18px does not reach the 24px/18.66px-bold large-text exemption).

It **fails on all seven bespoke chassis in light**, and one in dark.

## Measured

Found and machine-measured with `frontend/src/utils/contrast.ts` during #1252 (PR #1340) — every ratio computed, none hand-derived.

| chassis | light | dark |
|---|---|---|
| Warriors of Whimsy | **1.96:1** | — |
| S.N.I.D.E. | **2.41:1** | — |
| Cozy Coven | **2.98:1** | — |
| The Singularity | **3.67:1** | — |
| UA | **4.04:1** | — |
| The Ephemerists | **4.15:1** | **3.16:1** |
| The Everymen | **4.49:1** | — |

WOW at 1.96:1 is the worst reading; Everymen at 4.49 misses by 0.01.

## Three are already fixed — this issue is the other four

#1252 built the feed-row ink seam and used it to repoint the three factions that have an **existing** accent clearing the floor on their own chassis: **UA, WOW, Singularity**. Those are done and not in scope here.

**The remaining four are S.N.I.D.E., Cozy Coven, The Ephemerists and The Everymen.** They were deliberately left because **no existing token in their families survives their own ground** — nothing clears the SNIDE slip or the Ephemerists plate. So this is not a repoint; it needs a token **mint**, which is a different decision and a different diff from the seam work.

## The established shape for exactly this

**#1173's `-quiet` sibling.** That issue minted a per-faction quiet ink for the same situation — a role that needs to carry small text on a ground its existing inks do not clear. Follow it rather than inventing a fifth mechanism.

`WORLD_ZERO_STYLE §3` governs the alternative: **do not deepen a shared token to suit one ground.** The actor name is read on seven different faction grounds, so there is no single value that is correct on all of them — that is the whole reason the ink seam exists.

## The seam is already there

`components/feed/feedRowSkin.ts` (#1252) publishes an `ink` bag with an `actor` slot, resolved field-by-field. A faction supplying nothing keeps today's rendering. So the work here is: mint the token per faction in `index.css` (both cascades), publish it from that faction's feed frame, and assert the pairing.

[...]

PR #1822, opened 2026-08-15
Three accents move off the grounds they cannot clear, and one stops being a gradient stop

### LINKED ISSUE #1766: Four faction accents are still used as text on grounds they cannot clear
The last four measured failures in the nightly contrast sweep, after #1749
retired the other nineteen. All four are the same shape: **a faction accent hue
used as text on a ground that is not its fill.** Fixing these takes
`contrast.spec.ts` to zero measured failures and lets #1675 close.

## Owner ruling, 2026-08-14: fix site by site, mint no new token

The tier that would generalise this — a per-faction "hue guaranteed to clear AA
as text on the app's page grounds" — is genuinely missing, since
`--faction-*-on-fill` / `-on-accent` are inks to put ON a fill, not this. It is
**not** being built here. Three of these four are violations of a rule the repo
already documents, and the fourth is one site; a seven-faction token tier is a
larger, design-reviewed piece of work that should not arrive as a byproduct of a
test fix.

## The four

Measured in nightly run 31779247838 (2026-08-14).

| Site (selector from the run) | Colour | Ground | Ratio | Needs |
|---|---|---|---|---|
| `div.rounded-2xl.overflow-hidden > div > a.font-display > span.font-body`, 18px/700 | `--faction-ua` #c24a18 | sky #0f0e15 | 3.92 | 4.5 |
| `section.mt-6 > div.flex.flex-col > a > span`, 14px/600 | `--faction-ua-vermil` #d9542a | #241e18 | 4.12 | 4.5 |
| `div > div > span > span` "points", 10px | #ca8a04 | #ffffff | 2.94 | 4.5 |
| `div.page > div > div.flex.justify-end > button.hover:opacity-80.active:opacity-60`, 12px | #c1272d | #ddceac | 3.75 | 4.5 |

### What each one actually is

1. **The sky/meadow score numeral.** `Constellation.tsx` and `Meadow.tsx` paint
   it with `factionCssVar(character.faction_slug)`. Only UA's hue fails on
   `--sky-bg`; the other six clear it, so this is one faction's hue, not a
   systemic sky problem. Wants a darkened UA hue scoped to that surface.
2. **A misuse, not a missing tier.** `--faction-ua-vermil` is declared in
   `index.css` with the comment "the ensō score numeral — large display type
   only", and this site is 14px/600. Give it UA's normal ink.
3. **and 4. are type-scale violations as well as contrast ones.** 10px and 12px
   sit far under the content floor (WORLD_ZERO_STYLE §4; see also #769, where
   duel narration sinking below that floor was the whole bug). Fixing the size
   is part of the fix, not a separate cleanup.

## Do NOT

- Mint a per-faction on-page ink tier. Ruled out above.
- Take the "bump 18px to 19px so WCAG's bold-large bar applies" route on item 1.
  It is technically compliant and it was explicitly rejected: the numeral would
  be no more legible, only exempt.
- Mark any of these `aria-hidden` to duck the scanner. Same reason.

## Verify before you edit — these numbers are already a day old

Do not trust the table above as current. It was measured on 2026-08-14, and
main has moved since: #1749 changed several inks and #1753 repointed
`--color-text-secondary` and `--color-text-tertiary` in both themes. Re-derive
from the latest nightly before changing a value —
`gh run view <id> --log-failed`, strip the runner's two leading columns, split
on `<faction>/<theme>/<viewport>: text below WCAG AA`, and dedupe the way the
spec does. Measure every candidate on the ground the text actually sits on, not
the one the token's name implies.

## Done when

`contrast.spec.ts` contributes zero MEASURED failures on a nightly run. The
unmeasurable side is #1675's ruling and #1762's cleanup, not this.

## Not this issue

- The unmeasurable-report gaps — #1762.
- Any change to a faction's identity colour where it is used as a FILL.

*Filed by AI. Ruling by the owner, 2026-08-14.*

pixieofhugs:
Items **2, 3 and 4** of #1766. Item 1 is not here — it moved to #1792 when the
2026-08-15 re-measurement found four faction hues failing at that selector
rather than one, which collapsed the ruling it rested on. Item 4's **size** is
not here either: that is #1783's ruling and it runs after this.

Standing owner ruling honoured — **fixed site by site, no per-faction on-page ink
tier minted.**

## The seam I tested at

The token-value seam: `factionContrast.test.ts` resolves `index.css` through the
cascade and measures a declared ink against a declared ground. That is where
these three live, because all three are a **value on a ground**, and the nightly
sweep can only tell you after the fact. Each of the four new rows was confirmed
**red on the pre-fix values first** (restore the old value, the row fails;
restore the fix, green).

The one thing that seam cannot see is a token used *outside its declared scope*,
which is exactly item 2 — every ratio in the file stayed green while UA painted a
display-only ink at 14px. That half gets a reader allowlist instead.

## Re-derived measurements

All computed with `utils/contrast.ts` (`parseColor` → `compositeOver` →
`contrastRatio`) against the *composited* ground, on `origin/main` at 218bfd5a
and re-checked after rebasing onto 3f2573c1. Every number in the issue
reproduced exactly.

| # | Measured at | Ink | Ground | Before | After |
|---|---|---|---|---|---|
| 3 | `components/praxisCard/scoreStamp/DefaultScoreStamp.tsx:188` — `POINTS`, 10px, on `--faction-default-stamp-bg` | `--faction-default-gold` | `#ffffff` | **2.94** | **5.00** |
| 3 | `pages/taskDetail/archetypes/DefaultTaskDetail.tsx:276` and `AlbescentTaskDetail.tsx:153`, on `--faction-default-card-bg` | `--faction-default-gold` | `#fffdf9` | **2.89** | **4.92** |
| 3 | the same ink on the page ground | `--faction-default-gold` | `#f7f4ee` | 2.68 | 4.56 |
| 3 | the consumer that could not follow — `SingularityFactionBody.tsx:43`, `SingularityFactionPage.tsx:36`, `SingularityFieldDesk.tsx:39` | now `--faction-singularity-amber` | `#050f08` | 6.63 | **6.63 (unchanged)** |
| 2 | `pages/factionDetail/mobileArchetypes/UaFactionPage.tsx:303` — member-row rank, 14px/600, on `--faction-ua-card-bg` | `--faction-ua-vermil` → `--faction-ua-card-accent` | `#241e18` (dark) | **4.12** | **6.58** |
| 2 | the same site by day | ditto | `#f7e7d2` | 4.94 | 5.18 |
| 4 | `pages/fieldDesk/mobileArchetypes/EverymenFieldDesk.tsx` `actionPillStyle`, 12px, on `--everymen-paper-deep` | `--everymen-red` → `--everymen-deep-accent` | `#ddceac` | **3.75** | **5.03** |
| 4 | the same pill at night | ditto | `#17110d` | 4.55 | 6.66 |
| — | **found while here**, same file: the "view all" link, 11px on the Plate's `--everymen-paper` | `--everymen-red` → `--everymen-paper-accent` | `#ece1c6` / `#221a16` | **4.49 / 4.16** | **5.09 / 6.09** |

## What changed, and why each way

**Item 3 — `--faction-default-gold` stops being stop 3.** It was
`var(--faction-default-stop-3)`, and the borrow *was* the defect: a ramp position
is chosen for how it sits between its neighbours, not to be read. The token's own
comment already warned that "a gradient must not read it back"; the half nobody
had written down is that the borrow is equally wrong in its own direction. It is
now a literal — same amber, hue held exactly (h 40.6°, channels scaled on
lightness), walked to 5.00 on the white score plate. Worth taking first: it
failed **all fourteen light combinations on its own**, because the na family is
*site chrome* and every walked route paints it.

`--faction-default-stop-3` is deliberately unmoved. It is live in `-rainbow`,
`-rainbow-vertical`, `-rainbow-loop`, `-rainbow-conic`, `-total-rainbow`,
`--badge-victor-stop-4` and two radial washes; darkening it to fix a caption
would smear seven surfaces to mend one. Breaking the alias also broke the theme
tracking, so the dark half is now declared — at `#e6b94f`, exactly what the dark
stop resolved to, so **nothing repaints at night**.

**Item 3's second half — Singularity had to get its own gold.** The same token
inked three Singularity surfaces on a terminal that is near-black in *both*
themes, so the two consumers wanted opposite directions. The band is provably
empty rather than merely tight: white needs luminance ≤ 0.183 and the terminal
≥ 0.191. `--faction-singularity-amber` carries **exactly what the alias resolved
to in each cascade**, so no Singularity pixel moves — and all three files already
carried the comment *"if Singularity's credits panel ever wants its own gold,
that is a `--faction-singularity-*` token, not a brand one."*

**Item 2 — a misuse, not a missing tier.** `--faction-ua-vermil` is declared
"the ensō score numeral — large display type only" and the member row set it at
14px/600, which owes 4.5 and not the large-text 3. It takes
`--faction-ua-card-accent`, which `ua card accent` in `CARD_PAIRS` has measured
on that exact ground since #848 — so no new row and no new token. The four
`-vermil` consumers the issue said to leave alone are untouched; `UaFactionPage`
drops off that list because the const's only use *was* the defect.

**Item 4 — ink only, deliberately.** `actionPillStyle` already reads
`fontSize: 'var(--text-lg)'` — 12px is the label ramp's own step, not a raw
literal under the content floor, so the issue's "type-scale violation" framing is
stale. **I left the size alone on purpose: whether button chrome moves up a step
is exactly what #1783 asks, and deciding it here would fork that ruling.** The
ink takes `--everymen-deep-accent`, the family's third `(role, ground)` split
after `-sheet-accent` (the broadsheet panel) and `-paper-accent` (#1341's paper).
Neither existing rung survives this stock: `-paper-accent` is 4.25 here and
`-red-deep` clears by day at 5.82 and collapses to 2.92 at night.

**One un-reported sibling, fixed.** The same file's "view all" link is 11px in
the bare red on the Plate's paper — 4.49 / 4.16, under AA in *both* themes. It is
the identical pairing #1341 already minted `--everymen-paper-accent` for, and a
site that ink had not reached, so it was a one-word swap. Called out because it
is not in the issue's table.

## The guard left behind

Four new rows in `frontend/src/utils/__tests__/factionContrast.test.ts` — the
file this repo pins measured pairs in — plus a **reader allowlist** for
`--faction-ua-vermil`, since a scope stated in a comment is a rule an editor has
to remember and item 2 is proof they will not. Non-ratio assertions beside the
ratios are the shape #1449 and #1413 already set in that file.

Nothing was added to `BASELINE`; that list only ever shrinks.

## Verification

Every step in `.github/workflows/test.yml`'s `frontend` job, locally, after the
rebase: `npm run lint` (incl. the `no-raw-style-values` ratchet) · `npm run
typecheck` · `npm run typecheck:design-sync` · `npm test` (**4446 passed**, 248
files) · `npm run build` · `npm run budget` (JS 130.0 KB, CSS 21.4 KB — both ok).
The built stylesheet was grepped directly rather than trusted from source, since
a dropped brace proves nothing: `--faction-default-gold: #966603 / #e6b94f`,
`--faction-singularity-amber: #ca8a04 / #e6b94f`, `--everymen-deep-accent:
#a01c21 / #f4726c`, and `--faction-default-stop-3` still `#ca8a04 / #e6b94f`.
The `test` and `api-schema` jobs are untouched — no Python, no route signature,
no schema.

**Visual QA is outstanding.** I have no browser and no dev stack; every number
here is arithmetic on declared values, and nothing has been looked at.

## What to eyeball

1. **na / light / desktop + mobile** — any praxis card's score stamp. The `POINTS`
   caption under the struck disc is the darkened gold on the white plate. This is
   the only *visible* repaint in the PR and the one to judge: does it still read
   as gold rather than as brown?
2. **na / light** — a task detail page (`DefaultTaskDetail`), the "total" caption
   beside the points figure; same ink, on the cream sheet rather than the plate.
3. **albescent / light** — a task detail; the caption inside the turning prism
   ring is the same ink again, on the same cream.
4. **na and albescent / dark** — the same three. They must be **unchanged**;
   `#e6b94f` is what the alias already resolved to.
5. **singularity / light and dark / desktop + mobile** — the faction page's member
   roster readouts, and the mobile field desk's amber pill. All four must be
   **identical** to today; if anything moved, the split copied a value wrong.
6. **ua / dark / mobile** — the faction page's Top Members rows. The rank numeral
   left of each medallion is now UA's sienna accent rather than vermilion. Compare
   it with the "member" badge on the same page, which is the same ink.
7. **ua / light / mobile** — same rows, same question.
8. **everymen / light / mobile** — the field desk home. The CHARACTERS and EDIT
   chits at the top of the plate (deeper red now), and the "view all" link beside
   "On the job". The big score figure and the masthead band must **not** have
   moved — they keep `--everymen-red`.
9. **everymen / dark / mobile** — the same two, which lift rather than deepen.

Closes #1766


pixieofhugs:

**Review pass — no blocking findings.** CI green on all three required checks, both runs. I recomputed the two new inks from the WCAG definition rather than trusting the table.

## The numbers check out

- **`--faction-default-gold: #966603` on white** → L = 0.1600, ratio **5.00:1**. Clears, and sits under the 0.183 luminance ceiling the PR derives for a white ground.
- **`--everymen-deep-accent: #a01c21` on `#ddceac`** → L = 0.0842 against the ground's 0.6249, ratio **5.03:1**. Clears.

## Breaking the alias was the right call, and the file asked for it

`--faction-default-gold` was `var(--faction-default-stop-3)`. The comment already sitting above that declaration says the token *"is named for the caption it paints rather than for a position in a ramp — a gradient must not read it back."* The alias meant it was **simultaneously** both, so the ink was pinned to whatever the rainbow needed. Splitting it and leaving `--faction-default-stop-3` untouched keeps all seven gradient cuts pixel-identical while freeing the caption to be legible. Confirmed the stop is not modified.

Minting `--faction-singularity-amber` at exactly the value the gold used to resolve to is the correct companion move: it repaints nothing, and it makes the coupling visible instead of implicit.

## Item 2's fix is the durable one

The vermilion reader allowlist in `factionContrast.test.ts` is worth more than the ink change itself. The PR's framing is right: **a manifest asserts the pairing a token's declaration describes, never the one a component actually drew**, which is exactly how `--faction-ua-vermil` could say "large display type only" while UA painted it at 14px with every manifest row green. A declared scope with no reader guard is a comment, and this repo has now been bitten by a comment-as-guarantee twice today (see #1793).

## One disagreement with my dispatch brief — I accept it, and it leaves a decision open

I briefed that item 3's 10px size was part of the fix. The PR declined, arguing that `--text-base` is a real label-ramp token, that the caption is ornament type inside a drawn 96px disc (§4's #1147 carve-out for an immovable container), and that **10px and 12px both owe 4.5:1 anyway, so the size was never part of the contrast math.**

That last point is decisive and I was wrong to fold it in: this is a contrast issue, and the size question is a §4 type-scale question that does not change any threshold here. Raising it deliberately and openly rather than silently is the right way to disagree with a brief.

**Still an owner call**, though, because #1766's body explicitly said the size was in scope. Recorded on the issue; it is a one-line change if the answer is "raise it anyway".

Issue #1874, opened 2026-08-16
Invitation perks and two faction descriptions must state real backend mechanics


pixieofhugs:
The invitation letter's second perk bullet currently sells atmosphere while four factions have a real, enforced game mechanic that no player is ever told about. Two faction descriptions advertise things **every** faction gets. Owner ruling: the perk line states the faction's actual backend mechanic.

Found during a full audit of faction copy. `perks[1]` was originally ruled *generic* because five of seven used the identical *"…that turn a Tuesday into ___"* construction — filler, not a perk. That ruling is **reversed** here: once the slot states a real mechanic it has to vary per faction.

## Era 1 perks, as actually configured

| faction | mechanic | where |
|---|---|---|
| ua | `habit_bonus_points=5` — flat +5 on a praxis submitted within `habit_window_days` (7) of another of their own | `FactionConfig` |
| snide | `duel_win_modifier=2.0` / `duel_loss_modifier=0.0` — double on a win, nothing on a loss (baseline is 1.5 / 0.5) | `FactionConfig` |
| wow | `level_jump_reach=1` — once per level, claim one task a level above your own | `FactionConfig` |
| coven | `collab_own_modifier=1.1` — +10% on own-faction collabs | `FactionConfig` |
| everymen | `can_hold_multiple_memberships=True` — "Double Dipper" | `FactionConfig` |
| albescent | `can_always_rejoin=True`, plus a metatask apply-level bypass | `FactionConfig` + `services/praxis_metatask.py:45` |
| ephemerists | membership of `allow_praxis_on_retired_task_factions` — may create praxes on retired tasks | **`EraConfig`** |
| singularity | **none** | — |

Perks live in **two** places — six are `FactionConfig` fields, but Task Vision is an `EraConfig` frozenset. Any future audit must read both.

## New `{F}.invitation.perks[1]`

| faction | copy |
|---|---|
| ua | +5 points on any praxis submitted within seven days of your last. The practice rewards showing up. |
| snide | duels pay double when you win and nothing at all when you don't. no other crew takes that bet. |
| wow | The Leap of Whimsy — once each level, a knight may storm a single task one rung above their station and take it anyway. |
| coven | +10% points on any task you take on together with another witch ✦ |
| ephemerists | Task Vision — retired tasks stay walkable for you alone, long after the road has closed to everyone else. |
| everymen | Double Dipper — the only hands on the board that may take the same task twice over, while the first is still in progress. |
[...]

Issue #1871, opened 2026-08-16
Albescent should hold every other faction's perk but holds only two


pixieofhugs:
**Owner ruling:** Albescent is meant to hold every other faction's perk. It does not — it holds two of its own. That gap is a bug.

Surfaced during a full audit of faction copy, where each faction's invitation perk is being rewritten to state its real backend mechanic.

## What Albescent has today

| perk | where |
|---|---|
| `can_always_rejoin=True` — rejoinable after leaving, where leaving is otherwise era-final | `era_1.py:119` (`FactionConfig`) |
| Metatask apply-level bypass — may apply metatasks at any level; everyone else needs `metatask_apply_level` (5 in Era 1) | `services/praxis_metatask.py:45` — **hardcoded slug branch**, see below |

Its `FactionConfig` block otherwise carries baseline values throughout: `duel_win_modifier=1.5`, `duel_loss_modifier=0.5`, all four task/collab modifiers `1.0`, no `habit_bonus_points`, no `level_jump_reach`, no `can_hold_multiple_memberships`.

## What Albescent is missing

| faction | perk | where it's configured |
|---|---|---|
| ua | `habit_bonus_points=5` (+5 within `habit_window_days`=7) | `FactionConfig` |
| snide | `duel_win_modifier=2.0` / `duel_loss_modifier=0.0` | `FactionConfig` |
| wow | `level_jump_reach=1` | `FactionConfig` |
| coven | `collab_own_modifier=1.1` | `FactionConfig` |
| everymen | `can_hold_multiple_memberships=True` | `FactionConfig` |
| ephemerists | membership of `allow_praxis_on_retired_task_factions` | **`EraConfig`** |

Perks live in **two** places — six are `FactionConfig` fields, Task Vision is an `EraConfig` frozenset (`era_1.py:300`). Any fix must cover both, and any future audit must read both.

## Ruling 1 — Albescent takes Snide's duel deal WHOLE

`2.0 / 0.0`. Not `2.0 / 0.5`.

Owner's words: *"They have so many other advantages, they can deal with the full snide deal."* Albescent inherits the **ability**, downside included — double on a win, nothing on a loss.

**This is the one axis a naive derivation gets wrong.** Higher is better on both duel fields, so a plain "best value per axis" rule yields `max(0.5, 0.0) = 0.5` and quietly hands Albescent an upside-only version of the gamble that is strictly better than Snide's own.

[...]

Issue #1869, opened 2026-08-16
Singularity's Era 1 perk is hacking: the array reads the config in the console


pixieofhugs:
> *Rewritten 2026-08-18 from a /grilling session with the owner. Every heading below is a
> ruling. The original body — an open ask to "design and configure a perk" — is in the edit
> history; do not build from it.*

Singularity was the only joinable Era 1 faction with no mechanical perk. It gets one, and the
perk is **hacking**: Singularity reads the game's configuration in the browser console, on
surfaces that render it only as prose.

## Owner rulings

1. **Era 1, not Era 2.** The body warned that a perk changes balance mid-era. **This one does
   not** — it grants information, and alters no score, no modifier and no gate. That is what
   makes shipping it now safe, and it is the reason this shape was chosen over a modifier.
2. **The perk is "hacking."** Information the client already holds, surfaced in the devtools
   console, never on the page.
3. **Desktop-only is accepted.** The owner: *"I am okay with hacking requiring a computer."*
   There is no F12 on a phone and no in-app mirror is owed. Do not build a mobile equivalent and
   do not treat #494 as a blocker.
4. **It prints the CONFIG, not player data.**

## THE TWO STANDING RULES — read these before writing a line

> **1. The array may only print from responses the client has already received.
> Never a new endpoint, never a new field, never a new query.**
>
> **2. The array prints a CURATED ALLOWLIST of named values. Never a raw payload dump.**

Both must survive into the code as comments. Rule 1 makes *"can this leak personal
information?"* answerable by inspection: if a value is not already on the wire, the console
cannot print it.

**Rule 2 exists because rule 1 alone is not sufficient, and the failure is subtle.**
`GET /auth/me` carries `can_start_as_albescent` and `second_character_level_required` as **field
names**. A single `console.log(user)` therefore leaks the word *"Albescent"* to a player who has
never been revealed to it — **the secret escapes through the key, not the value**, and rule 1
permits it because the payload was already received. That would blow ADR-0027 in one line that
no reviewer would flag.

So the array enumerates the exact values it prints, by name, at its own call site. A future
field added to any of these payloads must not appear in the console until someone deliberately
adds it to the list.

Verified on `origin/main` @ `a7af24c3`: **no public outbound schema carries `email` or
`account_id`.** `email` appears only in `backend/schemas/admin.py` (admin-only surfaces) and
`backend/schemas/contact.py` (an *inbound* request body). `account_id` appears only in
[...]

Issue #1865, opened 2026-08-16
Spotlight labels promise a weekly/fortnightly rotation that does not exist (spotlight is all-time top score)


pixieofhugs:
Four of seven faction pages promise a rotating spotlight window that does not exist in the code.

## The mechanic

All seven faction bodies compute the spotlight the same way — **highest all-time score**, no time filter:

```
frontend/src/pages/factionDetail/archetypes/CovenFactionBody.tsx:126
  // ⑥ spotlight = highest all-time score; roster = the rest.
```

Same comment and same logic in `EphemeristsFactionBody.tsx:173`, `EverymenFactionBody.tsx:158`, `SingularityFactionBody.tsx:168`, `SnideFactionBody.tsx:179`, `UaFactionBody.tsx`, `WowFactionBody.tsx`. There is no weekly or fortnightly window anywhere in the backend either.

## The copy

| faction | label | claims a window? |
|---|---|---|
| snide | menace of the **week** | ❌ wrong |
| coven | Witch of the **week** | ❌ wrong |
| everymen | Standard-bearer of the **week** | ❌ wrong |
| wow | Champion of the **Fortnight** | ❌ wrong — *and disagrees with the other three on the cadence* |
| ua | Held up | ✅ correct |
| ephemerists | Keeper of the road | ✅ correct |
| singularity | PRIMARY NODE | ✅ correct |

A player who tops their faction sees "of the week" and reasonably expects to lose the slot in seven days. They never will, unless someone outscores them all-time.

## Owner ruling: fix the copy, not the mechanic

The spotlight stays **all-time top score**. The four wrong labels drop the false time period while keeping their faction's noun. The three correct ones are unchanged.

Suggested (wording is the owner's call, the requirement is only "no time period"):

| faction | from | to |
|---|---|---|
| snide | menace of the week | top menace |
| coven | Witch of the week | head witch |
| everymen | Standard-bearer of the week | Standard-bearer |
| wow | Champion of the Fortnight | Champion of the Court |

Copy lives in `frontend/src/locales/en/factions.json` under `<slug>.spotlight.label`.

## Note for the copy epic

`{F}.spotlight.label` is a deliberate exception in #1864: its surface (roster / spotlight) goes generic, but this slot **keeps faction voice**. That was ruled separately and is not a contradiction.

## If you'd rather have the feature

[...]

PR #1875, opened 2026-08-16
Record the faction copy review: 800 strings, every one ruled

pixieofhugs:
Docs only — no application code, no copy changes yet. This lands the **decision record** for a full faction-copy review so the seven implementation issues have a durable reference.

## What was reviewed

Every faction-specific English string in the app: **800 strings, 252 concepts, 9 factions**, drawn from the six i18n catalogs that carry per-faction keys plus 52 strings found hardcoded in `.tsx` (which became #1858).

Each string carries an owner ruling:

| decision | strings |
|---|---|
| stays faction-specific | 409 |
| collapse to one shared string | 246 |
| delete the slot | 133 |
| read `names.{F}` instead of duplicating it | 7 |
| not faction copy (extractor artifact) | 5 |

The governing principle, in the owner's words:

> *"We are erring on the side of taking out faction-specific flavour with the attitude 'we can put it back in intentionally'."*

Five domain words also settled to one form each — **level, points, task, praxis, submitted** — retiring `anno`/`rank`, `pvncta`/`huzzahs`/`cr`, `heist`/`quest`/`survey`/`job`/`function`, `spell`/`chronicle`/`transcription`, and `seal`/`file`. Applied by one rule: **replace where the word names the entity, keep it where it is imagery.**

## Files

| file | |
|---|---|
| `docs/copy/faction-copy-decisions.csv` | all 800 strings, with disposition, agreed wording and rationale |
| `docs/copy/faction-copy-voiced.csv` | the 409 that stay faction-specific — a filtered view |
| `docs/copy/README.md` | columns, the `{F}` convention, the word rulings, and a table linking every issue |

## Why this is a record and not a spec

The buildable specs are the issues, and each is self-contained — an agent does not need these CSVs to work from one. These files are for the whole picture, and for checking what was decided about a string the issues do not mention.

- #1863 — words only: the five nouns, 27 exact rewrites, 5 documented false positives, the `CONTEXT.md:96` vs `:551` contradiction
- #1864 — key structure only: 40 collapses, 97 deletions, 3 name references
- #1874 — invitation perks and two descriptions must state real backend mechanics
- #1858 — 52 strings hardcoded in `.tsx`; blocks #1864
- #1869 — Singularity has no mechanical perk
- #1871 — Albescent holds two perks, not seven; blocks its new copy
- #1865 — spotlight labels promise a rotation the code does not implement

## Bugs the audit turned up

Three cases where shipped copy promises something the code does not do:

- **Spotlight** — four factions advertise a weekly or fortnightly rotation, and disagree on which. `spotlight` is `highest all-time score` in all seven faction bodies; no window exists. (#1865)
- **Coven** — *"Full points on every task"* is true of every faction (`own/other_task_modifier=1.0` across the board). Its real +10% collab bonus goes unmentioned. (#1874)
- **Albescent** — *"Full points and any meta tasks from any group"*; both universal, since `faction_permits` states there is no active faction gate. (#1874)

Plus the invitation letter's fourth term row, cut entirely: it rendered static strings as if they were live standing (`InvitationLetterPopup.tsx:255` interpolates nothing), so `ua`'s *"#2 in the roster"* showed to every player regardless of rank — and the seven values disagreed on whether the row meant leaderboard rank, membership state, or level title.

## Two traps recorded in the README

Both cost real time during this audit and will cost it again on the next one:

1. **Faction perks live in two places.** Most are `FactionConfig` fields, but Ephemerists' Task Vision is an `EraConfig` frozenset (`allow_praxis_on_retired_task_factions`). Reading only the faction blocks makes a shipped, enforced perk look like fiction.
2. **Grep the domain name, not the copy's name.** The "Leap of Whimsy" is `level_jump` in code; searching the marketing phrase returns only the copy string itself.

## Risk

None to the running app. New directory, no existing file touched, nothing imported by the build.


pixieofhugs:
Heads-up from triage on #1865: the `{F}.spotlight.label` row in both `faction-copy-decisions.csv` and `faction-copy-voiced.csv` has been **superseded by a newer owner ruling** (2026-08-16).

The recorded note reads *"drop the false time period ... Keep faction voice."* The owner has since ruled that all **seven** labels read `Champion`, each faction keeping its own key.

Worth amending both rows before this merges, so the decision record does not land already out of date. Full spec on #1865.

PR #1922, opened 2026-08-16
A hidden praxis's media leaves the /media mount and waits in quarantine

### LINKED ISSUE #1593: Media on a hidden praxis still serves from the unauthenticated /media mount
Named in #1568's title but **deliberately left out of its scope** — the owner's rulings there covered erasure-vs-moderation and the orphan sweep, not the mount. PR #1592 shipped both and flagged this as still open.

> **Ruled 2026-08-16.** Quarantine on hide. Scope corrected: **`hidden` only, not `failed`** — see below.

## What is exposed

`backend/main.py:161` mounts every uploaded file unauthenticated:

```python
app.mount("/media", StaticFiles(directory=settings.MEDIA_ROOT, check_dir=False), name="media")
```

Paths are id-keyed in one of two shapes — avatars at `{character_id}/avatar/{uuid}/avatar.jpg`, praxis media at `{character_id}/{praxis_id}/{uuid}/…` (#1336). Since #1565 the uuid segment means a URL cannot be *derived*, but any URL that leaks — a shared link, a referer header, a cached page — is permanently and anonymously fetchable, and nothing expires it.

## The contradiction, in the codebase's own words

`models/praxis.py:46` describes the unscored statuses:

> *"`hidden` is off the site entirely; `failed` is an admin ruling that the work was not done, so it **keeps its banner and its place in the feed** but banks nothing."*

A hidden praxis is off the site. Its images are not. **That is the whole defect.**

## Scope correction — `failed` is NOT affected

This issue's title and body said *"a `hidden` or `failed` praxis still serves from `/media` even though the praxis itself is suppressed in the UI"*. **`failed` is not suppressed in the UI** — the comment above says it keeps its banner and its place in the feed. Quarantining a failed praxis's media would break a praxis that is still meant to be displayed.

**Quarantine on `hidden` only.** Verify the set against `models/praxis.py:44-56` rather than against `UNSCORED_MODERATION_STATUSES`, which is a *scoring* set (`{hidden, failed}`) and is the wrong predicate here. Two different questions that happen to share one status.

## Ruling — quarantine, do not authorize

When a praxis becomes `hidden`, move its files out of `MEDIA_ROOT` into a directory that is **not mounted**. The file survives for moderators; the URL 404s because `StaticFiles` cannot find it. Reverse the move when a praxis leaves `hidden`.

This resolves the tension #1568 left open rather than picking a side. PR #1592's ruling — **a ban must not shred evidence** — is about keeping the *file*. Whether that file is *publicly served* is a separate lever, and this pulls only the second one. Evidence preserved, publication withdrawn.

Rejected: **leaving it and documenting the posture.** Defensible on its own terms, but it leaves the app asserting a praxis is "off the site entirely" while serving its pictures. Rejected: **authorizing every media request**, which puts Python in front of every avatar on every page to fix a case affecting a handful of files.

## The quarantine directory must be a SIBLING of `MEDIA_ROOT`, not a subdirectory

This is the detail that decides whether the fix survives its first sweep.

`services/media_sweep.py` reconciles files under `MEDIA_ROOT` against rows and reports anything unclaimed as an orphan; `scripts/sweep_orphan_media.py` is the CLI over it. A quarantine directory **inside** `MEDIA_ROOT` holds files no row claims by that path — so the sweeper would classify every quarantined file as an orphan and delete exactly the evidence #1592 protected.

A sibling directory is outside the scanned tree, so the sweeper never sees it and needs no new concept. Confirm that empirically against `claimed_paths` and `classify_shape` rather than assuming, and check `FORBIDDEN_MEDIA_ROOTS` still behaves for the new path.

## Not urgent, and worth saying why

No credential, token or private document is served here. The failure mode is "an image you published stays fetchable", which is close to what publishing means. This is a posture question, not an incident — but the hidden case is a genuine contradiction and is worth closing.

## Where to look

- `backend/main.py:161` — the mount, and the comment block above it that should record the settled posture
- `backend/services/media.py` — `resolve_stored_media_path`, `delete_stored_avatar` (made public by PR #1592); they are the "is this a file we own" predicate a move also needs
- `backend/services/media_sweep.py`, `backend/scripts/sweep_orphan_media.py` — the reconciler
- `backend/models/praxis.py:37-56` — the statuses and why `failed` is different
- #1565 — the avatar path uniquification, and why the deterministic path was a problem

## Done when

- A praxis moved to `hidden` has its media 404 from `/media`, with the files still on disk outside the mount.
- Moving it back out of `hidden` restores the URLs.
- A `failed` praxis's media still serves — it is still on the site.
- The orphan sweeper does not report or delete quarantined files, demonstrated by running it (dry run) with files in quarantine.
- Avatars are untouched: they are public by design and belong to a character, not a praxis.
- `main.py`'s mount comment records the settled posture — unauthenticated, unguessable, permanent, with `hidden` praxis media withdrawn — so the next reader does not re-open this.

pixieofhugs:
`models/praxis.py` says `hidden` is "off the site entirely". That was true of the
HTML and false of the pictures: `/media` is an unauthenticated `StaticFiles`
mount, so a hidden praxis's uploads stayed permanently and anonymously fetchable
to anyone holding the URL.

Hiding a praxis now **moves its files off the mount** into a sibling directory
nothing serves. The URL 404s because Starlette cannot find the file; the bytes
survive for moderators; leaving `hidden` moves them back. One lever, not two —
PR #1592's ruling is about keeping the *file*, this one is about whether that
file is *published*.

Closes #1593

## The seam

`services.praxis.moderate_praxis` — the only writer that moves a praxis into or
out of `hidden` (grepped every `moderation_status` assignment; `flag_praxis`
only ever writes `flagged`, and creation only ever writes `visible`) — paired
with the `StaticFiles` mount in `main.py`. The integration tests drive the admin
route over HTTP and then `GET` the media URL, because "the praxis is hidden" and
"the URL 404s" were two separate facts before this change. The two withdrawal
tests fail on `main` (verified by disabling the hook and re-running: 2 failed,
4 passed).

## `failed` is NOT affected

The predicate is `moderation_status == hidden`, deliberately **not**
`UNSCORED_MODERATION_STATUSES`. That set answers a *scoring* question and also
holds `failed`, which keeps its banner and its place in the feed — quarantining
its media would break a praxis the site still displays.
`test_failed_praxis_media_is_untouched_and_still_serves` is the assertion that
distinguishes this change from the wrong one.

## The quarantine directory

`services.media.quarantine_root()` returns `realpath(MEDIA_ROOT) + "-quarantine"`.
Suffixing the realpath rather than joining a name makes the sibling relationship
structural: `<root>-quarantine` can never satisfy the `startswith(root + os.sep)`
containment test, which is the same test both scanners of the mounted tree use.
Inside `MEDIA_ROOT` it would be (a) still a URL and (b) swept — no row claims a
file at the quarantine path, so `sweep_orphan_media` would offer to delete
exactly the evidence a hide preserves. Confirmed empirically against
`claimed_paths` / `classify_shape` in
`test_quarantine_is_invisible_to_the_orphan_sweeper`, not assumed.
`FORBIDDEN_MEDIA_ROOTS` is untouched and still only ever consulted about
`MEDIA_ROOT`, which this change does not move.

No new setting, no new env var, nothing new stored. The mapping is
`MediaItem.file_path` mirrored under the other root, so no absolute path is
stored (root `CLAUDE.md`) and the reverse move cannot land elsewhere:
`file_path` is written once by `process_and_save_media` and never rewritten, so
a praxis or character that changed ids still restores to the path its own row
names.

## Safe to repeat, safe to fail

Both legs are best-effort and idempotent, the same posture as
`delete_stored_avatar`:

- **Source missing** — the file is already at the other end, or was never
  written. Not an error; returns "nothing moved". This is what makes hiding
  twice, or hiding a praxis whose file is gone, a no-op.
- **Target present** — a stale copy of the same upload. Removed first, because
  Windows `rename` refuses an existing destination.
- **`OSError`** — logged and swallowed. A filesystem problem must never block
  the moderation decision.
- `shutil.move`, not `os.replace`: the two roots can be different filesystems,
  where `rename` raises `EXDEV`.

The restore branch runs on **every** non-hidden moderation rather than only on
the transition out of `hidden`, so a half-completed withdrawal self-heals on the
next admin action.

## ⚠️ One thing you have to decide, and it is in a file I cannot edit

On Render, `MEDIA_ROOT=/app/media` **is the persistent disk's mount point**
(`render.yaml`, `disk.mountPath`). Its sibling `/app/media-quarantine` therefore
sits on the container filesystem and **would not survive a deploy** — evidence
withdrawn, then evaporated at the next merge to `main`.

The fix is one env value in `render.yaml`, no `mountPath` change and so no
stop-then-start:

```yaml
- key: MEDIA_ROOT
  value: /app/media/uploads      # was /app/media
```

That puts the quarantine sibling at `/app/media/uploads-quarantine` — on the
disk, persistent, still outside the swept tree. It needs a one-time move of the
existing `/app/media/<character_id>/…` tree into `uploads/` or old avatars 404.
`render.yaml` is outside my scope, so this PR does not touch it; the constraint
is written into `services/media.py`'s `QUARANTINE_SUFFIX` docblock so the next
reader hits it.

Locally and in CI (`MEDIA_ROOT=/tmp/media`) the sibling is on the same
filesystem and this is a non-issue.

## Known ceiling

Marked `ponytail:` in `withdraw_media_from_mount`: quarantine is deliberately
outside the orphan sweep's reach, so a praxis **deleted while hidden** leaves
its files there with no row pointing at them and nothing to collect them.
Upgrade path named in the comment — a `--include-quarantine` mode on
`scripts/sweep_orphan_media.py`.

## Checks

- `pytest --cov=. --cov-report=term-missing --cov-fail-under=80` from `backend/`
  against a dedicated `wz1593_test`: **1311 passed**, coverage 93.92%.
- `backend/scripts/dump_openapi.py` re-run: no diff. No route or schema shape
  changed, so the `api-schema` gate is unaffected. No frontend change. No
  migration — nothing schema-level moved.
- No browser or dev stack available to me: **visual QA is outstanding.**

## What to eyeball

1. **The manual 404.** Upload an image to a praxis, copy its `/media/…` URL and
   confirm it loads in a browser tab with no session. Hide the praxis from the
   admin moderation panel. Reload that exact URL — it must be a hard 404, not a
   cached 200. Force-reload; the mount sends no `Cache-Control`, so a browser
   may hold the old bytes.
2. Then check the file is still on disk under `<MEDIA_ROOT>-quarantine/` at the
   same relative path — evidence preserved, not deleted.
3. Un-hide it. The same URL serves the same bytes again, and the praxis detail
   page renders its image.
4. A `failed` praxis still shows its image in the feed and on its detail page —
   that is the scope correction, and it is the one that would be quietly wrong.
5. Hide a praxis that has **no** media at all, and hide one twice in a row. Both
   should just return 200.
6. Deployment: decide on the `MEDIA_ROOT` change above before this ships, or
   accept that on prod the quarantined bytes go away at the next deploy.


pixieofhugs:

**HELD — do not merge. Owner decision, 2026-08-16.**

The code is correct and CI is green, but merging it as-is would ship an **evidence-loss regression** that no test in this repo can catch.

## Why

`render.yaml` sets `MEDIA_ROOT=/app/media`, and that path **is** the persistent disk's `mountPath`. The quarantine directory is a sibling — `/app/media-quarantine` — which therefore lands on the **container filesystem**, not the disk, and is destroyed at the next deploy.

So in production this PR would withdraw a hidden praxis's media from public serving and then lose it. That directly contradicts PR #1592's ruling that **a ban must not shred evidence** — the ruling this issue was explicitly written to work alongside, pulling only the publication lever and not the retention one.

Locally and in CI `MEDIA_ROOT=/tmp/media`, so every test passes and will keep passing. The defect is invisible to the gates.

## What unblocks it

One env value in `render.yaml`, which is owner-only:

```yaml
MEDIA_ROOT: /app/media/uploads
```

That puts quarantine at `/app/media/uploads-quarantine` — on the disk, still outside the served tree. No `mountPath` change, so no stop-then-start deploy. It needs a **one-time move** of the existing `/app/media/<character_id>/…` tree into `uploads/`.

Credit where due: the agent that wrote this PR found the problem itself, wrote the constraint into `QUARANTINE_SUFFIX`'s docblock, and flagged it rather than shipping green.

## Status

Branch left intact. Merge once `MEDIA_ROOT` has moved and the existing tree has been relocated on the prod disk.

Issue #2195, opened 2026-08-18
One ornament per faction, and it alternates: a card on an ornamented ground goes plain


pixieofhugs:
Owner ruling, 2026-08-17, from a grill that started at "why does the Everymen
praxis card have ruled paper when the task card has a sunburst".

**Every ornament below is CONFIRMED by the owner** (2026-08-17), including the
three that were proposals in the first draft of this issue and the SNIDE
reversal. Nothing here is an agent's guess.

## The law

**Each faction has ONE ornament. A body on an ornamented ground goes plain; a
body on a plain ground wears its faction's ornament.** It alternates.

Applies to **task cards and praxis cards** (owner, explicitly). "Ornament" means
the card's BACKGROUND texture — never its frame, masthead, points mark or any
other chrome, which are untouched throughout.

Two supporting rulings from the same grill:

- **"Either burst, or plain."** No third texture. Where a faction's ornament is
  not being worn, the ground is bare — not a different pattern.
- **One drawing per faction, not one per surface.** Everymen currently draws
  seven different burst geometries; they collapse to one.

## The mechanism (lands FIRST, alone)

The seam already exists and needs no new machinery.

`<FactionBackdrop />` mounts **exactly once**, at
`components/Layout.tsx:61`. Exactly two places ever set its slug:

- `pages/CharacterProfile.tsx:106` — viewing a player of faction F
- `pages/factionDetail/useFactionDetail.ts:148` — F's own faction page

Everywhere else (`/tasks`, Home, FieldDesk, task detail, praxis detail) a card
sits on the plain global ground. So the predicate is dynamic — the same card is
on an ornamented ground on two routes and a plain one on every other — and it
cannot be answered by a static per-surface decision.

`useBackdropSlug()` is already exported from
`components/backdrop/BackdropContext.tsx:43`.

Add a sibling boolean — name it what fits, `groundIsBusy` in this brief — that
answers **"is the currently mounted backdrop ornamented?"** The predicate is
**any ornamented ground, not just the card's own faction** (owner ruled (a) over
the narrower same-faction reading): a Coven card on the Everymen wall is still
two textures fighting, and "sometimes it alternates" is a worse rule than "it
always does".

The declaration belongs to the **backdrop**, which is the only thing that knows
whether it is busy — not to nine cards each comparing slugs. Backdrops that are
a wash rather than a pattern (Coven's candlelight, UA) are NOT busy.

[...]

Issue #2146, opened 2026-08-18
Ephemerists CTA: blue in light, brass in dark, one control across three surfaces


pixieofhugs:
> ⛔ **BLOCKED** until #2141 has merged. It adds
> `--faction-ephemerists-plate-brass-rule` and the light half of the register,
> both of which this issue reads. Do NOT add those tokens here — that is how two
> PRs collide on the same rows and turn main red.

Parent: #2140. Depends on #2141.

Every Ephemerists surface painting a button through
`--faction-ephemerists-plate-cta-bg`: the task card, the task detail, the composer.
`EphemeristsRuneStrip.tsx` for the bracketing marks.

> ⚠ Staleness table in #2140. The rune strips already bracket the CTA on task
> detail and the composer (#2067) — that part is **done**, do not rebuild it.

## 1. The button is blue in light and brass in dark

| | ground | ink | border |
|---|---|---|---|
| light | `-plate-cta-bg` `#12151f` | `-plate-cta-ink` `#c9a24b` | `2px solid -plate-brass-rule` |
| dark | `-plate-cta-bg` `#c9a24b` | `-plate-cta-ink` `#12151f` | `1px solid -plate-brass` |

Both are set by #2141; this issue is about the surfaces consuming them
correctly and about not special-casing per surface.

**Yes, it inverts across themes, and that is the ruling.** The alternative —
keeping the blue in dark — was rejected on measurement: `#12151f` on the `#171a26`
plate is **1.10:1**, so the fill does nothing and a 2px rule ends up carrying the
primary action. With the rune strips sitting 11px above and below it, that outline
stops reading as an enclosure. The inversion is not a wobble in identity: it is the
same two colours trading places, which is what blued steel and a brass inlay
actually do depending on which is the ground.

## 2. One sign-up control, not three

The task card is the single definition of this button. The task detail and the
composer must strike theirs from it rather than each restating the treatment —
WORLD_ZERO_STYLE §6. If that means extracting the control, extract it.

## 3. The bracketing strips halve

`EphemeristsRuneStrip`'s sequence is 32 mathematical marks. **Halve it to 16** —
same sequence, every other mark dropped — to match the masthead's notation band,
which the owner halved in the same pass.

Keep the mathematical vocabulary. The design proposes replacing it with the seven
planetary metals plus eight alchemical element signs; **rejected**, because the
alchemical block is U+1F700 and is in no common system font, so it would buy a
[...]

Issue #2144, opened 2026-08-18
Ephemerists ground: the warped ephemeris net replaces the seven-layer wash


pixieofhugs:
> ⛔ **BLOCKED** until #2141 has merged. It adds
> `--faction-ephemerists-plate-brass-rule` and the light half of the register,
> both of which this issue reads. Do NOT add those tokens here — that is how two
> PRs collide on the same rows and turn main red.

Parent: #2140. Depends on #2141.

`frontend/src/components/backdrop/EphemeristsBackdrop.tsx`, the `.eph-backdrop`
recipe in `index.css` (~4332), and the task brief's ruling in
`pages/taskDetail/archetypes/EphemeristsTaskDetail.tsx:826`.

> ⚠ Staleness table in #2140.

## The page stops being aged paper and becomes a chart

`.eph-backdrop` today is a seven-layer wash: page colour, a gold corner bleed, a
**nile** corner bleed, two partial astrolabe rings, dot foxing, a straight 26px
ruling, and a linear gradient.

Replace all seven with **one layer**: the warped ephemeris net — meridians and
parallels bent around a point off to the right, with a fan of straight lines
converging on it. The source drawing is the design's `eph-backdrop.js` (a
`1000×600` SVG, `preserveAspectRatio="none"`, stroked `#c9a24b` at `0.7`).

Owner ruling: A says *this is aged paper*, B says *this is a chart* — and a chart is
a thing an ephemerist made, where foxing is just something that happened to the
paper.

Two of the seven layers reference `-plate-nile`, which #2141 deletes, so
this file has to change regardless. **Coordinate with that child — don't patch the
old recipe on the way past.**

## Opacity is not one number

| where | opacity |
|---|---|
| the page ground | **0.17** |
| a card or plate ground | **0.10** |
| the task brief | **0.17** |

The design uses `0.17` on cards and `0.42` on the brief. Owner walked both down:

- On a **card**, the net sits behind 12px body text rather than behind a solid
  fill, and at 0.17 its lines cross letterforms. 0.10 keeps it the same material at
  a weight that doesn't compete.
- On the **brief**, 0.42 puts the net's diagonals at roughly the same weight as a
  13px serif's strokes. This is body copy. Use the page's own 0.17.

## The brief's ruled notepaper

`EphemeristsTaskDetail.tsx:826` draws the brief on horizontal rules pitched to
`BRIEF_LEADING`. The design replaces them with the net. **Take the net at 0.17.**

[...]

PR #2414, opened 2026-08-20
The Ephemerists' gravity field lands on both cards, and alternates (#2398)

### LINKED ISSUE #2398: Ephemerists ornament: the gravity field lands on both cards, and alternates

Phase 3 of #2195, one faction per issue. **The mechanism landed in #2387** — `useGroundIsBusy()` in `components/backdrop/BackdropContext.tsx` answers "is the ground under this card busy?", and no card reads it yet. These five run in parallel: once phase 1 is in, no two of them share a file.

## The law

Each faction has ONE ornament. **A card on an ornamented ground goes plain; a card on a plain ground wears its faction's ornament.** "Ornament" means the card's BACKGROUND texture — never its frame, masthead, points mark, divider or any other chrome, which are untouched. **"Either burst, or plain"**: where the ornament is not worn the ground is bare, never a different texture.

## This faction

**The gravity field**, drawn in `components/factionMarks/ephemeristsPlate.tsx` (around `:362` on the 2026-08-17 survey — verify). **Worn on NEITHER card today; add it to both**, then alternate. The compass rose is the POINTS MARK, not the ornament — do not confuse them.

## Where the branch is true

Exactly ONE route: a **character profile** of a player whose faction paints a pattern (`pages/CharacterProfile.tsx`). Faction pages are **exempt** (owner, 2026-08-18) and already claim it via `useFactionBackdrop(slug, { cardsKeepOrnament: true })`. Every other route themes nothing and is plain by default, so the branch is a no-op there.

The predicate is **any** ornamented ground, not just the card's own faction (owner ruled (a) over the narrower same-faction reading): a Coven card on the Everymen wall is still two textures fighting, and "sometimes it alternates" is a worse rule than "it always does".

## Watch

- ⚠️ The inventory in #2195 was surveyed **2026-08-17** and several of these files have moved since (#2177 and #2185 both landed). **Verify every line number and file shape against `origin/main` before editing.**
- ⚠️ `main`'s critical CSS is at **WARN** with roughly 1.5 KB of headroom to FAIL. Measure with `node scripts/bundle-budget.mjs` — **gzip level 9**, the repo's own script, not a default-level measurement.
- Backdrops that are a wash rather than a pattern (Coven's candlelight, UA) are NOT busy — that is already encoded in `components/backdrop/ornamentedGrounds.ts`; read it rather than re-deciding it.

## Out of scope

Card chrome of any kind. Any backend change. Any other faction's files.

pixieofhugs:
Phase 3 of #2195, the Ephemerists: **the gravity field lands on both cards, and alternates.**

Closes #2398

## The seam I tested at

`the ground a page declared` → `what the two Ephemerists cards draw`, via `BackdropContext.Provider` under `renderToStaticMarkup` (effects never run in this harness, so a provider is how a ground is declared). New file: `frontend/src/components/factionMarks/__tests__/ephemeristsGravityAlternation.test.tsx` — 12 tests, red on the real defect first (8 of 12 failing before the change; the four chrome assertions passed throughout, which is what says the branch never reaches chrome).

It pins, on **both** cards: the field is drawn where nothing themed the ground; dropped on all five patterned grounds (`everymen`, `snide`, `singularity`, `ephemerists`, `wow` — *any* ornamented ground, per the owner's (a) ruling); kept over a wash (`coven`, `ua`, `albescent`, `na`); kept under the faction-page exemption (`cardsKeepOrnament`). And that the band, the cornice, the compass rose, the CTA's double rule and the rune strips are byte-identical on both sides of the branch.

## What changed

- **`factionMarks/ephemeristsPlate.tsx`** — one new mount, `CardGravityField`, beside the drawing it reuses. It reads `useGroundIsBusy()` and renders `GravityField` or nothing. **No second drawing was minted** (§6): the composer's field is the card's field, at the design's own geometry, differing only in the two things a card mount owes — `z-index: 1` (a ground: above the plate's fill, under the band's 2 and the cornice's 3) and a card weight of `opacity: 0.62`, for the same reason the chart net has three weights and takes its lowest on a card (#2144). One mount for both cards, so the two cannot drift and there is one place to invert if the law changes.
- **`taskCard/EphemeristsTaskCard.tsx`** — mounts it at the card's nominal width.
- **`praxisCard/desktop/EphemeristsPraxisCard.tsx`** — mounts it at `--praxis-card-basis`'s own default (394); the card states no width of its own (ADR-0067) and the field does not need a measured one, since its canvas anchors to the right edge and scales.

## ⚠️ The issue's premise was half wrong, and the correction is visible

The issue says the field is "worn on NEITHER card today". True for the praxis card. **The task card had a ground: the warped chart net at 0.10** (`EphemerisNet`, #2144's card weight, landed by #2146) — the 2026-08-17 survey missed it, and it is exactly the case the law's second clause names.

So this PR **takes the net off the task card ground** and gives the mount to the field. The reasoning, in the epic's own words:

- *"Either burst, or plain"* — where a faction's ornament is not worn the ground is **bare**, never a substitute pattern. Keeping both would put two textures on one card, which is the thing #2195 exists to end (it is the Everymen ruled-paper case, ruled).
- It would also have made the alternation a **no-op on the one route it exists for**: an Ephemerists card on an Ephemerists profile would drop the field and still be wearing the page's own net at 0.10 over the page's 0.17.

**The net is untouched everywhere it belongs** — `EphemeristsBackdrop` (0.17), `EphemeristsTaskDetail` (0.17) and `EphemeristsSelectCard` (0.10). Its three weights and their ruling stand; only the *card ground* mount moved. `ephemeristsCta.test.tsx`'s "lays the ephemeris net under the plate at the card weight" guard is rewritten in place to record the reversal rather than being deleted, so the next reader finds the argument, not a hole. **If the owner wants the net back on the task card, the revert is one line** — but then the Ephemerists alternation is decorative, and that is the call to make explicitly.

## Byte budget

Measured with the repo's own `node scripts/bundle-budget.mjs`, **gzip level 9**, before (at `main`) and after:

| blocking asset | before | after | delta |
|---|---|---|---|
| **critical CSS** | 22,993 B | **22,993 B** | **0 B** — same content hash, `index-89NHiAhz.css` both times |
| critical JS | 129,769 B | 129,797 B | +28 B |

**No CSS was touched at all** — the ornament is SVG in a lazily-loaded faction chunk, so none of `main`'s ~1.4 KB of headroom to the FAIL ceiling is spent. (The +28 B of critical JS is chunk-boundary noise; the Ephemerists modules stay in their own lazy chunk. The card's DOM actually gets *lighter*: the field is ~21 nodes where the net was 41.)

## Verification

`npm run lint` · `npm run typecheck` · `npm run typecheck:design-sync` · `npm test` (383 files, 6,852 passing) · `npm run build` · `npm run budget` — all green, on the rebased branch. No backend, no schema, no `.github/`, no other faction's files.

## What to eyeball — visual QA is OUTSTANDING

I have no browser and did not run the dev server, so nothing below has been *looked* at. In light **and** dark (the field's brass flips through the `[data-theme="dark"]` cascade):

1. **`/tasks`, an Ephemerists task card** — the field should read as bowed brass rows sagging toward the lower right, *not* as lined paper, and the description in italic Spectral must still read cleanly over it. This is the mount most likely to want a different `opacity` than my 0.62; that number is a judgement, not a ruling.
2. **The same card on a phone width (340)** — the well is anchored to the right edge; check it has not drifted into the CTA cartouche.
3. **An Ephemerists praxis card in the feed** — same field, first time this card has had a ground at all. Check the score stamp's crown, the masthead band and the byline rune strip all still sit *over* it (ground 1, band 2, cornice 3).
4. **A tall praxis card** (long body + a media well): the field stops at 1040px by design and the frame crops it — confirm no card in practice is tall enough to show the field end mid-sheet.
5. **A character profile of an EVERYMEN player** (`/players/<id>`) with Ephemerists cards on it — the one route where the branch is true. Both card kinds should be **bare**: no field, no net, chrome unchanged.
6. **A character profile of a COVEN or UA player** — a wash, so the cards should still wear the field. This is the case a per-route flag would have got wrong.
7. **`/factions/ephemerists`** — the exemption: busy page ground *and* busy cards, deliberately.
8. **The task card against the Ephemerists page ground** anywhere it appears — confirm the doubled net really is gone and the card reads as a distinct sheet on the chart.


pixieofhugs:

**Held as a draft at the owner's request — needs a look before it ships.**

The build is complete and all three CI jobs are green. It is not merged because it makes a visible change nobody ruled on.

## What the builder found

**#2398's premise was half wrong.** The issue says the gravity field is "worn on NEITHER card today — add it to both." The task card *did* already carry a ground: `EphemerisNet` at 0.10, mounted by #2144, which the epic's 2026-08-17 survey missed.

So the builder had to choose, and it took the net off the task card's ground and gave the mount to the gravity field. Its reasoning:

- Keeping both puts two textures on one card, against "one ornament per faction".
- Worse, keeping the net makes the alternation **a no-op on the one route it exists for** — on a patterned page the card would still be wearing a net over the page's net.

Only the *card ground* changed. The backdrop (0.17), task detail (0.17) and the faction select card (0.10) are untouched.

## The two things to look at

1. **The swap itself** — a chart net becomes a gravity field on the Ephemerists task card. This reverses a shipped #2144 mount. One line to revert.
2. **The opacity, 0.62** (peak row ≈ 0.15, against the net's 0.10). The builder flagged this itself as *"my judgement, not a ruling — the single most likely thing the owner will want moved."*

## Status

Draft, rebased, green. Nothing else in the batch depends on it; the other five ornament sub-issues merged without it. Ready to mark for review the moment there's a verdict on the swap and the opacity.

Issue #2453, opened 2026-08-22
Both lifecycle specs fail at their heads: collaboration times out at :248, duel fails a visibility assert at :60, and the pair suppress eight tests


pixieofhugs:
> *Filed during a backlog readiness audit with Molly (2026-08-22). Split out of #1674's CI-audit comment, item 3 of its suggested split.*

**The two lifecycle specs fail at their heads, and those two failures are what suppress the eight tests that have never run.**

Filed together rather than separately because both are `mode: 'serial'` lifecycle specs, both were "fixed" by #1676 and came back, and they may share a cause. Split them into two issues only if the investigation proves the causes are unrelated.

## The two failures, as measured

**`collaboration.spec.ts:248`** — *create via ModePicker, invite, accept, cast, and publish*

30-second test timeout, then at the `finally` block's `alice.ctx.close()`:

```
browserContext.close: Protocol error ... Failed to find context with id
```

**The close error is a symptom, not the cause.** The context is already torn down by the time cleanup runs — chasing the protocol error leads away from the bug. The timeout is what to diagnose: which step of create → invite → accept → cast → publish never completes.

**`duel.spec.ts:60`** — *challenge → accept → both seal → the settled rail shows both sides*

`expect(locator).toBeVisible()` failed. Which locator, and whether the rail is absent or merely different, is the first thing to establish.

## Why these two matter more than two tests

Both specs open with `test.describe.configure({ mode: 'serial' })`. In serial mode the first failure skips every test after it — so these two failures are what produce the nightly's **"8 did not run"**.

Those eight are the UI-driven specs from epic #953, the ones that click real buttons rather than driving the API, which is the entire reason that epic existed (see the `Split point (#953)` header in `duel.helpers.ts`). **They are the most valuable tests in the suite and the least exercised.**

Triaging them is **#1795** and is explicitly downstream of this issue. Do not attempt it here — they cannot run until these two heads pass.

## They were fixed once already

**#1676 fixed both and closed 2026-08-16.** #1674's 2026-08-18 comment records them red again. So, as with #2452, the question to answer before patching is whether #1676's fix was undone or whether this is a new cause. A second point-fix that regresses a third time is the outcome to avoid — and this is the third time these specs have been declared fixed.

#1674's own audit puts it plainly: three sub-issues here *"were closed as COMPLETED, and the symptom never went away."*

## Not this issue

- **The eight suppressed tests** — #1795, and only after this lands.
[...]

Issue #2450, opened 2026-08-22
36 of the nightly's 39 failures are one S.N.I.D.E. note: text measured against the paper it declares, not the ground it composites to


pixieofhugs:
> *Filed during a backlog readiness audit with Molly (2026-08-22). Evidence is the CI-audit comment on #1674 — every number below is measured from run logs, not inferred.*

**36 of the nightly's 39 failures are one S.N.I.D.E. note surface, and the fix is a design call: do the washes come down, or do the inks darken?**

## The finding

`contrast.spec.ts` fails 36 times — 9 factions × 2 themes × 2 viewports. That reads like nine faction regressions. It is not. Across all 36 there are exactly **three distinct colour pairings**, repeating unchanged for every faction:

```
88×  3.89:1 (needs 4.5:1) — rgb(84, 81, 67)    on rgb(188, 181, 155)
72×  3.85:1 (needs 4.5:1) — rgb(157, 23, 77)   on rgb(188, 181, 155)
72×  4.22:1 (needs 4.5:1) — rgb(255, 105, 172) on rgb(71, 56, 41)
```

The grounds are **faction-invariant** — `rgb(188,181,155)` in light, `rgb(71,56,41)` in dark, identical whether the run is exercising `ua`, `coven`, `singularity` or `na`. One surface on the page for every combination, not nine defects.

All three inks are S.N.I.D.E. tokens in `frontend/src/index.css`:

| ink | token | line |
|---|---|---|
| `#545143` | `--faction-snide-note-muted` | 1140 |
| `#9d174d` | `--faction-snide-note-pink-ink` (light) | 1154 |
| `#ff69ac` | `--faction-snide-note-pink-ink` (dark) | 3115 |

## Root cause: the declared ground is not the painted ground

`--faction-snide-note-paper` is **`#f4f1e8`** (line 1135). The browser paints **`rgb(188,181,155)` = `#bcb59b`**, markedly darker.

| ink | on declared paper `#f4f1e8` | on the painted ground `#bcb59b` |
|---|---|---|
| `#545143` | **7.06:1** ✅ | **3.88:1** ❌ |
| `#9d174d` | **6.98:1** ✅ | **3.84:1** ❌ |

3.88 and 3.84 reproduce the suite's 3.89 and 3.85 to rounding, so the arithmetic and the runner agree.

[...]

Issue #1861, opened 2026-08-16
Build the onboarding arc: scanned QR code to level 1


pixieofhugs:
Build the onboarding arc: the path from scanning a QR code in the real world to standing at level 1.

**The spec is the source of truth: `docs/spec/SPEC-onboarding.md`** (PR #1860). Read it before starting. It was charted decision-by-decision on the wayfinder map #1732, and every ruling links to the ticket holding its argument — zoom into those if a decision looks arbitrary.

## The shape

    scan → intro → auth → terms → [ hand off ] → character creation → the task → level 1

**Three net-new cards, then the flow ends.** Everything after the hand-off is existing app. Do not rebuild `CreateCharacter`, the task page, or the level-up Field Stamp.

## Two rules that will bite if missed

**1. All copy is a PLACEHOLDER. The owner writes every string.**
Use obviously-provisional text and route it through the i18n catalogs like everything else (`frontend/src/locales/en/`). **Do not lift the wording from `prototype/onboarding-flow`** — that branch is a throwaway built to be reacted to, not a copy source. Do not invent polished final copy either; placeholder means placeholder.

**One exception, and it matters: the terms card must render the REAL existing Disclaimer text** (`common.json` → `disclaimer.*`), versioned as v1. `main` auto-deploys, so placeholder legal text would ship to production. The revised wording is a separate owner task; until it lands, the card shows the copy that already exists and is already true.

**2. The cards wear the `na` / Unaffiliated kit — the spectrum sheet.**
Not generic paper. Read `components/taskCard/DefaultTaskCard.tsx` for the language. Rainbow appears in **exactly three places per card** — the restraint is the design, not a suggestion. Lora italic titles, Courier Prime elsewhere, every colour through a `--faction-default-*` token in both themes, no hardcoded hex. A `Default*` that invents its own metaphor is *stale* by the glossary's definition.

## Natural slices, by file footprint

Probably too large for one agent session. These are disjoint enough to run in parallel:

- **A · Backend — consent log.** Append-only table, one row per acceptance: account, version, timestamp. Plus the endpoint the terms card posts to. Existing accounts have no rows; that is the honest record.
[...]

Issue #1795, opened 2026-08-15
Nine e2e tests have never executed: triage them once the serial blocks stop failing early


pixieofhugs:
Nine e2e tests have **never executed**. Not skipped, not failing — never reached. Their state is unknown, and it cannot be established until #1676 lands.

## Why they never run

Both lifecycle specs open with `test.describe.configure({ mode: 'serial' })`. In serial mode the first failure in a block skips every test after it. `collaboration.spec.ts` fails at its second test and `duel.spec.ts` at its first, so everything downstream is dropped.

Run [31867605613](https://github.com/pixieofhugs/WorldZeroPlayground/actions/runs/31867605613) reports it plainly:

```
19 failed
 1 skipped
 9 did not run
20 passed (6.4m)
```

The nine, by block:

**`collaboration.spec.ts` — `collaboration lifecycle`** (after the failure at `:128`)
- `:144` the invitee ALSO sees the shared draft in their active-tasks sidebar
- `:162` the published praxis page credits both collaborators — carries a `KNOWN GAP (bug)` comment in-file and is *expected* to fail until a praxis-detail archetype iterates `praxis.members`

**`collaboration.spec.ts` — `collaboration UI (clicked buttons)`**
- `:227` create via ModePicker, invite, accept, cast, and publish
- `:288` C1: a member can kick a co-author from the roster pill
- `:307` C2: a member can leave the collab from a standalone control
- `:322` C3: a holdout on a pending collab is not shown the unsubmit control
- `:342` C4: the oldest of six pending invites is still reachable

**`duel.spec.ts` — `duel lifecycle`** (after the failure at `:60`)
- `:106` D1: an active duel can be neutrally dissolved by a participant
- `:133` D2: the opponent can withdraw a pending challenge

## Why this is a gate, not a spec

These are the UI-driven specs from epic #953 — the ones that click real buttons rather than driving the API, which is the entire reason that epic existed (see the `Split point (#953)` header in `duel.helpers.ts`). They are the most valuable tests in the suite and the least exercised.

At least one (`:162`) is a known-red baseline by design. The rest are unknown. **Do not assume they pass** because they were green when written — the era config, the seed and the sidebar have all moved since, and the seeded board is now a single cross-faction task where those specs were written against a richer fixture.

## What to do

1. Land #1676.
2. Dispatch the nightly: `gh workflow run e2e.yml --ref main`.
[...]

PR #2466, opened 2026-08-22
The guest smoke test stops chasing marketing copy (#2452)

### LINKED ISSUE #2452: guest.spec.ts is red: the marketing home smoke test was fixed by #1676 and re-broke, and nothing noticed
> *Filed during a backlog readiness audit with Molly (2026-08-22). Split out of #1674's CI-audit comment, item 2 of its suggested split.*

**`guest.spec.ts:6` — "marketing home renders for a guest" — is red. It is the most basic smoke test in the suite.**

A logged-out visitor loading `/` is the shop window, the QR-scan landing (`docs/spec/SPEC-onboarding.md`) and the single most-hit path on the site. Nothing else in the suite is cheaper to fix or worse to have broken.

## Why this is its own issue

It is filed alone deliberately. #1674's audit split the nightly's 39 failures into four causes and this is the only one that is a single test with a single assertion — the S.N.I.D.E. contrast family is #2450, and `collaboration` + `duel` are lifecycle specs that may share a cause and are filed together.

Do **not** batch this with those. It is the smoke test: if it is red, the value of triaging anything else is lower, because none of the other results can be trusted to be about what they claim.

## It has been fixed before, and it came back

**#1676 fixed `guest` and closed 2026-08-16.** #1674's 2026-08-18 comment records it re-breaking: *"the functional specs are red again (`guest` re-broke after #1676 fixed it)"*.

So this is a **regression, not a stale premise**, and that changes the job. #1676's diagnosis was a set of moved gates and changed fixtures. Whatever broke it a second time is either a different cause or a fix that did not hold. **Find out which before patching** — a second point-fix that regresses a third time is the outcome to avoid.

Worth reading #1676's patch first to see what it did, then checking whether that change is still present in `frontend/e2e/`.

## The standing conditions

- **The suite has never been green.** Zero successes in the last 100 runs. There is no green commit to bisect to; fix forward.
- **It is deterministic, not flake.** Runs `32455059644` (08-21) and `32557107772` (08-22) are byte-identical in shape: 39 failed · 8 did not run · 11 passed.
- **Nothing in CI typechecks `frontend/e2e/`** beyond what #1780's `tsconfig.e2e.json` now covers, and the browser run is nightly-only. A change here is not checked at PR time.
- Dispatch a run to verify rather than predicting from a patch: `gh workflow run e2e.yml --ref main`.

## Done when

`guest.spec.ts` passes in a dispatched run — an actual green, not an inferred one — and the issue says which of the two causes it was: #1676's fix having been undone, or a new one.

pixieofhugs:
Closes #2452

## Which of the two causes it was: **a new one.**

#1676's fix is still present, byte for byte. It only ever touched the *second* test in `guest.spec.ts` (`the dev-login button logs a bot in`) — `git show 6023a4b6 -- frontend/e2e/guest.spec.ts` is a 13-line diff entirely inside that test, and `guest.spec.ts` has had no commit since. That test **passes** on the nightly: run `32557107772`'s failure list names exactly one guest test, `guest.spec.ts:6:1`.

What broke test 1 is a copy change that landed *after* #1676:

| when | what |
|---|---|
| — | the spec has waited for a button named `/sign up here/i` since #389, matching `home.json` `hero.cta.loggedOut` = **"Sign Up Here"** |
| 2026-08-15 19:11 | **#1676** (`6023a4b6`) merges — fixes test 2 only |
| 2026-08-16 | **#1862** (`de8762a2`) **deletes** `hero.cta.loggedOut`: the hero became the two provider buttons, and its own commit message says *"`hero.cta.loggedOut` ("Sign Up Here") is deleted — the buttons say it"* |
| 2026-08-16 | **#1899** (`c2a09388`) re-adds the key as a single CTA into the onboarding arc, with the value `"PLACEHOLDER — the way in, for a stranger"` |

So the string the test waits for exists nowhere in `frontend/src` today. The nightly log agrees: `Locator: getByRole('button', { name: /sign up here/i })` … `element(s) not found`.

## The seam

**The guest branch of `RootLanding` — that `/` renders `Home`, not `FieldDesk`, and renders its *logged-out* variant.** Not "the hero says a particular sentence."

That distinction is the whole fix. The value in that slot is still literally the word `PLACEHOLDER`; it is waiting on final wording, so re-anchoring on whatever it says today is a third regression already scheduled. The test now anchors on `data-testid="home-hero-cta"`, which:

- lives on `Home.tsx` and nowhere else in the app (`FieldDesk` has no such node), so its presence *is* "the marketing Home rendered"; and
- is already pinned at PR time by `src/pages/__tests__/homeLoginNotice.test.tsx`, so it is a maintained seam rather than a fresh hook — nothing new was added to `src/`.

A bare test id would drop something the old locator carried for free: "Sign Up Here" only ever rendered **logged out**, whereas one CTA with that test id renders either way. Clicking it recovers that as behaviour instead of wording — signed out the CTA enters the arc at `/start` (#1861, unprotected route), signed in it goes to `/tasks`. A route is a contract; copy in a slot marked PLACEHOLDER is not.

The same rotted string appears once more in this file, as test 2's `toHaveCount(0)` — which had been **passing vacuously** against copy that exists nowhere. Same test id, same intent, and now it can actually fail.

## Not fixed here, and not shared with the lifecycle specs

The cause is specific to this file: one locator against deleted copy. It is **not** shared with `collaboration.spec.ts` / `duel.spec.ts` (#2453), and nothing here touches #2450 (S.N.I.D.E. contrast), #1795 or #1780. Scope is one file, +22/−2, no `src/` change.

## Verification

- Dispatched E2E run on this branch (`gh workflow run e2e.yml --ref claude/2452-guest-smoke-copy-anchor`) — result recorded in a comment below, not inferred from the patch.
- `npm run typecheck` — clean (this is `tsc --noEmit && tsc -p tsconfig.e2e.json --noEmit`, so the e2e suite is typechecked per #1780/PR #2306).
- `npm run lint` (covers `e2e`), `npm run typecheck:design-sync`, `npm test` (392 files / 7001 passed), `npm run build` — all clean.
- `npm run budget`: CSS 22.3 KB warn is pre-existing on `main` and untouched by an e2e-only diff.
- No browser/dev stack available to this agent, so **visual QA is outstanding** — the check that matters is the dispatched run.

## What to eyeball

- **The marketing home at `/`, logged out.** One hero CTA under the wordmark; it currently reads `PLACEHOLDER — the way in, for a stranger`, and clicking it must land on `/start`. That copy is the thing this PR deliberately stops asserting — when it is finally written, this test should not need touching.
- `/` while logged in, for contrast: FieldDesk, no hero CTA at all.


pixieofhugs:
## Dispatched run: `guest.spec.ts` is green

Run [`32601099164`](https://github.com/pixieofhugs/WorldZeroPlayground/actions/runs/32601099164), `workflow_dispatch` on this branch, checked out at `ca04ed1a` (the commit in this PR — the log shows the checkout ref).

| | nightly `32557107772` (main) | this run (branch) |
|---|---|---|
| failed | **39** | **38** |
| did not run | 8 | 8 |
| passed | **11** | **12** |
| `guest.spec.ts` in the failed list | `guest.spec.ts:6:1 › marketing home renders for a guest` | *(absent)* |

`gh run view 32601099164 --log-failed \| grep -c "guest.spec"` returns **0**. Playwright's `html` reporter does not enumerate passes, so the positive proof is the pair: the one guest entry left the failed list, and the passed count went up by exactly one, with `did not run` unchanged.

The run is still red overall, as expected and as scoped: 38 = the 36 S.N.I.D.E. contrast failures (#2450) + `collaboration.spec.ts` and `duel.spec.ts` (#2453). Neither shares a cause with this one — this was a locator against copy that had been deleted from `home.json`, and it is confined to this file.

Issue #1812, opened 2026-08-15
Collab submission status speaks one vocabulary: delete the eight faction voice blocks


pixieofhugs:
> *This was generated by AI during /builder-bot.*

**Owner ruling, 2026-08-15.** Collab submission status must read the same on every faction. No faction voice — no "still walking" for *not submitted*.

Observed on the live roster (Coven): `COLLABORATORS · 2` — `1 of 2 have cast` — `Pixie · you [still walking]` — `Nemesis [cast]`.

## Why

Workflow state is a **mechanical fact a player must read correctly in order to act**. Nine vocabularies for one mechanic means a player who misreads it loses work or publishes early. Faction identity belongs in the caret colour, the card archetype and the chrome — not in the name of a state you have to act on.

This is a deliberate exception to ADR-0065's per-faction dressing, and it **extends a decision already made**: #1154 rewrote the shared fallback block out of its witchy register for the same kind of reason — it "read as a voice nobody owned".

## This is a deletion, and it needs no new copy

`frontend/src/components/collab/collabCopy.ts` already resolves per-faction with a fallback: it probes `forms:editPraxis.<slug>.collab.<key>` via `i18n.exists()` and falls back to the shared `forms:editPraxis.collab.<key>` block — **which is deliberately the tier with no voice and already speaks the domain's own words.**

Verified on `origin/main`, the shared block already reads:

| key | shared (voiceless) value |
|---|---|
| `pillCast` | `submitted` |
| `pillWeaving` | `not submitted` |
| `castStatus` | `{{cast}} of {{total}} submitted` |
| `progressAria` | `{{cast}} of {{total}} members have submitted their part` |
| `castAction` | `Submit my part` |
| `castFinalAction` | `Submit — this one publishes it` |
| `pullBackAction` | `Pull my part back` |
| `bannerWaiting` | `Submitted. Waiting on the others.` |
| `bannerHoldout` | `Waiting on you — submit your part.` |
| `bannerPublished` | `All parts submitted. Every member is credited.` |
| `successHeading` | `The praxis is submitted` |
| `successBody` | `Every part is in — your collab is public and open to voting.` |
| `successContinue` | `View the praxis` |
| `you` | `you` |

[...]

Issue #1811, opened 2026-08-15
The composer grows three buttons and stops lying about what they do (frontend)


pixieofhugs:
> *This was generated by AI during /builder-bot.*

Frontend half of the collab submit redesign. Design and rejected alternatives: **#1809** (the ADR). Backend contract: **#1810**.

**Blocked by #1809 and #1810** — the state machine must exist before a surface can drive it.

## What changes

One Submit button becomes three affordances, each meaning one thing:

- **Done** — "my part is finished". Social. Reversible. No warning, no consequence.
- **Propose** — "I think we're ready". Starts the publish window. Consequential: say so.
- **Approve** — "I'm happy with this text". A vote on the live proposal.

Plus **Withdraw proposal**, for a member who has read the draft and has no edit to make.

## Confirm on the first keystroke after a proposal goes live

**Owner-specified.** Editing cancels the proposal, so the first keystroke after one is live must ask — once — rather than silently ending everyone's countdown.

- **Once per proposal**, not per keystroke. It fires on the first edit and then gets out of the way.
- The player must never be *blocked* from writing. Declining the confirm means the keystroke does not land; it must not leave the editor in a half-state.
- Copy names the real consequence: the publish window ends and everyone's approval clears.
- The editor is a CodeMirror instance bound to the room (`archetypes/controls.tsx`, #1742). The `EditorView.editable` compartment is the existing seam.

## Rewrite the copy — do not patch it

`frontend/src/locales/en/forms.json`'s `awaiting*` group describes a flow that no longer exists, and part of it is **wrong on production right now**:

- `awaitingEditConfirm` — "**Only your own submission is re-opened** — but the moment you save a change…" Both halves have been false since PR #1803: reopening clears every member's submission, immediately.
- `awaitingEditTitle` ("Take your part back out?") and `awaitingEditAction` read per-member too.

Do not fix these in place. The flow they describe is being deleted; write the group against the new one.

`frontend/src/pages/editPraxis/waiting/PraxisWaitingSurface.tsx` (ADR-0059) is the surface that explains a live proposal and offers re-entry. Copy lives in the i18n catalogs, never inline.

## ⚠️ Submission-status copy is shared, NOT faction-skinned

**Owner ruling, and a deliberate exception to ADR-0065's per-faction dressing.** The words for Done / Propose / Approve and every state they produce are **one literal vocabulary across all nine factions.** No faction voice — no "still walking" for *hasn't approved*.

[...]

PR #2472, opened 2026-08-22
Both lifecycle e2e heads had one cause: #1417 moved the invite box and the challenge badge (#2453)

### LINKED ISSUE #2453: Both lifecycle specs fail at their heads: collaboration times out at :248, duel fails a visibility assert at :60, and the pair suppress eight tests

pixieofhugs:
Closes #2453

## The seam

**The locator layer between the two lifecycle specs and the composer** — what
`frontend/e2e/` reaches for, whether the thing it reaches for exists, and whether
what it then asserts can tell success from silence. No app behaviour is changed
by this PR: every element it touches gains an attribute and nothing else.

The failing test was already written and already red — the nightly. The loop was
run at that seam, on the branch, dispatched, not predicted.

## Shared cause or coincidence: **SHARED**

Both heads were killed by one commit, `f1e4eb18` — *"The + invite chip and the
compose-stage duel pair (#1417)"* (#1509, 2026-08-01). It reworked exactly two
things in `archetypes/controls.tsx`, and each one broke one spec:

| #1417 changed | broke |
|---|---|
| the collab invite search moved **behind a `+ invite` disclosure chip** | `collaboration.spec.ts` filled a textbox that was not rendered → 30 s hang |
| `statusChallenged` **"challenged" → "Challenge sent"** | `duel.helpers.ts:127` `getByText('challenged')` → element not found |

So they are one bug wearing two symptoms, and #2453 was right to file them
together. **Not split.**

And a dispatched run then found a **second** shared cause behind the first:
`useEditPraxis.publish()` refuses an untitled praxis client-side, and neither
flow mints a title — `handleSignup` posts a task id and a type and nothing else,
`accept_duel` mints the opponent's praxis with none at all. It refuses
*silently*, because `publish()` dismisses the seal sheet and the propose confirm
on its first line, so both specs' post-conditions were green on a submit that
never left the browser. Full account in the run-proof comment below.

## Was #1676's fix undone?

**No — it is intact, and it was never enough.** All three of its changes are
still on `main`: `seed.py::ensure_duel_fixture_task`, the `'Duel'` mode chip
locator, and the `'Submit'` seal locator. #1676 was reasoned from source rather
than from a dispatched run, so it fixed the locators it could see and stopped —
and the very next line, `getByText('challenged')`, had already been dead for two
weeks. It moved the failure one step later, exactly as its own commit message
warned about the seed fix.

Same finding as its sibling #2452: a new cause, not a regression of the old one.

## What the collaboration timeout actually was

The `browserContext.close: Protocol error` is a symptom, as #2453 says. So is
the `finally` block it is reported at. With no `actionTimeout` configured, a
Playwright action on a locator that matches nothing waits until the **test**
timeout, so the 30 s is an action that never resolved and the reported line is
merely where the error surfaced.

The head has **five** dead anchors, and it hung on the first of them:

1. `/PULL THIS JOB/i` — in no catalog anywhere in this repo.
2. `/THE GANG/i` — likewise. Both were "Snide's voice" for a task
   `pickOpenTask` never guaranteed to be Snide.
3. the invite textbox — behind the `+ invite` chip since #1417.
4. `/Submit my part|this one publishes it/i` — ADR-0079 (#1811) replaced
   cast/cast-final with **propose → approve** and deleted `castAction`.
5. `/out there/i` — #1812 rewrote the collab register out from under it.

And the flow itself was stale, not just the words: since ADR-0079 the first
keystroke *after* a proposal goes live **cancels the proposal**, so the old
"edit, cast, edit, cast" order can no longer terminate. Both members now write
first, then Alice proposes and Bob approves.

## What the duel visibility assert actually was

`duel.spec.ts:60` is the test declaration; the assert is `duel.helpers.ts:127`,
`expect(getByText('challenged')).toBeVisible()`. **The rail is not absent** —
the line before it, `getByText(opponentName)`, passes, so the pair renders with
the opponent in it. Only the badge's wording moved. Its downstream neighbour was
dead too: #1928 renamed the seal dialog `"Seal the duel?" → "Lock the duel?"`
and its confirm `"Seal it" → "Lock it"`, the day *after* #1676 landed.

## The fix shape

Slots, per #2452's precedent — `data-testid`, plus a `data-*` attribute where
one shared slot means two different acts:

| slot | why |
|---|---|
| `composer-duel-pair` | mounts only once `praxis.duel_id` is set — the fact the spec was actually asserting |
| `composer-invite-open` / `composer-invite-search` | the disclosure and the box behind it |
| `composer-primary` | one slot, three labels by state; on **both** `PublishButton` and `CollabSignals`, exactly one of which is on screen |
| `data-collab-signal="propose\|approve"` | Propose and Approve are one button told apart only by wording |
| `duel-seal-sheet` / `duel-seal-confirm` | the sheet whose heading moved in #1928 |
| `confirm-dialog` / `data-confirm-kind` / `confirm-accept` | which confirm opened, without reading its title |
| `collab-success` | the closing beat |
| `task-signup-cta` (x8 archetypes) | all eight already read one shared label since #1497; the spec quoted one faction's voice for a task whose faction was never pinned |

The one label kept is the `Collab` mode chip, because every archetype builds its
chips from the same shared key — the symmetric twin of the `Duel` chip
`duel.helpers.ts` already presses.

## The check left behind

`frontend/src/__tests__/e2eAnchors.test.ts` — **this is the part that stops a
fourth regression.** It reads the slot ids and `data-*` assertions out of
`frontend/e2e/` itself and fails the *PR-blocking* suite if the app no longer
carries one, so a rename is caught by the reviewer's own CI instead of by a
nightly nobody reads. It also pins the two surviving mode-chip labels against
the catalog. Verified red by removing a `data-testid` and re-running (17 cases).

A hand-written list would rot in step with the thing it guards, so it derives.
The derivation's own liveness is asserted first, for the same reason this whole
issue exists.

## Deliberately not fixed

**`C3`'s `toHaveCount(0)` is vacuous** — `{ name: 'unsubmit', exact: true }`
against a control that reads `"unsubmit to edit"` — i.e. the same trap #2452
found. It is left as it is, with the reason written at the line: C3 had never
once executed when this branch was cut, so tightening it would have been
predicting its result rather than reading it. It now executes and reports green;
what that green is worth is #1795's to rule on.

**`C4` is the next head, and it is a fixture gap, not a UI one.** It wants six
distinct seeded tasks and the e2e database holds two — the same class as #1676's
missing duel task, from the same root (#1398). Closing it means a
`seed_dev_demo` helper in `backend/seed.py`, outside this PR's file scope, so it
is reported rather than fixed. It is also not muted: turning a missing fixture
into a `test.skip` would unblock P1/P2 by hiding the reason, which is the move
that produced three "fixed" e2e regressions already.

## Verification

- `tsc --noEmit` — clean. `tsc -p tsconfig.e2e.json --noEmit` — clean.
- `tsc -p tsconfig.design-sync.json --noEmit` — clean.
- `eslint src .ds-kit e2e` — clean.
- `vitest run` — 394 files, 7032 passed, 1 skipped.
- `npm run build` — ok. `npm run budget` — JS ok; **CSS warn at 22.3 KB vs 22.2**, pre-existing and untouched (this branch changes no CSS).
- **Dispatched `e2e.yml` run on this branch — `32605378628`.** Baseline
  38 failed / 8 did not run / 12 passed → **37 failed / 2 did not run / 19 passed.**
  Both heads pass (neither spec appears in `--log-failed`); six of the eight
  suppressed tests now execute and five of them pass. Full table, and the next
  head it exposes, in the comment below.
- No new API endpoints consumed. `e2e.yml` untouched: nightly + `workflow_dispatch`
  stays as its header records.
- No browser tools in this environment, so visual QA is outstanding.

## What to eyeball

1. **The collaboration publish flow.** Two members on one collab: open the
   `+ invite` chip, invite, accept from `/updates?filter=requests`, both write
   their part, then **Propose publishing** (it asks first) and **Approve — this
   one publishes it**. The closing beat should be the collab success screen. The
   `+ invite` chip, the propose confirm and the success beat all gained a
   `data-testid` and nothing else — no styles, no copy, no layout.
2. **The settled duel rail.** `/praxis/{id}` after both sides seal: the
   "The duel" panel with both names and the live tally
   ("…floats with the votes until era reset"). Unchanged by this PR and still
   asserted on its shipped copy, since those strings are current.
3. **The sign-up button on all eight task-detail archetypes** — it gained an
   attribute. Nothing should look different on any of them.


pixieofhugs:
## Dispatched-run proof

Read from the runs, not predicted from the patch.

| | baseline (`main`, run `32557107772`, and again after #2466) | run 1 (`32604430826`) | **run 2 (`32605378628`)** |
|---|---|---|---|
| failed | 38 | 38 | **37** |
| did not run | 8 | 8 | **2** |
| passed | 12 | 12 | **19** |

**Both heads pass.** Neither `duel.spec.ts` nor `collaboration.spec.ts:247` appears
anywhere in run 2's `--log-failed`; the 37 failures are the 36 S.N.I.D.E./faction
contrast cases (#2450, untouched here) plus one new head described below.

**Six of the eight suppressed tests now execute**, and five of those pass:

| test | run 2 |
|---|---|
| `collaboration` C1 — kick from the roster pill | **passes** |
| `collaboration` C2 — standalone Leave | **passes** |
| `collaboration` C3 — holdout is not shown unsubmit | **passes** — but see the vacuous-anchor note in the PR body; it matches nothing |
| `collaboration` C4 — oldest of six pending invites | **fails** (see below) |
| `presence` P1 — each member sees the other caret | still skipped, behind C4 |
| `presence` P2 — a solo author draws no caret | still skipped, behind C4 |
| `duel` D1 — an active duel can be dissolved | **passes** |
| `duel` D2 — the opponent can withdraw a pending challenge | **passes** |

### Run 1 → run 2: the third shared cause

Run 1 was not wasted — it walked both lifecycles far past where they had ever
reached and exposed a defect invisible from the source:

> `useEditPraxis.publish()` refuses an **untitled** praxis client-side, and
> neither flow mints a title. `handleSignup` posts a task id and a type and
> nothing else; `accept_duel` mints the opponent's praxis with `body_text=""`
> and no title at all.

And it refused *silently*, because `publish()` dismisses the seal sheet and the
propose confirm on its **first line** — deliberately, so an error lands in plain
sight rather than behind an overlay. Both specs' post-conditions were therefore
green on a submit that never left the browser:

- run 1's duel failed four steps later, on a rail that could not exist, because
  the opponent's side was never submitted and the duel stayed `active`;
- run 1's collab failed with Bob still seeing `data-collab-signal="propose"`,
  because Alice's proposal had never been made.

Both specs now type a title, as a player does, and the seal asserts the praxis
actually left the composer instead of asserting that an overlay closed.

Run 1 also proved a second real constraint: the room's document is written to the
record on a **trailing-edge 2 s debounce**, and that same flush is what fires
`on_room_edit`. A propose issued inside that window is cancelled two seconds
later by the proposer's *own last keystroke*. The collab spec now waits for the
flush (`body_text` reaching the record) before proposing.

### What run 2 newly reveals — the next head, and it is not this issue's

`C4` fails on a **fixture premise, not on the app**:

```
Error: need 6 distinct seeded tasks for the >5-invite test
Expected: 6
Received: 2
```

`pickTasks` wants six distinct-titled tasks at level ≤ 8 and the e2e database
holds **two** — the onboarding task and the UA duel fixture `ensure_duel_fixture_task`
that #1676 added. This is the same class of gap as #1676's, from the same root
(#1398: Era 1 declares no tasks), and it is why P1/P2 remain skipped: C4 is now
the first failure in that serial file.

Closing it needs a `seed_dev_demo` helper beside `ensure_duel_fixture_task` in
`backend/seed.py` — **a backend change, outside this PR's file scope**, and the
triage of what the eight then say belongs to #1795 by this issue's own terms. It
is recorded here rather than fixed here, and rather than muted: turning C4's
missing fixture into a `test.skip` would unblock P1/P2 at the cost of hiding it,
which is the move that produced three "fixed" e2e regressions in the first place.

pixieofhugs:
## Confirmed on the final SHA — and it is deterministic, not a lucky run

Run 2 (`32605378628`) and run 3 (`32606229452`, the head commit) are **identical**:

```
37 failed · 2 did not run · 19 passed
```

against the baseline's `38 failed · 8 did not run · 12 passed`. Same shape as the
issue's own standing condition — this suite does not flake, it reports.

Run 3 exists because run 2's SHA did not survive CI. `data-testid="composer-title"`
contains the literal substring `id="composer-title"`, and
`composerRule.test.tsx` asserts the composer markup does **not** contain it —
#2179 removed that `<label htmlFor>` id and that assertion is what keeps it gone.
The slot was renamed to `praxis-title`; the guard was not touched. (My local
`vitest run` had been green because it predated that one attribute, which is the
argument for the full suite at the end and not only the file you just edited.)

Nothing else changed between run 2 and run 3, and the two agree line for line.

**Still outstanding, and not this PR's:** `C4` fails on a fixture premise — it
wants six distinct seeded tasks and the e2e database holds two — which keeps `P1`
and `P2` skipped. It needs a `seed_dev_demo` helper in `backend/seed.py`, outside
this PR's file scope. See the previous comment.

Issue #2425, opened 2026-08-22
CI runs every job twice on every commit, and no concurrency group cancels superseded runs


pixieofhugs:
> *Filed from a CI audit with Molly (2026-08-22).*

**Every commit runs all three CI jobs twice.** `.github/workflows/test.yml` is `on: [push, pull_request]`, and for a PR branch in this repo both events fire, so `test`, `api-schema` and `frontend` each run on two separate run ids about four seconds apart.

Verified on #2424's head `c2300cbf`:

```
api-schema  success  started=2026-08-21T19:20:22Z
api-schema  success  started=2026-08-21T19:20:26Z
frontend    success  started=2026-08-21T19:20:22Z
frontend    success  started=2026-08-21T19:20:26Z
test        success  started=2026-08-21T19:20:23Z
test        success  started=2026-08-21T19:20:26Z
```

That is 2x the Actions minutes and 2x the runner contention on every push, on the slowest job in the repo (`frontend` does `npm ci` + lint + two typechecks + 6975 vitest tests + a Vite build + the byte budget).

## There is no `concurrency` block either

Neither workflow declares one. Pushing a fixup while CI is mid-flight leaves the superseded runs going to completion, so a branch that gets three quick corrections can have nine jobs alive at once.

## The duplication is not purely waste — read this before deleting it

The twin run is currently being **used as a signal**. Because `backend/pytest.ini` had no timeout, a wedged `test` job presented as "in progress forever" with no named failure, and the documented triage move was: *if one `test` hangs and its twin passed on the same SHA, it is a flake and not the diff.*

That workaround is now largely obsolete — #1816 shipped `timeout = 60` in `pytest.ini`, so a hang becomes a named failure on its own. But the loss is real and should be a conscious trade, not a surprise.

## The change

`.github/workflows/**` is protected by a PreToolUse hook (`~/.claude/hooks/deny_ci_config_edits.py`) because `main` has `autoDeploy: true` — a merge is a production deploy, and an agent that gutted a required job while keeping its name would report green and merge itself. **So this one needs Molly's own editor. An agent cannot do it and should not try to route around the hook with shell commands.**

At the top of `.github/workflows/test.yml`, replace:

```yaml
on: [push, pull_request]
```

with:

```yaml
on:
  push:
    branches: [main]
  pull_request:

# One run per ref. A superseded push cancels its own in-flight jobs rather
# than racing them to completion.
concurrency:
[...]

Issue #1816, opened 2026-08-15
Backend CI hung in the praxis-room suite, and no pytest timeout turned it into a failure


pixieofhugs:
## What happened

On PR #1806 the backend `test` job wedged with no further output after:

```
tests/integration/test_praxis_media_batch.py ............                [ 49%]
```

The next file in collection order is `tests/integration/test_praxis_room.py` (the CRDT room suite, #1788). The job sat in progress for ~10 minutes and was cancelled and re-run by hand.

Run: https://github.com/pixieofhugs/WorldZeroPlayground/actions/runs/31908502501

**It is a flake, not a regression.** The workflow fires on both `push` and `pull_request`, so the same commit ran twice; the twin run's `test` job passed in 3m54s (https://github.com/pixieofhugs/WorldZeroPlayground/actions/runs/31908500956). #1806's diff is `frontend/e2e/` only, which PR CI does not run.

## Ruling: ship the guard, close the hunt

One green control run on the same SHA is the only evidence either way, and a flake cannot be reproduced on demand. **The timeout is accepted as the guard.** This issue closes when it lands; the next occurrence arrives as a named test with a stack trace instead of a percentage, which is what makes hunting cheap. Do not go hardening the room suite on a suspicion as part of this.

## Scope

1. Add `pytest-timeout` to `backend/requirements.txt` (the dev/CI file — it must not reach `requirements-prod.txt`, see that file's header).
2. `backend/pytest.ini` gains `timeout = 60`.

That is the whole agent-facing change.

**Take the plugin's defaults deliberately — do not set either of these:**

- `timeout_method` stays `signal` (the POSIX default). It raises `pytest.fail`, so the session continues and fixtures still finalise. The `thread` method would `os._exit(1)`, killing coverage reporting and the `--cov-fail-under=80` gate with it.
- `timeout_func_only` stays `false` (the default), so the clock covers **fixture setup and teardown**, not just the test body. This is load-bearing — see below.

## Why 60 seconds

- The whole backend suite runs in 3m54s green.
- The longest deliberate `sleep` in any test is `0.1`.
- `_TIMEOUT_SECONDS = 5.0` bounds the room suite's own convergence waits.
- `_LOCK_WAIT_SECONDS = 30.0` (`backend/services/praxis_room.py:128`) is the largest bounded wait in the code under test — a legitimate test may sit there for 30s.

60 is double that ceiling. No honest test can trip it. If a test does, that is the finding.

[...]

PR #2568, opened 2026-08-24
Modify GitHub Actions workflow for main branch

pixieofhugs:
Updated workflow triggers and added concurrency settings.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,15 @@
 name: Test
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+# One run per ref. A superseded push cancels its own in-flight jobs rather
+# than racing them to completion. main is exempt — see below.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
   test:


pixieofhugs:

**An agent cannot make this edit.** `~/.claude/hooks/deny_ci_config_edits.py` blocks agent writes to `/.github/workflows/` (exit 2), because `main` has `autoDeploy: true` — a merge is a production deploy, and an agent that gutted a required job while keeping its name would report green and merge itself. Routing around the hook with shell commands is explicitly out. So this one needs your own editor.

This duplicates **#2425**, which carries the full analysis and is already `ready-for-human`. Worth closing one of the two.

## The exact edit, verified against `origin/main` today (not the issue snapshot)

`.github/workflows/test.yml` currently reads, at the top:

```yaml
name: Test

on: [push, pull_request]

jobs:
```

There is **no** `concurrency` block anywhere in the file. Replace that header with:

```yaml
name: Test

on:
  push:
    branches: [main]
  pull_request:

# One run per ref. A superseded push cancels its own in-flight jobs rather
# than racing them to completion.
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true

jobs:
```

## Two things to keep

1. **Keep `push: branches: [main]`.** That is the post-merge trunk run, and it is the only thing that catches the failure mode where parallel PRs each restate a shared registry row: every branch green, `main` red after squash. Dropping it to `pull_request` alone would remove that net.
2. **Do not touch `.github/workflows/e2e.yml`.** Verified today: it is `schedule` (nightly cron) + `workflow_dispatch` only, so it never races itself and needs no concurrency group.

## The one thing being given up

The twin run was being used as a flake signal — a wedged `test` whose twin passed on the same SHA meant flake, not diff. That crutch is now mostly obsolete: #1816 shipped `timeout = 60` in `backend/pytest.ini`, so a hang becomes a named failure on its own. Worth making that a conscious trade rather than a surprise.

## Verify after merging

```
gh api repos/pixieofhugs/WorldZeroPlayground/commits/<sha>/check-runs \
  --jq '.check_runs[] | "\(.name) \(.status)"'
```

Each of `test`, `api-schema` and `frontend` should appear exactly **once** instead of twice.

Issue #970, opened 2026-07-24
na kit fidelity: bring Default comment to the redesign (rainbow @mentions + spectrum hairline)


pixieofhugs:
**Context.** `default` ≡ `na` ≡ Unaffiliated is ONE visual identity — the spectrum-band editorial na kit (CONTEXT.md "Default archetype"; ADR-0039/0046/0048). Design project `1eb2665a-85f8-46cc-acc9-62e054f06100` freshly redrew this surface's na look. The shipped code is token-clean (not a wrong-identity bug like the cork-board edit-praxis) but predates the redesign, so it needs a fidelity pass. Design URLs are reachable only from the main Claude session via DesignSync — vendor the design into `.design-sync/` before building (see docs/agents/design-fidelity.md).

**Surface:** Comment box / voice (default/na).
**Shipped:** `DefaultComment` defined inline in `frontend/src/components/comments/CommentThread.tsx` (~L30). It is token-clean but uses `factionCssVar(slug,'card-accent')` (neutral) — it is MISSING the na tell the design specs.
**Design:** project `1eb2665a-85f8-46cc-acc9-62e054f06100`, file `Default Comment Edit.dc.html` — 'a clean bubble with a spectrum hairline and rainbow @mentions.'
**Acceptance:** the na comment shows the spectrum hairline + rainbow @mentions (reach the rainbow via `factionFill`/`--faction-default-rainbow`, not `factionCssVar` which is grey for na); other-faction comment voices unaffected.

Issue #2649, opened 2026-08-25
Epic: 26,000 lines of faction chrome are hand-authored nine times — give the five seamless families a model


pixieofhugs:
## The problem

> **Table re-measured 2026-09-01 (owner-requested, during `/bug-bot`). Every `shared 0` below was stale**, and three families were missing. The old figures are kept underneath so the epic's original sizing is still readable. The *argument* is unchanged — if anything the corrected numbers sharpen it, because the families that DO have a shared module still have archetypes that ignore it.

Archetype families largely hand-author their chrome. Four now have a shared module and one still has none at all — but **having a shared module is not the same as the archetypes using it**, which is the distinction the original table could not draw:

| family | lines | per archetype | shared module | archetypes that mount it |
|---|---|---|---|---|
| characterPaths | 9,458 | 525 (18 files) | **1,230 borrowed** — `editPraxis/archetypes/shared.tsx`, plus 508 local | **14 / 18** — na's create + edit and Albescent's two wrappers do not (#2992, #2991) |
| taskDetail | 7,770 | 823 | 364 — `taskDetail/archetypes/shared.tsx` | 9 / 9 |
| praxisDetail | 6,997 | 777 | **864** — `praxisDetail/shared.tsx` | 9 / 9, but 9–15 names each (#2718) |
| proposeTask | 6,459 | 718 | 158 — `factionSurfaces.ts`, **dying with #2993** | **7 / 9** on the borrowed chassis; na + Albescent on neither |
| factionDetail | 4,856 | 540 | **372** — `factionDetail/sectionDisclosure.tsx` | 9 / 9 |
| characterProfile | 3,956 | 347 | **830** — `archetypes/profileSkin.tsx` | **7 / 9** — na and Albescent render it zero times (#2996) |
| comments/voices | 2,079 | 260 (8 files) | **323** — `components/comments/shared.tsx` | 8 / 8 |
| **factionHero** | **1,893** | **210** | **none — 0 lines, no shared module at all** | **—** |

[...]

PR #2654, opened 2026-08-25
Comment voices become a token set: three of nine migrate, and the six that did not are the finding (#2650)

### LINKED ISSUE #2650: Comment voices become a token set: the cheapest test of "chrome resolves from tokens"
## Why this one first

Child 1 of the seamless-families epic, and it is first because it is the **cheapest possible test of the rule**: *chrome resolves from tokens; ornament is drawn by an archetype; a surface that is only chrome should have no archetype file at all.*

If the easiest case fails, the rule dies here for the price of one small family instead of 26,000 lines.

## State

`components/comments/voices/` — eight files, 2,035 lines, **zero shared pieces mounted, zero `<svg>`, zero shapes, zero `factionCssVar` calls**. The only family in the repo with nothing on either sharing axis.

| voice | lines | inline `style={{` | module consts | `--faction-*` refs |
|---|---|---|---|---|
| Albescent | 48 | 0 | 0 | 0 |
| UA | 203 | 7 | 0 | 18 |
| S.N.I.D.E. | 249 | 11 | 2 | 20 |
| Coven | 273 | 12 | 11 | 15 |
| WOW | 303 | 11 | 13 | 16 |
| Ephemerists | 306 | 11 | 3 | 5 |
| Everymen | 310 | 9 | 7 | 10 |
| Singularity | 343 | 19 | 15 | 14 |

**They are already halfway there.** Every voice reads `--faction-*` tokens directly — 5 to 20 references each. What is duplicated is not the decision, it is the *plumbing*: eight files each writing out which token goes in which slot.

And `shared.tsx` already carved the right seam. Its own docblock:

> These helpers keep the three invariant slots — author identity · body · timestamp+edited — and the composer mechanics out of every voice **so each voice only owns its chrome.**

That sentence is this issue's premise. The chrome is what is left, and the measurement says the chrome is tokens.

## Scope

**1. Name the slots.** Read all eight voices and write down every distinct chrome slot: sheet, ink, muted ink, rule/edge, the composer's field, the metadata label's voice, the author line, the owner-controls foot row. Expect roughly eight to twelve.

**2. Mint `--faction-{slug}-comment-*` for them**, one set per faction, values taken **verbatim** from what each voice paints today. This is a transcription, not a redesign — see the pixel-identical constraint below.

**3. One shared `Comment` reading those tokens**, dispatched as today by `CommentThread`. The three invariant slots and the composer mechanics stay in `shared.tsx`, untouched.

**4. Delete the voice files that turn out to be pure chrome.** That is the whole point: a faction whose comment is a sheet, an ink and a rule needs a token set, not a component.

**5. Keep a slot for anything genuinely drawn, and say so.** Known cases from a first read:
   - **UA** draws *"one dashed orange rule down the left edge, an ensō on the composer"*. A dashed left rule is a border token; the ensō is a mark from `sigil/` and needs a mount.
   - **Singularity** is the heaviest — 343 lines, 19 style literals, 15 constants. It may have real structure (a terminal frame) that a token set cannot express.
   - **Albescent** is 48 lines and already a wrapper; it should come out of this unchanged.

## The finding this issue owes

**The most valuable output is the list of what did not fit.**

If all eight collapse to token sets, say so and the rule is proven. If two of them keep a component, **name the slot that forced it** — that is the boundary between chrome and ornament, stated in evidence rather than asserted, and children 2–5 and the ADR will all cite it. A PR that quietly special-cases Singularity without explaining why is worth less than one that migrates six and writes a paragraph about the other two.

## Constraints

- **Pixel-identical, per faction, both themes.** A token migration that changes any rendered colour is a redesign wearing a refactor's clothes. Screenshot before and after for all eight.
- **No new colour, no new drawing.** Values are transcribed from what ships.
- Do not touch `shared.tsx`'s three invariant slots, the composer mechanics, `OwnerControls`, `FlagControl` or `useMentionAutocomplete`. This issue is the chrome only.
- Comment threads are **the densest text surface in the app** (UA's brief §5) — no ornament gains ground here as a side effect.
- Deleting a voice file is only allowed when the surface is pure chrome. `frontend/CLAUDE.md` forbids unifying archetypes; a faction that draws something keeps its file.
- `factionContrast.test.ts` is the gate — eight grounds, both themes, measured on the ground each ink composites to.
- Albescent's dispatch must still resolve; check `surfaceMap('comment')` still answers for it.

## Done when

- Every faction's comment chrome comes from `--faction-{slug}-comment-*`.
- Voice files remain **only** for factions that draw something, each with a one-line docblock naming what it draws and why a token could not carry it.
- All eight render byte-identical to today, verified visually in both themes.
- `factionContrast.test.ts` green on eight grounds.
- The PR description names what did not fit, or states that nothing did.

## Guard

A test that the shared comment resolves its chrome for every slug in `SURFACE_KEYS` — so a tenth faction with a token set and no component renders correctly by construction, which is the property this whole epic exists to buy.

Grilled 2026-08-24.

pixieofhugs:
Closes #2650

## The seam

**`--faction-<key>-comment-*`, between `index.css` and one shared `Comment`.** The failing test at that seam is `frontend/src/components/comments/__tests__/commentChrome.test.ts`: for every faction the chassis can be asked to paint, a *complete* token set must exist in `index.css`. It went red on the real defect first (no chassis, no tokens), and it is non-vacuous — it asserts `na` is always in its list and that the slot vocabulary has more than twenty entries.

## What landed

- **`frontend/src/components/comments/Comment.tsx`** — one chassis, fifty named chrome slots, not a colour in the file. It resolves every slot off the author's slug via `factionCssVar`, and `COMMENT_CHROME_SLOTS` is the exported vocabulary two guards now derive from.
- **~140 declarations in `frontend/src/index.css`**, three sets (`default`, `coven`, `ephemerists`). Every colour slot is an **alias** onto the token that faction already shipped — the one its voice named inline — so the migration cannot change a pixel by construction, and `commentChrome.test.ts` refuses a `#` in the block.
- **Deleted: `CovenComment.tsx`, `EphemeristsComment.tsx`** (579 lines) and their manifest rows. `na`'s voice moved out of `CommentThread.tsx` into the chassis. Albescent's wrapper is unchanged and `surfaceMap('comment')` still answers for it (`albescentWrapperKinds.test.tsx` and `surfaceDispatch.test.ts` both green).
- **`--faction-coven-braid`** — the braid data URI became a token so the class and the chassis read one value; `.cvn-braid` is repointed at it and its dark override is gone. Byte-neutral.

### How pixel-identity was verified without a browser

A throwaway probe rendered each old voice and the new chassis for the same props, resolved every `--faction-*-comment-*` through `index.css`, and diffed the per-element style streams. **Every difference is either a no-op declaration (`inherit` / `0` / `transparent` / `none`) or one of two deliberate structural changes that compute to the same box:**

- the foot rule becomes its own zero-height element with a `border-top` instead of a `border-top` + `padding-top` on the controls row (16px margin → 1px line → 8px gap → controls, both ways);
- Ephemerists' marginalia rule becomes the body block's own `border-left` + `padding-left` instead of a stretched flex sibling (1px + 16px + content, both ways; `opacity: .55` on solid ochre composites identically to `color-mix(… 55%, transparent)`).

The probe is not committed — it needs the deleted files to run.

---

## THE LIST OF WHAT DID NOT FIT

Six of nine kept their files. **In five of those six, what kept them is not ornament** — and that is the headline finding for children 2–5 and the ADR.

| voice | what kept it | could a token carry it? |
|---|---|---|
| **S.N.I.D.E.** | `Ransom` — the author's name cut letter by letter from five scraps, **one DOM node per character** | No. It is a drawing of a string. (Its halftone raster and −0.6° tilt *would* have been tokens.) |
| **UA** | mounts `UaSigil` where the avatar goes on the composer, and puts the avatar **inside** the dashed frame rather than in the margin | No. A `var()` cannot name a component, and cannot move a child. |
| **Everymen** | the author identity and timestamp **leave the body** for a red byline band above it — on the composer too, where every other voice draws no header at all | No. A different tree, not a different paint. |
| **Singularity** | adds **content** — a `>` prompt, a `0x…` handle derived from the comment id, `username` in place of `display_name`, a blinking caret — and **re-points three shared `--color-*` tokens around three subtrees** | No. A token set has no way to say "for the controls inside this element". |
| **na** | resolved @mentions are a gradient **CLIP** (`.rainbow-ink`, #970) | No. `MentionText` takes an ink *string*; a `background-clip` fill is not a colour. The chassis carries this one as a predicate (`!isKnownFaction(slug)`), which is the single non-token branch in the file. |
| **WOW** | the crest in the margin is **44px**, and `FactionAvatar`'s prop is `size?: 'sm' \| 'md' \| number` | **No — and this is the cheapest one to fix.** A custom property cannot cross a JavaScript prop. Its ribbon crown, its 3px section rule at 0.75, and the rule at the composer's head were all chrome and would have been three tokens. |

**So the chrome/ornament axis the epic proposes is the wrong cut for most of this.** Only one of six is held by ornament. The real split is **paint versus tree**: a token changes what a node looks like; it cannot add a node, move a node, or scope a variable to a subtree. Children 2–5 should be measured against *that* question, not against `<svg>` counts.

### Three more findings the epic should price in

**1. The vocabulary is fifty slots, not eight to twelve.** A slot is not "the byline"; it is every property of the byline that any voice varies, and the voices vary face, size, weight, style and tracking independently. Sheet 9 · author line 6 · caption 7 · body 6 · foot 6 · composer 9 · misc 7. A "no opinion" value has to be `inherit`, never a neutral — `font-weight: 400` *overrides* what a sheet inherits, and only `inherit` transcribes "this voice did not set this".

**2. The seam trades a free byte for a render-blocking one.** The deleted voices were `lazyArchetype` dynamic imports, entirely off the initial-load critical path. Their replacement is `index.css`, which is not. **CSS went 22.1 → 23.2 KB gzipped: +1.1 KB for three factions on one surface**, against a 22.2 warn and a 24.4 fail. `npm run budget` now WARNs (non-blocking; nothing else in CI moved). Colour, layout and type may never be deferred — only motion may — so there is no lazy path. **Four more families at this rate does not fit**, and the epic needs an answer before child 2: shorter names, a `[data-cmt="<key>"]` block instead of a `--faction-<key>-comment-*` set, or a raised ceiling with the reason written down.

**3. A token migration blinds four guards that scan source text, and each one wanted a *tighter* pin, not a wider one.**
- `factionTokensReferenced` cannot see a name built as `` `comment-${slot}` `` — its suffix list is now **derived** from `COMMENT_CHROME_SLOTS`, so it cannot go stale.
- `fontsLoaded` reported Cormorant Garamond 500 as dead: the only surface rendering it stopped naming a family and a weight in one brace scope. It now reads a `-face`/`-weight` **token pair by prefix** — which is a *tighter* pin than a brace scope and cannot cross-pair the way reading a whole `:root` block would.
- `frozenThemeAliases` took **forty-two** entries in one go. That is the tax, in a list a human is supposed to read.
- `factionContrast` — see below.

### Contrast

Per the mid-task correction: `factionContrast.test.ts` walks hand-curated pair arrays and would never have seen a new token family. A `COMMENT_PAIRS` block was added — 21 pairs × 2 themes = **42 new assertions, all green, none allowlisted**. It measures the **slot**, not the token behind it, because an alias can be repointed: the day someone moves `--faction-coven-comment-body-ink` off `-slip-ink`, the existing rows keep measuring `-slip-ink` and say nothing about the comment. Ground is `-comment-sheet` (Ephemerists' `-plate-wash` is `none` in both cascades, so a veil would composite nothing). `default`'s mention ink is deliberately excluded — na renders it as a clip and `MentionText` omits `color` entirely, so a row would measure an ink nothing paints.

## CI

`npm run lint`, `npm run typecheck`, `npm run typecheck:design-sync`, `npm test` (437 files / 8248 tests), `npm run build` — all green. `npm run budget` WARNs on CSS as described. No backend, route or schema change, so `test` and `api-schema` are untouched. `.design-sync/config.json` and the `.ds-kit` barrel were regenerated with `node .design-sync/gen-barrel.mjs`; the dead `EphemeristsComment` preview is deleted and `Comment` is published in its place.

**Visual QA is outstanding.** There is no browser and no dev stack here — everything above is tests, `tsc`, eslint and a markup diff. Nothing has been looked at.

## What to eyeball

Every item is a comment thread on a praxis or task detail page, in **both themes**, in the row state, the "edited" state, the composer, and the owner's hover-revealed foot row.

1. **Cozy Coven, light and dark** — the 4px candle-glow crown across the top of the slip, and the **braided foot rule** under a comment you own (the `.cvn-braid` mark, now driven by `--faction-coven-braid`). The braid is the riskiest single item in this PR: it moved from a class to a token and back.
2. **Cozy Coven, both themes** — the "· edited" mark must still be **lowercase Caveat**, not the uppercase Quicksand small caps the timestamp beside it uses. Those are the only two slots in the family that differ from each other.
3. **The Ephemerists, light and dark** — the **ochre margin rule** beside the note body, in the row AND in the composer. It is now a `border-left` rather than a stretched flex sibling; check it runs the full height of the prose and that the 16px gap to the text is unchanged.
4. **The Ephemerists, light** — the leaf has **no crown**. Confirm nothing draws a hairline under its 1px top border (a zero-height element is now always rendered there).
5. **na / Unaffiliated, light and dark** — the 3px **spectrum hairline** across the top of the sheet, and gradient-clipped `@mentions` in the body. Both are na's only tells and both changed hands.
6. **Albescent, both themes** — must be byte-identical to na. It is a pass-through wrapper over the chassis now.
7. **Any faction, dark** — a comment rendered **inside a Singularity or S.N.I.D.E. profile page**, which nests `data-theme="dark"` on a skin container. The new tokens are aliases and are declared on `:root, [data-theme]` for exactly this; if one froze, it shows here first.
8. **WOW and Everymen** — unchanged, but worth a glance beside the migrated three to confirm the family still reads as nine distinct sheets rather than three defaults and six exceptions.


pixieofhugs:
Closed unmerged by owner ruling (2026-08-25). The build did its job as a kill-test and the finding is recorded on #2650 and #2649; the migration is not shipping. Branch `worktree-agent-aff3899f4cb5b1fad` is deliberately left undeleted for inspection.

Issue #2519, opened 2026-08-23
One carrier per object: three Albescent surfaces ship the faintest spectrum where the design draws a 3px border


pixieofhugs:
Found by auditing the merged #2504 (PR #2517) against `Faction-Detail.dc.html` on the epic's
design canvas. **The canvas was never vendored before #2504 was dispatched** — that was an
orchestration miss, and the agent built from the issue's prose, which is faithful to epic
#2496 but does not carry the drawing's geometry or its copy.

Nothing here is a bug in the sense of "broken". Both are "correct, but not what was drawn".

## 1. The plate ring is 1px at 60% where the design draws a 3px border at full strength

**Shipped.** `.alb-plate-edge` joins the shared `.spectrum-frame::before` selector list in
`index.css` and overrides nothing, so it takes that rule's defaults:

```css
padding: 1px;
opacity: 0.6;
```

— a 1px masked ring, dimmed to 60%.

**Drawn.** `Faction-Detail.dc.html` gives every plate a real 3px spectrum border at full
strength, travelling:

```css
.alb-task .alb-plate{border:3px solid transparent!important;
  background-image:linear-gradient(var(--faction-default-card-bg),var(--faction-default-card-bg)),var(--faction-default-rainbow-loop)!important;
  background-size:100% 100%,300% 100%!important;
  background-repeat:no-repeat,repeat!important;
  background-origin:border-box!important;background-clip:padding-box,border-box!important;
  animation:alb-edge-travel 16s linear infinite}
```

The board says why it matters, and it is the whole hierarchy of the page:

> *"The spectrum border is back on every plate — **it is the Albescent tell**. The prism WASH
> stays on the hero alone, so the borders sing against a flat cream ground instead of a tinted
> one, and body copy keeps its contrast."*

Three times the width and two-thirds again the strength. On a page whose entire light-mode
delta is "the borders talk", shipping the dimmest possible version of the border is the one
thing that would make the page read as unchanged.

**Note the two idioms are not interchangeable.** The shared ring is a *masked overlay* (`mask
… content-box exclude`); the design uses a *real border* with the ramp in `border-box`. The
card surfaces already use the border idiom (`DefaultTaskCard`), so this is the kit's own
default-card treatment, not a new one.

Also unbuilt from the same board: the two card-holding sections (`.alb-open`) are drawn
carrying the **rainbow rule instead of the border** — *"the edge is for plates that hold text,
[...]

Issue #2504, opened 2026-08-23
The faction page gets an na frontispiece, a plate and a champion, and Albescent wears them


pixieofhugs:
Part of #2496.

Needs the foundation child (for the plate primitive).

`/factions/albescent` routes through `AlbescentGate`, which hands revealed accounts `<FactionDetail slug="albescent" />` -- so manifest dispatch **does** reach it.

### 1. `DefaultFactionHero` -- new na component

`pickVariant(surfaceMap('factionHero'), slug)` has **no fallback**, so Albescent (and any future neutral faction) gets bare title-and-description chrome while seven factions get a frontispiece. Build the na one, derived from the canvas's Albescent hero by subtracting the wash and the motion:

- 3px **static** spectrum in `border-box` -- the kit's own default-card idiom, already worn by the na task card and select card
- flat `--faction-default-card-sheet`
- grid `auto minmax(0,1fr) auto`: sigil slot | kicker + name + tagline | stat trio behind a hairline
- phone: stacked, counts as a three-column row under a hairline

**Five slots -- sigil, kicker, name, tagline, counts -- all fields on the faction record.** Nothing about the shape is Albescent-specific.

The sigil is `FactionSigil` by slug, **not** a hardcoded mark: na renders its single swept ring, Albescent the labyrinth. The hero knows about neither.

Ruled in scope although na has no faction page today, because a future era may have one.

### 2. Champion plate -- no backend needed

The design's Champion is the highest-scoring member of a list the page already has. Derive it client-side from `members`. Only ask for a server field if the ranking should be something other than points.

### 3. `AlbescentFactionBody` / `AlbescentFactionHero`

One wrapper plus one manifest row each. Delta: the hero takes the prism sheet, the labyrinth turns, the plates take the travelling border -- and, in dark only, the bloom (ruling 9).

### Also missing in na

`DefaultFactionBody`'s Members section is marked `PLACEHOLDER: design to restyle` in the file itself, and there is no Champion plate at all.

Issue #2332, opened 2026-08-19
Owner copy pass: 47 faction strings, including a UA rename, S.N.I.D.E.'s new punk register, and seven deliberate PLACEHOLDERs


pixieofhugs:
Owner copy pass, taken from her working spreadsheet and reconciled against the shipped catalogs during /grill-me on 2026-08-18. **Every value below is final and proofread -- apply them verbatim. Do not re-derive them from the sheet, and do not "improve" them.**

## What this is

The sheet is the voiced set from the 2026-08-16 faction-copy review (`docs/copy/faction-copy-voiced.csv`) -- 107 concepts x 9 factions. Reconciled against `frontend/src/locales/en/`:

- **218** cells already match what ships -- no action
- **47** differ -- **this issue**
- **108** name keys the catalogs do not have -- see the note at the bottom, NOT in scope
- 590 cells are blank -- not a signal, just unfilled

## Three owner rulings that change how you read this

**1. `PLACEHOLDER` is intentional. Ship it.** Seven values below are the literal string `PLACEHOLDER`. That is deliberate:

> *"I'm okay replacing some things with placeholder. Right now, that copy is placeholders. Genuinely it's my job to write the website, and if I see PLACEHOLDER in a place, I might realize what needs to be there later."*

Do not substitute the existing copy, do not write your own, do not skip the row. Two of the seven (`descriptions.wow`, `factionHero.wow.motto`) arrived as sentences that stopped mid-clause; the owner ruled they become `PLACEHOLDER` too, so a trailing comma never reaches a player and nobody "fixes" it later without knowing it was deliberate.

**2. S.N.I.D.E.'s register changed on purpose.** `Let's F%$*ing GO`, `Go Break some S#!*`, `Wanna F%&* some S^&* up?` are not typos or leaks:

> *"Snide is a different tone that I'm trying. Snide is punk. If I don't like it I will change it back."*

Ship the masked profanity exactly as written -- same characters, same casing.

**3. UA is being renamed.** `names.ua` goes from `UA` to `Unwavering Artisans`, and the owner confirms this is intentional. **This one needs a look beyond the catalog edit:** that string is the faction's name on the select card, the hero, the roster, taunts and the faction page, and some of those were laid out around a two-character name. Check each surface renders and does not overflow; the abbreviation survives inside `descriptions.ua` ("also known as UA"), so nothing else needs to carry it.

[...]

PR #2777, opened 2026-08-27
Copy: retire the select-card eyebrows, coords and tag chips; cut wow's tagline to the Pig Latin

pixieofhugs:
Copy-Bot batch — the `factionSelect` block of the copy sheet.

## Applied

| Key | Was | Now |
|---|---|---|
| `wow.tags[0..2]` | Playful / Ceremonial / Absurdly Sincere | **deleted** (chip row removed from the card) |
| `ephemerists.eyebrow` | Exhibit C · no single here | **deleted** |
| `ephemerists.coords` | x 41.7° | **deleted** |
| `ephemerists.coordsPolar` | r ∞ · θ 12 | **deleted** |
| `albescent.eyebrow` | World Zero | **deleted** |
| `na.eyebrow` | World Zero · no colours yet | **deleted** |
| `wow.tagline` | Whimsy for everyone. — Imsy-whay or-fay every-oneway. | Imsy-whay or-fay every-oneway. |

DELETE means the slot is gone, so the render sites go with it: the corner
coords block and the small-caps line in `EphemeristsSelectCard`, the chip row in
`WowSelectCard`, and the label beside the sigil in `DefaultSelectCard` (mounted
by both the Albescent and Unaffiliated tiles). Two docblock rows that measured
retired inks were struck with them.

## Reconciled, no change

`wow.chosen` ("CHOSEN") and `everymen.banner` ("THE EVERYMEN WANT YOU") already
match the sheet byte for byte. So do all five shipped `name` values — Cozy
Coven, Warriors of Whimsy, Singularity, Albescent, Unaffiliated.

## Not in this PR (would mint keys)

The sheet gives a `name`, `tagline` and `banner` for all nine, but only five
factions have `factionSelect.*.name`, one has `tagline`, one has `banner`. The
other twenty-two cells name keys the catalogue does not have, and minting them
is card-layout work, not a copy edit — raised separately.

Also flagged, unchanged: the sheet's `name` for Ephemerists reads "Ephemerists";
`factions.json` ships "The Ephemerists". That is `names.ephemerists` under
ADR-0038, a different key family from this row.

Frontend suite green — 454 files, 9,293 tests. `tsc --noEmit` and eslint clean.

### DIFF
--- a/frontend/src/components/selectCard/CovenSelectCard.tsx
+++ b/frontend/src/components/selectCard/CovenSelectCard.tsx
@@ -133,12 +133,14 @@ export default function CovenSelectCard({ state = "locked", members, onVisit }: 
     }}>
       {/* The masthead, in the band's anatomy rather than the band itself: the
           faction's mark hard left, the coven's name hand-lettered beside it. */}
-      <div style={{ display: "flex", alignItems: "center", gap: "var(--space-sm)", padding: "var(--space-lg) var(--space-xl) 0" }}>
+      <div style={{ ...CAPTION, padding: "var(--space-lg) var(--space-xl) 0" }}>{i18n.t("feed:factionSelect.coven.banner")}</div>
+      <div style={{ display: "flex", alignItems: "center", gap: "var(--space-sm)", padding: "var(--space-sm) var(--space-xl) 0" }}>
         <CovenSigil size={34} color={DEEP} />
         <div style={{ fontFamily: HAND, fontSize: "var(--text-title)", lineHeight: 1, textTransform: "lowercase", color: INK, whiteSpace: "nowrap" }}>
           {i18n.t("feed:factionSelect.coven.name")}
         </div>
       </div>
+      <div style={{ padding: "var(--space-xs) var(--space-xl) 0", fontFamily: READING, fontStyle: "italic", fontSize: "var(--text-content)", color: SOFT }}>{i18n.t("feed:factionSelect.coven.tagline")}</div>
       <div style={{ flex: 1, padding: "var(--space-md) var(--space-xl) 0", position: "relative" }}>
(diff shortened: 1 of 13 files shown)


pixieofhugs:
Second pass, on the owner's three rulings.

**1. Uniformity — all nine get all three slots.** The sheet's `banner`, `name`
and `tagline` are now minted for every faction and rendered on every tile, in
this order: banner, name, tagline. Where a card already drew a slot it keeps its
own treatment; where it did not, the slot is drawn in that card's existing
register off tokens the contrast gate has already measured — no new paint, no
new tokens, no new contrast rows.

Snide and Ephemerists drew their name through `factionName()`; both now read the
same `factionSelect.{F}.name` key the other seven use, so one lookup answers all
nine. That is the first piece of the reuse this exercise is for.

The sheet leaves Albescent's and na's `banner` blank, so those two tiles have
none. Say the word if they should get one.

**2. Misspellings resolved to the shipped names.** Aritsins / Artists ->
Artisans, SNIDE -> S.N.I.D.E., Warriers -> Warriors, Ephermerits ->
Ephemerists. Casing and grammar are voice and are untouched — `The cozy coven
wants you` keeps its lowercase, though the Coven caption register uppercases it
on the tile.

**3. `factionHero.{F}.motto` is now its faction's tagline.** Eight blocks;
`factionHero` has no `na`. `ua` had no motto and gains one.

Two pinned lists follow those values: `factionHero.wow.motto` leaves the
PLACEHOLDER registry (#2332 put it there, this ruling takes it back out), and
the default-frontispiece test now asserts the new Albescent line that #2504 and
#2519 argued over.

**Worth knowing:** motto and tagline are two keys holding one string, which will
drift. Folding the hero onto `factionSelect.{F}.tagline` is one line per hero
and its own PR — not done here.

Frontend suite green — 455 files, 9,304 tests. `tsc --noEmit` and eslint clean.

Issue #2605, opened 2026-08-24
vite 8's Rolldown baseline puts initial-load JS into WARN, and the line was calibrated for esbuild


pixieofhugs:
> *Filed during `/builder-bot` on 2026-08-24, while merging the vite 5 -> 8 harness bump (#2567, PR #2604). Numbers measured, not inferred.*

**vite 8 replaces esbuild with Rolldown, and the initial-load JS grows ~4.4 KB gzipped as a result — enough to cross the `warn` line the moment #2604 lands.**

## The measurement

| | before (vite 5) | after (vite 8) | line |
|---|---|---|---|
| **JS** | 128.3 KB | **132.7 KB** | `warn>130.9 KB` — **crossed** |
| **CSS** | 22.3 KB | **21.7 KB** | `warn>22.2 KB` — cleared |

The growth is real and not a compression artifact: raw critical-path JS went **412,710 -> 425,649 bytes**, and concatenating the fourteen chunks before gzipping recovers only ~1,950 of it.

Note the CSS column. **vite 8 cut 600 bytes of CSS on its own**, which is a second, independent win on top of #2535's 211-byte deletion — the sheet is now comfortably under its line rather than 59 bytes beneath it.

## Why this is a decision and not a number to bump

This is the same fork #2469 resolved for CSS, and the ruling there was unambiguous: **the warn line follows a measured win downward; it never rises to meet a regression.** Four CSS ratchets, all downward.

But the JS case is genuinely different in one respect, and that difference is the whole issue:

- **The CSS regression was our own code growing.** Five days of faction work ate a 272-byte band. Raising the line would have excused work we did.
- **This regression is a bundler swap.** No application code changed — PR #2604 touches exactly two files, `package.json` and `package-lock.json`. The 130.9 KB line was calibrated against esbuild/rollup output; Rolldown emits a different, larger baseline for the same source.

So the honest question is whether the line is still measuring what it was set to measure. A budget calibrated for one bundler, applied unchanged to another, is not obviously a ledger of our progress any more — it may just be a units mismatch. **That is the owner's call, and #2604 deliberately did not price it.** `bundle-budget.mjs` was left untouched.

## The options, as they were put on PR #2604

1. **Re-cut the line for the new bundler.** Set `warn` just above the post-Rolldown number, with a ledger entry saying the baseline changed and why — same shape every prior entry takes. Accepts that the ratchet restarts from a new floor.
[...]

PR #2804, opened 2026-08-28
One CTA paint per faction: the card's button, worn by the page it opens

### LINKED ISSUE #2642: One CTA paint per faction: the task card and the task detail share a component but not a colour
## State

#2554 (PR #2637) put `CardCtaControl` on both surfaces — every task card and every task detail now mounts the same component, and both render `cta.label` from the one resolver. **The element and the copy are shared. The paint is not.**

Each faction passes a *different style object* at its two call sites:

| faction | task card | task detail |
|---|---|---|
| UA | `{...CARD_CTA, fontFamily: UA_DISPLAY, fontWeight: 600, …}` | `primaryAction` |
| Coven | `{...CARD_CTA, display: flex, gap, …}` | `pinkButton` |
| S.N.I.D.E. | `{...CARD_CTA, …}` | `plateButton(false)` |
| Everymen | `{...CARD_CTA, …}` | `primaryBar` |
| Ephemerists | `{...CARD_CTA, …}` | `primaryButton` + `.eph-cta` |
| Singularity / WOW / Default | `{...CARD_CTA, …}` | `primaryButton` / `ctaStyle(false)` / `primaryButton` |

Two hand-maintained paints per faction, **eighteen in total**, with nothing forcing the pair to agree — so they don't. #2554 left this deliberately; the comment above every detail call site says *"`CardCtaControl`, paint and geometry still this skin's, spread last."* That was the right call for that issue's scope and it is what this one changes.

**Owner ruling (2026-08-24): the card and the detail wear the same button — same colour, same effect, same component — per faction. The card's paint is canonical (option a).**

This is *within* a faction, not across them. UA's two buttons must match each other; UA's and Coven's must not.

## Why the card's paint wins

`components/taskCard/cardCta.ts` already exports `CARD_CTA` (and `CARD_CTA_ROW`), and eight cards spread it — one shared base, nine tints, the 44px hit-target floor in one place. The detail side has no shared base at all: eight unrelated local constants, three of them coincidentally named `primaryButton` in three different files. The drift lives on the detail side, so that is the side that moves.

## Scope

1. **One style constant per faction.** Lift each card's CTA style out of its component into a named export beside `CARD_CTA` (or into the faction's own module, whichever the existing layout prefers — say which and why). Both call sites spread that one constant.
2. **Delete the detail's local constant** where it served only the CTA. Several of them (`primaryButton`, `primaryAction`) are also used by *other* controls on the same page — **check every reader before deleting**, and leave the constant in place if something else spends it.
3. **Size, not paint, is the one legitimate difference.** The card CTA sits in a narrow column; the detail's stands alone in a panel and is larger. Carry that as a **prop on the shared control** (`size="card" | "detail"`, or a spread-in size token) — not as a second style object, which is how the pair drifted in the first place.
4. **Albescent inherits.** It has no task-detail or task-card archetype of its own; both wrap Default. Verify nothing needs a fourth thing.
5. **The `.eph-cta` class comes too.** Ephemerists' detail button carries a stylesheet class alongside its inline style (#2146: *"blue in light, brass in dark, one control across three surfaces"*). Whatever that class does, the card must do too — or the class moves onto the shared constant.

## Constraints

- `CardCtaControl` owns the **affordance semantics** and that contract does not move: a skin cannot reattach an affordance the server shut. Paint arrives as `style`, spread last. Do not push colour into the control.
- Do not unify the archetype components; `frontend/CLAUDE.md` forbids it. One shared control that nine archetypes **paint** is the sanctioned shape — the same as `CardMasthead` and `ScoreStamp`.
- The 44px WCAG 2.5.5 hit-target floor is geometry, not spacing, and must survive on both surfaces.
- Re-measure. The two grounds differ — the card's sheet is not the detail's — so a paint that was AA on one is not automatically AA on the other. `factionContrast.test.ts` is the gate, nine grounds.
- `data-testid="task-signup-cta"` stays on both; E2E reads it.

## Done when

- Each faction has exactly **one** CTA style constant, spread by both its card and its detail.
- No detail archetype declares a CTA-only style of its own.
- The two buttons are visually identical per faction apart from size, verified in a browser at both surfaces, in both themes, for all nine.
- `factionContrast.test.ts` green on both grounds for all nine.

## Guard

A test asserting that the card and the detail resolve the **same** style constant per faction — the failure mode is someone tweaking one call site and not the other, which is exactly how the eighteen came about.

Grilled 2026-08-24. Sibling of the "card gains the submitted state" issue filed alongside; they touch the same call sites and should probably land in that order.

pixieofhugs:
Closes #2642

Eighteen hand-maintained CTA paints become eight. A faction's task card and its
task detail now spread **one** style constant, and the only thing that differs
between the two mounts is a geometry token.

## The seam I tested at

**The rendered inline style of the sign-up control on both surfaces.** Not the
constant — a test that imported `UA_CARD_CTA` and compared it with itself passes
while a call site quietly spreads `{ ...UA_CARD_CTA, background: … }` over the
top, and a call-site override is exactly how the eighteen came about.

`frontend/src/components/taskCard/__tests__/ctaPaintPairs.test.tsx` renders each
card and each detail to static markup, finds the CTA by the `min-height:44px`
floor (asserting there is exactly one such element per surface), strips the
declarations `CTA_DETAIL_SIZE` owns — read off the token, so the test learns
nothing about which keys those are — and compares the rest. **Nine named
factions, one `it` each**, never a count.

It was proved red before it was left green: spreading `letterSpacing: "0.5em"`
over `UA_CARD_CTA` at the card call site alone fails with
`- "letter-spacing": "0.5em" / + "letter-spacing": "0.02em"`, naming UA.

## Scope decisions

**1. Layout: eight named exports in `components/taskCard/cardCta.ts`, beside
`CARD_CTA`.** The alternative was to export each card's constant from its own
`*TaskCard.tsx` and import it into the matching detail — a faction's paint next
to the rest of its paint. It makes a page-sized component the module a detail
route imports, and `UaMandala` / `BalloonBunch` / the Everymen fist path would
follow it into the detail's chunk for the sake of one object. `cardCta.ts` was
already the CTA's home and holds nothing else.

**A surface-prefixed role var cannot travel, so none is used.** Three cards read
their face through their own root's prefix (`--wow-task-card-face`,
`--ev-task-face`, `--sg-task-card-face`) and those prefixes exist only under the
card. The constants call `factionRoleVar(slug, 'face')` — the same value one hop
earlier, pixel-identical on the card. `--na-task-card-*` keeps its fallback arm.

**2. No detail constant was deleted: every one of the eight has another reader.**
`primaryButton` / `primaryAction` / `pinkButton` / `primaryBar` all still paint
"view your praxis" and "continue editing", and `pinkButton` is what `goldButton`
is composed from. Two collapsed rather than went, because the CTA was their only
caller of one arm:

- `SnideTaskDetail.plateButton(done)` → `plateButton` (the inverted exit). The
  false arm had no caller left.
- `WowTaskDetail.ctaStyle(ghost)` → `ghostCta`. Same shape. `PLUM_EDGE` lost its
  last reader on that page and is tombstoned in place.

**3. Size is a prop on the shared control, not a second style object.**
`CardCtaControl` gains `size?: "card" | "detail"` and spreads `CTA_DETAIL_SIZE`
— `{ width: "100%", padding: "var(--space-md) var(--space-xl)" }` — **after** the
skin's paint. That is the one deliberate inversion of the component's "paint
last" rule, and it is why the token is geometry only: a card's narrow-column
padding must not survive into a panel, and a colour in there would silently beat
the skin's.

**The size token deliberately carries no type rung.** Measured before it was
written: of the eight details, three set the same `font-size` their card did, two
a smaller one and three a larger. There is no "the detail is a rung up" rule to
preserve, so moving type would be inventing a difference. `font-size` sitting
beside `font-family`, `font-weight` and `letter-spacing` is the faction's voice,
not its scale.

**`cursor` moved into `CardCtaControl`, spread first.** It is affordance, not
paint: a pure function of `cta.denied`. All eight cards wrote the same ternary
and all eight details wrote a flat `cursor: pointer` that lied on a refusal. A
constant cannot hold a ternary, and this is the seam that already owns the
question. No colour went into the control.

**4. Albescent needs no fourth thing** — it wraps `DefaultTaskCard` *and*
`DefaultTaskDetail`, and the guard carries it as a row of its own rather than
assuming the inheritance.

**5. `.eph-cta` was already on both** call sites and still arrives as
`className`. It is what `EPHEMERISTS_CARD_CTA` deliberately leaves out: the
enclosure changes width between the cascades and an inline `background` or
`border` would beat the class and pin the light half in both themes.

**The link branch is covered.** #2799's `href` CTA takes the same `style` and the
same `size` — `CardCtaControl` applies both to the `<Link>` and the `<button>`,
and `data-testid="task-signup-cta"` is untouched on the detail.

## What actually changes on screen

The detail's sign-up takes the card's paint, so per the ruling these move:

- **na / Albescent** — solid inverted slab → the card's transparent outlined
  button.
- **Singularity** — keeps `--faction-singularity-term-cta-glow`, the ONE ruled
  per-faction exception (owner, 2026-08-28, recorded on #2642). Every other
  faction's change was a SWAP, one paint for another; subtracting Singularity's
  bloom was the only purely subtractive one, so the owner carved it out. It
  arrives as the named `SINGULARITY_DETAIL_GLOW`, spread after the shared paint,
  and `ctaPaintPairs.test.tsx` both ALLOWS its keys to differ and REQUIRES them
  to still be present — an allowance alone would let the glow be dropped again in
  silence. Note it resolves to `none` in the light cascade, so this is a
  dark-mode-only difference: `0 0 18px rgba(74, 222, 128, 0.18)`.
- **S.N.I.D.E.** — the skewed hard-shadowed plate becomes the sprayed stencil,
  halo and `rotate(-1.6deg)` and all. `plateLabel`'s counter-skew is dropped at
  that one site: counter-skewing a label whose button is no longer skewed is how
  a shared paint drifts back apart one span at a time.
- **WOW** — border moves from `--faction-wow-plum-edge` to the fill itself.
- **Everymen / Coven / Ephemerists / UA** — same family, small deltas in
  enclosure, tracking and type rung.

Children are unchanged: the Singularity prompt+cursor, the WOW star, the
S.N.I.D.E. X marks and the Ephemerists signs still only appear on the detail.
That is tree, not paint.

## Contrast

`factionContrast.test.ts` is green. Seven of the eight constants carry an opaque
background, so the measured pair travels with the button and no ground question
changes. The one that does not is na's transparent CTA, which lands on the
detail's `--color-bg-page` instead of the card sheet — re-measured:

| | ink on card sheet | ink on detail page |
|---|---|---|
| light | 18.22:1 | **16.86:1** |
| dark | 13.75:1 | **15.00:1** |

The 44px floor is asserted on both surfaces, per faction, in the new guard.

## Three guards followed the paint to its new home

Not widened — pointed at where the card's CTA is now spelt:

- `snideWearsTheClippingRegister` / `singularityWearsTheTerminalRegister` slice
  the faction's own constant out of `cardCta.ts` (which holds all eight), the
  same way they already slice one band out of `factionBands`.
- `everymenWearsTheBillRegister` reaches it through its existing
  `propsBehind(declarations(…))`.
- `ephemeristsCta.test.tsx` was unchanged: `cardCta.ts` spells the class with a
  leading dot so a module that only *talks* about `.eph-cta` cannot join a
  census of its mounts.

## Byte budget

Run before and after on this branch:

```
before   WARN JS 137.2 KB (warn>130.9)    ok CSS 21.7 KB (warn>22.2)
after    WARN JS 137.3 KB (warn>130.9)    ok CSS 21.7 KB (warn>22.2)
```

**The initial-load budget does not move, and no CSS was added at all** — every
file touched lives in a lazy route chunk. The JS WARN is pre-existing and is not
this change's.

Across the *whole* `dist` the change adds **+444 B gzipped** (761,633 → 762,077,
level 9), and I am not going to pretend a consolidation should have. It does not
remove bytes because scope item 2 says not to: seven of the eight detail
constants keep other readers, so nothing was deleted from the detail side, while
`cardCta.ts` now loads in the detail chunk as well as the card one and the size
token and prop are new code. What it removes is the eighteen-into-eight
maintenance surface, not weight.

## Verification

`npm run lint`, `npm run typecheck`, `npm run typecheck:design-sync`, `npm test`
(463 files, 9,561 passing), `npm run build`, `npm run budget` — all green on the
rebased tip. No backend, route or schema file is touched, so `api-schema` has
nothing to regenerate.

**Visual QA is outstanding.** I have no browser and no dev stack; everything
above is markup, tokens and arithmetic.

## What to eyeball

Each faction's **task card beside its task detail**, in **light and dark**, all
nine:

1. **na** — the transparent outlined button on the detail's inner box; it is the
   biggest single change in the PR.
2. **Albescent** — the same, under the prism ground, and that it is still na plus
   light.
3. **Coven** — the pink gradient box at full panel width; the 12px radius against
   the slip's corners.
4. **Ephemerists** — `.eph-cta`'s enclosure at 2px light / 1px dark on both
   surfaces, between the rune strips.
5. **Everymen** — the struck block full-bleed in the panel, and still `0 0 auto`
   between the two marks on the card.
6. **Singularity** — the lit key WITH its glow kept on the detail, in **dark
   mode** (the token is `none` in light, so there is nothing to see there); the
   panel button is no longer a grid child (`grid-column: 2` is inert there).
7. **S.N.I.D.E.** — the halo and the 1.6° rotation on a 100%-wide detail button:
   check it does not clip against the acid plate, and that the label no longer
   leans the wrong way.
8. **UA** — the chip paint at panel width, between the mandalas on the card.
9. **WOW** — the plum call with the fill as its own border; the knights still
   clear the 44px target on the card.

Plus, on any one faction: a **denied** CTA (`not-allowed` now reaches the detail,
which used to show a pointer on a refusal) and the **"Read your praxis" link**
branch from #2799, which takes the same paint and the same size.

---

## Correction to this body, 2026-08-28

The two passages above about Singularity's glow originally said the detail **lost**
it. That was true of commits 1–5 and stopped being true at commit 6, which added
the carve-out. The body has been corrected in place rather than appended to, so a
reviewer reads the branch as it stands.

The carve-out was held unmerged until the owner confirmed it: the commit and the
test docblock cited an owner ruling on #2642 dated 2026-08-27 that was **not in the
thread**, and the ruling that was recorded there (2026-08-24) said the opposite —
*"the card's paint is canonical."* Asked directly, the owner ruled on 2026-08-28 to
keep the glow. That ruling is now recorded on #2642 so it is no longer uncited.


pixieofhugs:
> *This was generated by AI during `/builder-bot`. **Handoff — the batch stopped here at the
> owner's instruction. This PR is finished and reviewed but was deliberately NOT merged.***

**State: ready for review-and-merge by whoever picks this up. Nothing is outstanding on the code.**

## Where it stands

- Head is `73af76f6`. **CI was still running at handoff — verify it against that SHA before
  merging, do not trust this line.** The previous head (`b9633dae`, before the glow commit) was
  green on all three required checks.
- Marked ready (not a draft), rebased onto `main` past #2803.
- Reviewed by the orchestrator. The one technical catch worth knowing is recorded below.

## The owner ruling that landed after the build

The consolidation cost Singularity's detail button its `--faction-singularity-term-cta-glow`,
because the glow is a treatment the CARD never had and the card's paint is canonical. **Owner ruled
2026-08-27: ship the consolidation, keep Singularity's glow.** The other three visible changes ship
as ruled.

That ruling is implemented in the final commit as `SINGULARITY_DETAIL_GLOW` — a **named export**
beside `CTA_DETAIL_SIZE`, not a loose `boxShadow` at the call site, because a one-off spread into
the detail's JSX is the exact shape this issue exists to remove.

`ctaPaintPairs.test.tsx` learns it two ways, both keyed by slug and both read off the constant:

1. the pair check **ALLOWS** those keys to differ — for `singularity` only; the other eight
   allowances are empty, which is what names them;
2. a new assertion **REQUIRES** them to still be present.

The second is the load-bearing one. An allowance only lets a key differ; it does not demand the key
exist. Without it, dropping the glow again passes the pair check in silence — which is exactly how
it was dropped in the first place. Proved red by removing the spread:
`singularity detail still declares box-shadow: expected false to be true`.

## Verified locally before push, with the toolchain proved running (tsc 5.9.3 reported)

`lint` · `typecheck` · `typecheck:design-sync` · **463 files / 9,570 tests** · `build` ·
`budget` — CSS **ok** at 21.7 KB; the JS WARN is pre-existing (#2605) and untouched.

## VISUAL QA IS OUTSTANDING AND THIS IS THE PR IN THE BATCH THAT MOST NEEDS IT

No agent in this batch had a browser. Four detail-page CTAs change appearance, and `main`
auto-deploys:

| faction | change |
|---|---|
| **na** | solid inverted slab → transparent outlined button — **the default kit, so most players** |
| **S.N.I.D.E.** | becomes the rotated sprayed stencil at full width (skew gone, so `plateLabel`'s counter-skew was dropped at that one site) |
| **WOW** | border moves to the fill |
| **Singularity** | glow PRESERVED per the ruling above |

Also re-check the `denied` and `href` branches — #2799 made the CTA slot resolve to a LINK
("Read your praxis") as well as a button, and the size prop has to work on both.

## The catch worth carrying forward, whoever reviews this

Three cards read their typeface through their **own root's prefix** — `--wow-task-card-face`,
`--ev-task-face`, `--sg-task-card-face` — and those prefixes do not exist on the detail page. A
naive lift of the constant would have **silently dropped three typefaces with lint, tsc, tests and
the build all green.** The constants reach `factionRoleVar(slug, "face")` instead: same value one
hop earlier, pixel-identical on the card.

**A constant spread on two surfaces may never read a surface-prefixed role var.** That is worth a
line in `WORLD_ZERO_STYLE.md`'s role-vocabulary section, since the failure is invisible to every
automated gate.

## Flagged by the builder, not acted on

- `EphemeristsTaskDetail.tsx` hardcodes `color={CTA_INK}` on the two `<Sign>` glyphs inside the CTA
  where the card's equivalent SVGs use `stroke="currentColor"`. Same value today.
- `DefaultTaskCard.tsx` spreads `factionRoleVars("na", "na-task-card")`, which returns `{}` by
  contract — so every `--na-task-card-*` read in the repo is always its fallback. Dead machinery.
- `SnideTaskDetail.plateLabel` and the `<XMark>` children are detail-only extras on a now-shared
  control; whether children follow paint wants its own ruling.

Issue #2486, opened 2026-08-23
The disabled Create control dims to 1.81:1 in dark, and that is the state character creation opens in


pixieofhugs:
> *Filed during `/builder-bot` on 2026-08-22, raised by #2350's builder and deliberately not forked into one faction's answer.*

**The disabled primary action on the character-creation archetypes dims to `opacity: 0.5`, which folds the band AND its label together over the sheet: 3.14:1 in light, 1.81:1 in dark.**

At 1.81:1 the label is effectively not there.

## Why it is cross-archetype and not a WoW bug

It is the shipped pattern on **both** sibling archetypes that predate the faction fan-out — `DefaultCreateCharacter`'s phone bar and `EphemeristsCreateCharacter`'s cast — so the WoW builder followed it rather than inventing a seventh answer to a question nobody had ruled. That was the right call for one PR and the wrong outcome for the surface. **Seven archetypes now ship it.**

## Why "disabled controls are exempt" is the wrong reading here

WCAG 1.4.3 does exempt an inactive component by name, so this is not a formal AA failure. But look at **who sits in that state**:

A first-run player, on a phone, in dark mode, lands on character creation with an empty form. **The Create button is disabled until they type a name** — so the disabled state is not a brief edge case a user passes through, it is *the state the page opens in*, and the primary action is the thing they are looking for. An exemption written for a greyed-out toolbar button is being spent on the first control a new player ever needs to find.

`opacity` also folds into the text alpha as well as the ground, which is why the composite is so much worse than the token suggests.

## The decision this wants

Not "which faction fixes it" — **one rule for all eight archetypes**, applied in one place:

- Dim the **band** and keep the label at full strength, so the control still reads as a control; or
- dim less (the ratio, not the alpha, is the thing to hit); or
- keep the dim and rule explicitly that a disabled primary need not be legible, which at least makes it a decision rather than an inheritance.

Whichever it is, it belongs beside the other seven archetypes' shared chassis, not copied eight times.

## Verify

- Character creation, **dark mode, 375px**, empty form, before typing anything: the Create control at rest.
- Then all eight archetypes, since they share the pattern.
- Confirm the enabled state is unchanged — this is only about the at-rest disabled reading.

pixieofhugs:
## Ruled 2026-08-23 — fix it. → `ready-for-agent`

[...]

Issue #2485, opened 2026-08-23
Four editPraxis composers ink their quiet tier on a washed ground and measured it flat: UA, Everymen, Coven and Singularity are below AA


pixieofhugs:
> *Filed during `/builder-bot` on 2026-08-22. **Four independent agents, working on four different factions in parallel, each found the same defect on their faction's `editPraxis` composer.** None of them fixed it — each was scoped to character creation — and none of them knew the other three had found it.*

**`editPraxis` composers measure their inks against the declared token and ship them over a washed ground.** The composer sheet paints an ornament layer under the content column, so the ground the type actually lands on is not the token the ratio was computed from. Four factions are below AA in at least one theme, today, on `main`.

## The four readings, each measured on the composite

| faction | ink | flat | composited | note |
|---|---|---|---|---|
| **UA** | `--faction-ua-card-muted` on `ComposerGround` | 5.45 / 5.64 | **4.38 light / 3.79 dark** | labels, counters, `Save draft`, `Drop`, roster quiet ink |
| **Everymen** | `--everymen-muted` on `.em-burst` paper | 5.09 | **4.26 gold corner / 4.09 olive corner** (light) | dark unaffected |
| **Coven** | `-slip-label` / `-slip-ink` on a hard `opacity={0.7}` bloom | — | **2.80 / 1.95** and **3.33 / 3.25** | a miss for every tier |
| **Singularity** | `-term-dim` on the raster + `.sg-scan` band | 5.03 / 5.80 | **4.12 / 4.20** | its own docblock claims "everything clears 4.5:1" |

## The single root cause

**A flat-token reading is optimistic by up to a full ratio point, and every one of these was signed off on the flat reading.** Two of them say so in their own comments:

- `factionContrast.test.ts` rules UA's composer covered because it *"draws entirely from `--faction-ua-{card-bg,panel}`"* — true of the **tokens**, and not true of the **ground**, because `ComposerGround` washes the lotus and ensō over them.
- `SingularityEditPraxis`'s docblock says *"the tightest are `dim` on the chrome bar (4.52) … everything clears 4.5:1"*. That is the flat reading. Composited it is 4.12.

This is the same failure #1674 recorded on 2026-08-22 for three S.N.I.D.E. pairings — *"specified correctly against the paper, failed against the four washes stacked on it"* — and the reason #2346's acceptance criterion now says **measure the composite, not the declared token.**

## Coven is a different flavour and worth calling out

[...]

Issue #2039, opened 2026-08-17
Task cards F: faction backdrops should follow the new cards


pixieofhugs:
Parent: #2027. **Needs a design before it can be built.**

Owner: the faction backdrops should follow the new task cards. **This is not drawn anywhere** — `Task Cards.dc.html` covers the cards only, so the look needs deciding before an agent can start.

## What exists today

Eight backdrops in `frontend/src/components/backdrop/`:

`CovenBackdrop` · `EphemeristsBackdrop` · `EverymenBackdrop` · `SingularityBackdrop` · `SnideBackdrop` · `UaBackdrop` · `WowBackdrop`, dispatched by `FactionBackdrop.tsx`.

There is **no Albescent and no `na` backdrop** — `na` falls through to `WatercolorBackground` and Albescent inherits it. That is the designed neutral, not a gap: #1317 established it and `docs/kit-structure.md` records it. **Any design here should say what it intends for those two rather than assume they are missing.**

## What "follow the new cards" has to mean

The cards gain a shared masthead anatomy (#2029), a shared CTA shape (#2030), and per-faction ornament (#2031–#2037). A backdrop that follows them is answering a question the cards do not: a card is a small object on a page, a backdrop is the page. The same ornament at page scale is usually wrong — the Coven cauldron works at 158px and would be absurd at 1400px.

So the design needs to say, per faction, **which of the card's marks scales up, which becomes a texture, and which does not appear at all.**

## Next step

A design page covering the eight backdrops beside their new cards. Once it exists, vendor it into `.design-sync/task-cards/` alongside `TaskCards.dc.html` and this becomes buildable.

pixieofhugs:
## Owner ruling 2026-08-17 — palette alignment only. Now `ready-for-agent`.

This was mislabelled `ready-for-human`. It is not human-only work; it just had one
ambiguity in "reflect the new cards", and that is now settled.

**The scope is palette alignment.** Each backdrop's wash is retuned so the page and the
card stop disagreeing — same token vocabulary, same light/dark behaviour. **No new
artwork.**

**Page-scale ornament is explicitly NOT this issue.** The compass rose, the cauldron, the
stamped seal and their siblings belong to the **point card** (the praxis-card score stamp)
— tracked in #2042 — not to the page behind it.

## What makes this small

Most backdrops are a `<div>` and a stylesheet class:

| component | lines | where the look lives |
|---|---|---|
[...]

Issue #1819, opened 2026-08-15
controls.tsx hardcodes --color-text-tertiary on .label-caption, defeating the per-faction --label-ink seam


pixieofhugs:
> *This was generated by AI during /builder-bot.*

Found while building #1793/#1800 (PR #1818), which set `--label-ink` on the second Ephemerists composer root.

`frontend/src/pages/editPraxis/archetypes/controls.tsx:1004` and `:1023` — the `bodyConnecting` and `bodyFrozen` notices — apply `.label-caption` **and then hardcode** `color: "var(--color-text-tertiary)"` inline.

That inline colour overrides the class, so it defeats the `--label-ink` seam the class exists to read. Measured **2.01:1** on the Ephemerists sheet in light. The composer is faction-dispatched, so the same two lines land on **S.N.I.D.E. and Singularity** at the same failing number — this is cross-faction, not an Ephemerists bug.

## Why it is its own issue

PR #1818 fixed the *root* (`dress.pageStyle` now sets `--label-ink`), which is what #1800 asked for. These two sites opt out of that root by restating the ink inline. Fixing them means **deleting the inline `color`**, not changing it — the class already reads the right variable once the restatement is gone.

That is the same shape as #1783's ruling: an inline restatement of a class value is a fork, and the fix is deletion rather than a swap.

## Check before assuming it is only these two

`controls.tsx` is shared composer chrome. Grep for other inline `color:` next to a `.label-caption` / `.label-heading` className in that file and its siblings before scoping — the two above are where the Ephemerists work happened to look, and this repo's issues have repeatedly understated their own extent.

## Done when

- The `bodyConnecting` / `bodyFrozen` notices read `--label-ink` through `.label-caption` with no inline override.
- The measurement is re-derived on each faction composer ground the notices actually render on, not just Ephemerists.

## Related

- **#1800 / PR #1818** — set the seam these two bypass
- **#1754 / #1636** — the two-Ephemerists-roots precedent
- **#1783** — the "delete the restatement, do not edit it" ruling

pixieofhugs:
## Owner ruling (2026-08-16): fix the sites, and add a narrow guard

Verified on `origin/main` @ `e5c2c726`. Two corrections to the issue as filed.

## Correction 1 — the line numbers drifted

The notices are at **`controls.tsx:1025` and `:1044`**, not `:1004`/`:1023`.

## Correction 2 — it is four sites in that file, not two

[...]

PR #2830, opened 2026-08-28
The composer's unavailable write-up box takes a measured ground, not an unmeasured 0.6

### LINKED ISSUE #2574: The composer's unavailable write-up box dims to an unmeasured 0.6, and it is the one site .control-off cannot reach
> *Filed during `/builder-bot` on 2026-08-24, from reviewing #2566 (which closes #2557) against #2573 (which closes #2486). Not a defect #2566 introduced in isolation — it is the one site the disabled-contrast fix cannot reach.*

**The composer's write-up box dims to `opacity: 0.6` while the collaborative room seeds, and that ratio has never been measured.** It is one number covering eight faction skins in both themes.

## Where it is

`frontend/src/pages/editPraxis/archetypes/controls.tsx`, in `BodyTextarea`:

```ts
const BODY_UNAVAILABLE_STYLE: CSSProperties = {
  cursor: "not-allowed",
  opacity: 0.6,
};
```

It is applied to the editor host while `awaitingRoom` is true. #2566's own author marked it with a `ponytail:` comment naming the ceiling and proposing `--composer-unavailable-opacity` as the upgrade path.

## Why #2573 does not already cover it

#2573 (closing #2486) minted the house answer for this exact failure mode — an opacity that folds a band **and its label together**, so the text loses contrast twice:

```css
:root               { --control-off-fill: #e2ddd2; }
[data-theme="dark"] { --control-off-fill: #2a2833; }
:root, [data-theme] { --control-off-ink: var(--color-text-secondary); }
.control-off:disabled { … }
```

6.31:1 light, 6.49:1 dark. But it hangs off **`:disabled`**, which only exists on form controls. The write-up box is a CodeMirror `contenteditable` host, not a button, so the class does not apply and there was nothing for #2566 to adopt.

## Why this is lower priority than #2486 was, and filed rather than blocking

#2486's disabled Create control is **the state character creation opens in** — a first-run player on a phone in dark mode meets it immediately. This one is transient: it shows only while the room seeds, and it resolves on its own within a second or two on a healthy connection.

It is also a strict improvement on what shipped before it. On `main` prior to #2566 the box carried **no** unavailable treatment at all — it looked perfectly ready, which is the defect #2557 was filed for. A dim at an unmeasured ratio is better than no dim.

## What this needs

- **Measure `0.6` on the real composited ground, in both themes, on all eight skins** — not against the declared token. The three near-black composer sheets are the ones to check first: a dim there can read as "broken" rather than "waiting", which is the failure #2486's ruling names ("the disabled state must still read as disabled" has a twin — it must not read as *broken*).
- If it fails, the likely shape is the one #2573 landed: stop dimming the composite. Give the unavailable box its own quiet ground and keep the placeholder ink at a measured value, rather than multiplying opacity into both.
- The `ponytail:` comment proposes a `--composer-unavailable-opacity` custom property a skin may repoint. Prefer extending #2573's `--control-off-*` family over minting a parallel one, if the geometry allows — two vocabularies for one state is how the fifth site repeats itself.

## Also worth sweeping while here

#2573's author flagged two more inline opacity dims of the same shape, deliberately left alone as out of scope:

- `frontend/src/components/comments/shared.tsx:312` — `opacity: disabled ? 0.5 : 1`
- `frontend/src/pages/Contact.tsx` — Tailwind `disabled:opacity-50`, 4 sites

Both are buttons, so both would take `.control-off` cleanly. Neither is a state a page opens in.

## Check

`factionContrast.test.ts` is the gate. Whatever value lands, assert it there by name rather than by count — a census that counts sites lets a new one be wrong twice.

pixieofhugs:
Closes #2574

## The seam

The **composite the unavailable write-up box actually paints** — not the declared token, and not the rendered markup (`bodyUnavailableDress.test.tsx` already owns that half and stays green). While `awaitingRoom` is true the only text in the box is CodeMirror's placeholder, which `bodyEditorTheme.ts` inks from `var(--label-ink)`; the ground under it is the skin's `fieldBox` background; and `opacity` on the host composites **both of them together** over the composer sheet behind. So the pairing is `dim(placeholder)` against `dim(field)`, per skin, per theme, resolved out of `index.css` in node the way `factionContrast.test.ts` and `composerGround.test.ts` already do. No browser needed, which is good, because I have none.

## The measurement

`opacity: 0.6` **fails.** Eight skins × two themes; the placeholder live, then dimmed on the real composited ground (the washed sheet where a skin washes one, modelled the way `composerGround.test.ts` does).

| skin | light: live → dimmed | dark: live → dimmed | ground shift at 0.6 (ΔE, light / dark) |
|---|---|---|---|
| na | 8.46:1 → **3.16:1** | 7.90:1 → **3.77:1** | 4.1 / 5.7 |
| ua | 6.67:1 → **2.83:1** | 7.17:1 → **3.52:1** | 4.5 / 6.5 |
| wow | 6.98:1 → **2.75:1** | 7.57:1 → **3.60:1** | 3.7 / 3.3 |
| coven | 7.67:1 → **2.91:1** | 9.13:1 → **4.05:1** | **1.0** / 4.9 |
| snide | 6.88:1 → **2.74:1** | 8.33:1 → **3.77:1** | **1.9** / **2.2** |
| ephemerists | 8.21:1 → **3.12:1** | 5.52:1 → **2.85:1** | 3.1 / **1.6** |
| singularity | **2.07:1** → **1.52:1** | 8.99:1 → **4.11:1** | 5.7 / 6.1 |
| everymen | 7.29:1 → **3.01:1** | 7.40:1 → **3.61:1** | 4.8 / **1.9** |

**All sixteen pairings land under AA. Nine land under 3:1.** Worst honest reading 2.74:1 (S.N.I.D.E., light). This is #2486's failure mode verbatim, at its sixth site: `opacity` composites the element as a group, so the ground fades toward the sheet *and* the ink fades over the faded ground.

### And no alpha rescues it — which is the finding, not the number

Every composer field is the same family of stock as the sheet behind it, so fading one toward the other spends the whole budget on the ink and buys almost no ground. Sweeping the alpha (worst placeholder ratio across all sixteen, and the median ground shift; Singularity-light excluded because it is broken at α = 1.0 — see below):

| α | light: worst ink / median ground ΔE | dark: worst ink / median ground ΔE |
|---|---|---|
| 0.60 | 2.74:1 / 4.5 | 2.85:1 / 5.8 |
| 0.70 | 3.39:1 / 3.4 | 3.41:1 / 4.4 |
| 0.80 | 4.24:1 / 2.3 | 4.03:1 / 3.0 |
| 0.85 | 4.77:1 / 1.7 | **4.38:1** / 2.2 |
| 0.90 | 5.35:1 / 1.1 | 4.74:1 / **1.5** |

The ink only clears AA in *both* themes past ~0.9, by which point the ground has stopped moving at all (ΔE 1.1–1.5 is under the just-noticeable threshold). **There is no value that reads as disabled and stays legible.** So the `ponytail:`'s proposed `--composer-unavailable-opacity` is refused on the arithmetic rather than on taste: a token whose every possible value fails is a name for a fix that does not exist. Nothing in that family is minted, and a guard asserts none ever is.

## The fix — #2573's shape, #2573's vocabulary

`BODY_UNAVAILABLE_STYLE` keeps only `cursor: not-allowed` (which has to be inline — it is inherited, and `@codemirror/view`'s base theme declares none, so this is what reaches `.cm-content`, the element the pointer is really over). The ground is **replaced, not faded**, in `index.css`:

```css
[data-composer-body][aria-disabled="true"] {
  background: var(--control-off-fill) !important;
  color: var(--control-off-ink) !important;
  --label-ink: var(--control-off-ink);
}
```

- **The vocabulary is `--control-off-*`, unchanged.** `.control-off` itself cannot reach here — it hangs off `:disabled`, which only exists on form controls, and this is a `contenteditable` host. That is why the *selector* is new and the *family* is not. Two names for one state is how the seventh site repeats itself.
- **The fill is opaque, which is the load-bearing half.** The ink lands on the token instead of on eight faction sheets, so there is **one** ratio: **6.31:1 light / 6.49:1 dark**. Against the eight live fields the neutral sits ΔE 3.4–20.4 on fifteen of sixteen pairings, so it reads as a change of state without a second measurement.
- **`--label-ink` is the third declaration and it is not optional.** `.cm-placeholder` is painted from that seam (#1819), not from `color`. Without it the one visible string keeps the skin's own quiet tier over a ground it was never measured against.
- **`!important` for the reason `.control-off:disabled` gives:** the live paint is inline, because it is the skin's `fieldBox`, and inline beats any stylesheet.
- **The border is left alone.** The frame is what still says whose composer this is, and what keeps the box a box; only the fill and the ink are the state.

### Why the Singularity needs a second override, and why it is a ground and not a family

`--control-off-fill` measures **ΔE 80.6** from the Singularity field in light — the terminal is theme-invariant, so this would be a full-width, 200px-tall pale slab on a black chassis, which is the exact failure #2573 minted `.sg-control-off` to avoid. But it cannot simply wear `.sg-control-off`: that class answers the **publish band's** ground, which stands on the chassis so the panel is a raised box and a visible change. The write-up box *is* the panel, so the same override would move the ground **ΔE 0.0** and say nothing.

So it takes the rung below — `--faction-singularity-term-page`, the desk under the chassis (#2039) — reached through a new `.sg-composer-off` that declares the same two custom properties and no paint of its own:

| | ratio | ΔE from the live field |
|---|---|---|
| light | 5.17:1 | 7.0 |
| dark | 5.93:1 | 4.7 |

Both inside the band the other fifteen pairings sit in. One optional `className` on `BodyTextareaSkin` carries it — the same seam `PublishButton` already gives its band, one caller.

## Found while measuring, deliberately NOT fixed here

**The Singularity composer's placeholder is 2.07:1 in light *even when the room is live*.** `.cm-placeholder` reads `--label-ink`, the Singularity composer does not re-point it, and the global `--color-text-tertiary` on a near-black terminal panel is 2.07:1. That is #1819's unreachability at a site nobody had checked, it is true of the working box and not only the unavailable one, and repointing `--label-ink` on that page moves every `.label-caption` on it — a different change with a different blast radius. Recorded in `index.css` beside the override and worth its own issue.

## Out of scope, as #2574 asks

`components/comments/shared.tsx:312` (`opacity: disabled ? 0.5 : 1`) and `pages/Contact.tsx` (`disabled:opacity-50`, 4 sites). Both are **buttons**, so both take `.control-off` cleanly; neither is a state a page opens in. Left alone; one small separate issue if wanted.

## Where the guard went, and why not `factionContrast.test.ts`

`frontend/src/pages/editPraxis/archetypes/__tests__/bodyUnavailableGround.test.ts`, new, 33 assertions. The brief asked for `factionContrast.test.ts`; `composerGround.test.ts` argues at length (from #2485, and from `factionContrast`'s own #1179 note recording two red `main`s) that a *composite* row does not go in that file, and the landed pair is `--control-off-*`, which that file does not own either. So the ratio is not restated anywhere — it stays gated once, in `characterPaths/__tests__/disabledControlContrast.test.ts`.

**Everything is by NAME, nothing by count.** The rule is read out of the stylesheet and its three declarations are named; the override is read out of the stylesheet and both halves are named, with a ΔE floor of 3 (the *smallest* shift the house neutral makes on any live field) so it cannot land on the panel it stands on; the archetype wearing it is named and the other seven are asserted not to. The refusal is arithmetic, so it survives re-derivation: at 0.6 the placeholder misses AA on all eight skins in both themes. That half is measured on the **bare** sheet — the optimistic ground — because `composerGround.test.ts` already proves the wash is the tighter reading, and re-deriving its layer model here would be a second copy free to drift.

Verified red first: 7 of the 33 fail against `origin/main`'s sources.

## Verification

`npm run lint`, `npm run typecheck`, `npm run typecheck:design-sync`, `npm test` (470 files, 9725 passed), `npm run build`, `npm run budget` — all clean. CSS 21.7 → **21.8 KB**, still `ok` under the 22.2 warn line; the JS `WARN` at 137.2 KB is **pre-existing and byte-identical on `origin/main`** (checked). `api-schema` and the backend `test` job are untouched — this diff has zero backend contact.

**Visual QA is outstanding.** No browser, no dev stack.

## What to eyeball

The composer's write-up box **while a collaborative room is seeding** (it shows for about a second on a healthy connection — a throttled network or a cold socket holds it longer), in **both themes**:

1. **Singularity** — the near-black terminal, the one skin re-pointing the pair. The box should read as a panel that is not lit yet, one rung under the chassis, with the placeholder in dim phosphor. It must not go pale, and it must not become the loudest thing on the page.
2. **S.N.I.D.E.** — the other near-black composer sheet, and the tightest ground shift the house neutral makes anywhere (ΔE 3.4, light). Check the box still visibly changes state rather than just sitting there.
3. **Ephemerists** — the third near-black sheet, and the one skin whose placeholder ink is its own (`--faction-ephemerists-plate-quiet`) rather than the global tertiary. Confirm the neutral ground does not fight the plate's frame.
4. **Coven, light** — where the old dim was ΔE 1.0, i.e. invisible. The new state should be plainly different from the live box; this is the before/after worth flipping between.
5. **Any skin** — that the box goes **all the way back** to its faction paint the moment the seed lands. Dressing a working field as dead is the same bug with the sign flipped.
6. **The border on every skin** — it is deliberately untouched, so the frame should stay the faction's while the fill goes neutral.


pixieofhugs:
> *This was generated by AI during `/builder-bot`, 2026-08-28.*

**Handoff — built, green, and deliberately left UNMERGED for the next batch at the owner's
instruction.** Not blocked, not in doubt: the batch was ended after the build.

## State as I left it

- **Draft.** `gh pr ready 2830` is the first step, and note it does **not** re-trigger CI —
  `test.yml` is `on: [push, pull_request]` with no `types:`, so `ready_for_review` fires nothing.
- All three required checks **pass** against head `3916f895`; `mergeStateStatus` was `CLEAN` at
  hand-off, but `main` moves — `gh pr update-branch` just-in-time and confirm the run belongs to the
  current `headRefOid` before believing it.
- `+470/−16` across 5 files. Author and head repo both `pixieofhugs`.

## What has NOT been done, and is owed before merge

**No review pass has been run against this PR.** Everything above is CI plus the building agent's
own report. Eight PRs merged in this batch each had a review pass; this one has not had one, and
CI green is not review. Run it.

Three things a reviewer should look at specifically, all flagged by the building agent itself rather
than found by me:

1. **The guard did not go where the issue asked.** #2574's brief says to assert the landed value in
   `factionContrast.test.ts`. It went to a new sibling, `bodyUnavailableGround.test.ts`. The stated
   reason is that `composerGround.test.ts` (#2485) owns composite rows and `factionContrast` has its
   own #1179 note about two red `main`s, plus the landed pair is `--control-off-*`, which
   `factionContrast` does not own. **That reasoning sounds right and I did not verify it** — check
   it rather than taking it, because "the guard moved" is exactly the kind of deviation that reads
   as reasonable and hides a gap.
2. **It edits two stateful `.tsx` files** — `controls.tsx` (named in the dispatch brief as the site)
   and `SingularityEditPraxis.tsx` (one `className` line). The agent flagged this against its own
   scope line. Both edits are presentational; confirm that.
3. **A new `.sg-composer-off` rule** exists because `.sg-control-off` could not be reused: that
   class answers the publish band's ground (panel on chassis), and the write-up box *is* the panel,
   so reusing it would have moved ΔE 0.0. Worth confirming, since "why not reuse the existing class"
   is the first question a reviewer should ask and it has a real answer.

## Why the measurement mattered — `0.6` genuinely failed

This issue could legitimately have ended in "measured, fine, change nothing", and the brief said so
explicitly. It did not. All sixteen readings (8 skins × 2 themes) came in under AA, nine under 3:1.
The alpha sweep is the decisive part: **no alpha reads as disabled and stays legible**, because every
composer field is the same stock family as its sheet, so opacity spends everything on the ink. That
is what refused the `ponytail:` comment's proposed `--composer-unavailable-opacity` — on arithmetic,
not on taste.

## One thing found while measuring and filed, not fixed here

**#2831** — the Singularity composer's placeholder is **2.07:1 in light while the room is LIVE**.
`.cm-placeholder` inks from `--label-ink` (#1819) and that skin never re-points it. It is a defect in
the composer's *working* state, so it is out of this PR's scope, and this PR's `.sg-composer-off`
rule means the *disabled* state on that skin is fine. Do not let #2831 block this.

## Out of scope, confirmed left alone

`components/comments/shared.tsx:312` and `pages/Contact.tsx` (4 sites). Both buttons, both would
take `.control-off` cleanly, neither is a state a page opens in. #2574 says explicitly that sweeping
them is a separate issue.

Issue #2843, opened 2026-08-28
Both budget lines sit in WARN: find the cuts, do not move the lines


pixieofhugs:
> *Filed during `/builder-bot` on 2026-08-28, carrying forward the owner ruling recorded on #2605.*

**Both budget lines in `frontend/scripts/bundle-budget.mjs` are in or near WARN, and the ruling is
that WARN is acceptable while a cut is looked for — not that the lines move.** This issue is the
look.

## The standing ruling (owner, 2026-08-28)

> I am okay with JS and CSS being in WARN territory as long as [there is] an issue to look into
> bloat and see if we can make cuts.

Two things follow, and they are the constraints on any PR that closes this:

1. **`warn` and `fail` do not move.** `bundle-budget.mjs`'s own header argues this, #2469 ruled it,
   and #2605 was filed to ask whether a bundler swap was an exception. It is not — the answer is to
   find bytes, not to re-cut the line. A PR that closes this by editing a number is the wrong PR.
2. **A standing WARN is accepted in the meantime.** Nothing is red, the step exits 0, and no PR is
   held up. Agents should stop reporting it as a finding on unrelated work.

## Where the two numbers stand

**These are the last figures written down, not a fresh measurement. Re-measure before acting on
them — the first job of this issue is a current build.**

| | last recorded | line | source |
|---|---|---|---|
| JS | ~132.7 KB gzipped (135,900 bytes) | `warn: 134_000` | #2605, measured across the vite 5 -> 8 bump (PR #2604) |
| CSS | 22,641 bytes | `warn: 22_700` | the `#2535` ledger entry — 59 bytes of headroom at the time |

The CSS figure predates roughly two weeks of faction work that has merged since (#2804, #2818,
#2819, #2823, #2827, #2830, #2832, #2837–#2840 among others). The `#2535` entry says out loud that
59 bytes is "tripped by one selector", so CSS is expected to be over the line now. Confirm it.

## What is known about the JS half

`vite 8` replaced esbuild with Rolldown and the entry grew ~4.4 KB gzipped with **no application
code change** — PR #2604 touched only `package.json` and `package-lock.json`. Raw critical-path JS
went 412,710 -> 425,649 bytes, and concatenating the fourteen chunks before gzipping recovers only
~1,950 of it, so it is real emitted code rather than a compression artifact.

That means the JS cut has to come from what the entry path *pulls in*, not from the bundler:

[...]

PR #2920, opened 2026-08-29
Upgrade to React 19 (with the dev-dependency group)

pixieofhugs:
Supersedes #2906, #2910 and #2911.

Dependabot split the React 19 upgrade across two PRs that each do half the job, so each fails `npm ci` on its own: `react-dom@18` pins `react: ^18.3.1` and `@types/react-dom@18` pins `@types/react: ^18.0.0`. Bumping either side alone leaves a dangling peer edge. All four move here in one commit. The dev-dependency group from #2906 rides along because it touches the same two files and would otherwise conflict on the lockfile.

| package | from | to |
|---|---|---|
| `react` | `^18.3.1` | `^19.2.8` |
| `react-dom` | `^18.3.1` | `^19.2.8` |
| `@types/react` | `^18.3.3` | `^19.2.18` |
| `@types/react-dom` | `^18.3.0` | `^19.2.5` |
| `@types/node` | `^26.2.0` | `^26.3.0` |
| `@typescript-eslint/parser` | `^8.67.0` | `^8.68.0` |
| `eslint` | `^10.8.1` | `^10.9.1` |

The lockfile is a plain `npm install`, then verified with a clean `npm ci`. **No `--legacy-peer-deps`, no `--force`, no `overrides` block.** None was needed: every other react consumer in the tree declares an open-ended peer range — `react-markdown` and `react-router` at `>=18`, `react-easy-crop` at `>=16.4.0`, `react-i18next` at `>= 16.8.0`, `use-sync-external-store` already listing `^19.0.0`. Nothing was forced past a genuine incompatibility.

The resolved-version diff against `main` contains only the seven packages above plus their direct consequences (`scheduler` 0.23.2 → 0.27.0; `@types/prop-types` and `loose-envify` dropped, since React 19 no longer depends on them; the `@typescript-eslint/*` internals that move with the parser). Notably **`typescript` stays at 5.9.3** — inside `@typescript-eslint/parser@8.68.0`'s `>=4.8.4 <6.1.0` peer — and `openapi-typescript` does not move, so the `api-schema` gate's generator is unchanged and cannot drift.

## The seam being tested

Not a pure version bump after all. Two React 19 behaviour changes broke 26 tests across 5 files, and the seam in each case is **the point that produces raw markup for a structural assertion**, not the components themselves. Both are absorbed there rather than papered over at the assertions.

**1. React 19 hoists image preloads.** Every `` reaching `renderToStaticMarkup` now also emits a `<link rel="preload" as="image">`, hoisted to the front of the output. React 18 emitted none, so suites read the first tag as the component's own root and broke: `undress()`'s anchored `^<div class="alb-praxis ...">`, `factionAvatar`'s `^<[a-z]+[^>]*>`, and an `indexOf` on an image filename that now finds the hint's `href` before the ``'s `src`. New `src/test/ssrMarkup.ts` peels the leading run; applying it inside `markup()` in `src/test/praxisDetail.tsx` fixed 20 of the 26 on its own. Nothing under `src/` authors a `<link>`, so a leading one is always React's — and the strip is scoped to the leading run, so a `<link>` appearing mid-markup would still fail a test rather than vanish.

**2. `useId` changed format** — React 18 minted `R3`, React 19 mints `_R_3_`. `detailPanelReuse` already normalised the Everymen roundel's per-mount SVG id for exactly this reason, but anchored on the `R` sitting directly after the dash, so under 19 it silently normalised nothing and failed on ids rather than on the markup it is about. The suffix is now matched loosely, since the format is React's to change.

One production change was needed: `PortraitPicker`'s `inputRef` prop. `@types/react` 19 folds `MutableRefObject` into `RefObject`, so `useRef<HTMLInputElement>(null)` now yields `RefObject<HTMLInputElement | null>`. Nine character-path archetypes pass such a ref and every one failed `TS2322`. All nine route through a single prop declaration whose own fallback `ownRef` already has the nullable type, so widening the prop once fixed all nine call sites with no archetype changes.

This app is a client-rendered Vite SPA — `renderToStaticMarkup` is a test inspection tool here and none of that markup is ever served.

## Gates

Every step of `.github/workflows/test.yml`'s `frontend` job run locally against a real `npm install` in this worktree (not a junction to another checkout's `node_modules`, which would have measured the old React):

- `npm ci` — clean, zero ERESOLVE
- `npm run lint` — pass
- `npm run typecheck` — pass
- `npm run typecheck:design-sync` — pass
- `npm test` — **478 files, 9935 passed, 1 skipped, 0 failed**
- `npm run build` — pass
- `npm run budget` — pass (exit 0)

The `api-schema` job's `npm ci` is covered by the same lockfile, and neither the backend nor the generator version moved, so `git diff --exit-code` stays clean. The backend `test` job is untouched by a frontend dependency bump; it was not run locally (a worktree carries no venv).

### One number worth knowing

The byte budget passes but moved: initial-load JS went **136.9 KB → 151.2 KB gzipped** (+14.3 KB, measured against `main`'s own CI run for 98b385f). That is still under the `fail>175.8 KB` ceiling so the step exits 0, but `main` was *already* over the `warn>130.9 KB` line, and this consumes roughly a third of the remaining headroom (38.9 KB of slack → 24.6 KB). CSS is unchanged at 21.9 KB. Not a blocker, and not something I traded away quietly — flagging it because the next payload-growing change has meaningfully less room.

## Visual QA is outstanding

I have no browser or dev stack; everything above is tests, types, lint and build. **Nothing here has been looked at.** React 19 changes batching and effect timing, which is exactly what the following depend on:

### What to eyeball

- **The collaborative composer** (`pages/editPraxis`) — CodeMirror + Yjs + `y-websocket`. The highest-risk surface by far: it wires an imperative editor to React effects, and effect ordering/cleanup changes are precisely where an editor binding double-mounts or drops. Check that two windows still sync, remote cursors and presence render, and reconnect works.
- **The avatar cropper** — `react-easy-crop` inside `ImageEditModal`. It declares a `>=16.4.0` peer, so it installs happily, but it does measurement in effects. Check the crop box tracks the pointer and the saved image is not off by a frame.
- **Any transition or animation that depends on effect ordering** — anything sequencing an enter/exit against a state flip, and the reveal/theme-flip surfaces.
- **The Everymen roundel** on task detail, since its SVG `<defs>` id format changed (`R3` → `_R_3_`). The `<textPath href="#...">` must still resolve, or the "★ VERIFIED ★ ON THE RECORD" ring silently disappears. Tests assert the markup contains it; only an eye confirms it paints.
- **Images generally**, given the new hoisted `<link rel="preload">` — avatars, praxis media, the media gallery.


pixieofhugs:
*Review note, /builder-bot.*

**The +14.3 KB is mostly a chunk boundary moving, not React 19 alone — and it costs cache in a way the budget number cannot show.**

| chunk | main | this branch |
|---|---:|---:|
| `index-*.js` (entry) | 44.4 KB | **98.3 KB** |
| `react-dom-*.js` | **41.4 KB** | 1.3 KB |
| `react-*.js` | — | 2.8 KB |
| `jsx-runtime-*.js` | 2.9 KB | 0.6 KB |

`vite.config.ts` declares no `manualChunks`, so main's 41.4 KB react-dom chunk was Rollup's **automatic** split. React 19's changed module graph dissolved it and the runtime moved into the entry chunk, leaving a 1.3 KB re-export shim behind.

The budget measures a cold first load, where that move is close to neutral. The part it cannot see is the second visit: a content-hashed vendor chunk survives an app-code deploy and the entry chunk does not, so a returning visitor now re-fetches ~40 KB of unchanged React on **every** deploy — and `main` auto-deploys.

**Deliberately not fixed here.** The fix is an explicit `manualChunks` vendor group: a build-config change with its own measurement and its own revert story, and folding a build experiment into a version bump makes both harder to undo. Recorded on #2843, whose lane this is.

Merging on the strength of: all three required checks green on the exact head, budget under the FAIL ceiling, and no peer-resolution flags anywhere in the diff.

Issue #2690, opened 2026-08-25
factionRoleVars answers for all nine slugs: the last 69 arms, and the ADR


pixieofhugs:
`factionRoleVars` returns `{}` for `na`, Albescent, null and unknown slugs. That is
deliberate and documented — `utils/factionRoles.ts` states it as an acceptance criterion
inherited from #2361:

> AN UNAFFILIATED VIEWER IS PIXEL-IDENTICAL BY CONSTRUCTION. For `na`, for `albescent`,
> for null and for a slug the server invents tomorrow this returns `{}` — not one property
> is declared, so every `var(--x, <today's token>)` at every read site is the value that
> already shipped, byte for byte. There is no second render path to keep in step and
> nothing to measure twice.

This change reverses it: the guarantee moves from **pixel-identical by construction** to
**pixel-identical by proof**. Owner call, taken 2026-08-25.

**Blocked on the predecessor issue.** Its green rule *is* the proof; this cannot land first.

## Why the criterion is already half-gone

The singular `factionRoleVar` answers for `na` and Albescent today, deliberately —
`utils/factionRoles.ts:257` says so, on the grounds that "a single role read has no
all-or-nothing seam to protect". And the contrast gate already measures both:
`ROLE_SLUGS` is nine slugs including `na` and `albescent`, asserted at
`utils/__tests__/factionContrast.test.ts:2617`. So "nothing to measure twice" is not the
case — the measurement exists and passes. What the criterion still buys is the structural
guarantee, which is what this change replaces with a mechanical one.

## What makes it a no-op

After the predecessor lands, 69 arms remain across 13 files, and each is pinned by the
derived rule to `factionRoleVar(slug, role, ground)`. Green rule means: at every site, the
value the map would emit is already the value the arm names.

Two mechanisms could have broken that, both checked on `main` and ruled out:

- **No prefix collision.** None of the 14 prefixes in play (`na-task-card`,
  `na-praxis-detail`, `na-profile-body`, `na-edit-praxis`, `na-field-desk`, `na-avatar`,
  `na-band`, `alb-band`, `na-seal`, `na-praxis-card`, `na-score-stamp`, `na-select-card`,
  `na-task-detail`, `rail`) is declared anywhere in `index.css`, so the spread cannot
  override a host channel the way `--snd-praxis-*` would.
- **No portal.** None of the files carries one, so no read escapes the subtree the spread
  declares on.

## What to do

[...]

Issue #2669, opened 2026-08-25
S.N.I.D.E.'s chrome ground leaves `accent` behind on a wall it no longer matches: 1.03:1 in light


pixieofhugs:
Found by the role-contrast loop on landing (#2661) — **the first thing it caught that a human
review had not.**

## The defect

S.N.I.D.E.'s `chrome` ground map (`frontend/src/utils/factionRoles.ts`) moves six roles onto the
wall family and leaves **`accent`** inheriting the sheet's `--faction-snide-card-accent`.

That token is `#b6ff2e` — the acid — and it is declared **invariant across both cascades**
(`index.css:514` and `:3219`, same value). It was measured for the near-black photocopier slab.
The wall flips: in light it is `--faction-snide-wall` `#efece3`.

**`#b6ff2e` on `#efece3` is 1.03:1.** Effectively invisible.

## It is the same shape as a bug already fixed

#2659's review caught `onFill` being left behind when `fill` moved to `-wall-credit` (2.07:1), and
#2663 made that specific pair **unrepresentable in the type** — `GroundOverride` no longer compiles
if a ground overrides `fill` without `onFill`.

`accent` is the same class and the type cannot catch it: the type can force one *pair* to travel
together, but it cannot know that moving `paper` invalidates the measurement of **every** ink on
it. That is a ratio, so it belongs to the loop — which is precisely how it was found.

## Likely satisfiable, and the candidate is named

`--faction-snide-wall-credit` is the rung of the same hue that `fill` already moved to on this
ground: **6.28:1 light / 9.22:1 dark** as a bare ink on the wall, already gated by
`SNIDE_WALL_INKS` (#2343, "the marker scrawl, bare").

#2173 already ruled that `--faction-snide-acid` never touches paper, which points the same way.

**Choosing it is a design call, so it is filed rather than taken.** Confirm it reads right as an
accent — a decorative/meta ink — and not merely that it passes.

## Scope

1. Give `GROUND_OVERRIDES.chrome.snide` an `accent` entry, or rule that it should keep the sheet's.
2. Delete the `"light | snide/chrome accent on paper"` entry from `BASELINE` in
   `factionContrast.test.ts` **only** once the pair actually clears — the ratchet fails if an
   allow-listed pair starts passing, which is the prompt to remove it.
3. Consider whether the "moving `paper` invalidates every ink on it" rule wants stating where the
   `GroundOverride` type's own docblock already argues the `fill`/`onFill` case.

[...]

Issue #2664, opened 2026-08-25
Snide's take-the-tie ability is a slug branch on both sides of the wire, and they must move together


pixieofhugs:
Found while writing the behaviour law in #2660. The law says: **branch on a slug for paint and
tree only — never for behaviour, capability or copy.** Snide's take-the-tie duel ability is
computed from a slug on **both** sides of the wire.

## Verified on `origin/main`, 2026-08-25

**Backend** — `backend/services/scoring.py:160`:

```python
def snide_tie_winner_slug(faction_a: str, faction_b: str) -> Optional[str]:
    """... Snide's ability is that it takes ties — but only when it is the *sole* Snide side."""
    a_snide = faction_a == SNIDE_FACTION_SLUG
    b_snide = faction_b == SNIDE_FACTION_SLUG
```

Three services read it.

**Frontend** — `frontend/src/components/duel/shared.tsx:188`, in `useDuelStakes`:

```ts
const exactlyOneSnide =
  (viewerFactionSlug === SNIDE_FACTION_SLUG) !== (opponentFactionSlug === SNIDE_FACTION_SLUG)
```

That is the same rule, re-derived client-side, and it is **shown to the player as their stakes**.

## Why this is one issue and not two

The two halves are **consistent with each other today**. Fixing one alone would make the stakes a
player is shown disagree with the points the server actually banks — a worse bug than the one
being fixed. They move together or not at all.

## The shape

`FactionConfig` in `backend/game_config.py` is already a list of exactly this kind of flag —
`level_jump_reach`, `can_hold_multiple_memberships`, `habit_bonus_points`,
`can_apply_metatask_at_any_level` — each defaulted so an era file states only what is non-baseline,
each with a docstring saying *"never branch on a faction slug in a service"*. This ability is the
one that never got the treatment.

Something like `takes_duel_ties: bool = False`, set on S.N.I.D.E. in `backend/eras/era_1.py`. Then:

- `snide_tie_winner_slug` becomes a rule about *which side has the flag*, not which side is Snide,
  and probably loses "snide" from its name.
- The frontend reads the flag off `FactionConfigOut` the way `reads_the_array` is already read
  (`components/TheArray.tsx:135`) — note the duel code already has `config.factions.find(...)` in
  hand at the point of the branch, so the config is right there.
- If you emit a new field you MUST commit the regenerated `backend/openapi.json` **and**
[...]

Issue #2661, opened 2026-08-25
The contrast gate becomes a loop: roles x factions x grounds x themes, instead of a hand-curated pair list


pixieofhugs:
Batch 01 item 2 of the faction-reuse plan ("Nine Kits, One Vocabulary" — the rewrite recorded on
#2649, not that issue's body).

**Blocked by the role-vocabulary issue** — there is nothing to enumerate until `FactionRoles`
exists. Do not start this one first.

## Decision 08 of the plan

> The contrast gate becomes a loop. Roles × factions × grounds × themes is finite.

## What the gate is today

`frontend/src/utils/__tests__/factionContrast.test.ts` reads `index.css` **as text** and measures
a **hand-curated list** of `{ what, surface, text }` literals. Its own docblock is honest about
the half it cannot do — it catches a token whose declared colour fails AA against the surface it
is *documented* to sit on, and it cannot catch a pairing, which only exists once rendered
(`e2e/contrast.spec.ts` is part B).

The consequence is the thing this issue exists to kill: **a new token family is unmeasured while
green.** Nobody adds the rows by hand, so the gate silently stops covering the newest work — and
"the contrast gate covers it" has been asserted in this repo and been false.

## The change

Enumerate `roles × factions × grounds × themes` instead of scanning text against literals. Once a
faction supplies a role map, the gate can walk it: every role, every faction, every ground that
faction declares an override for, both cascades.

**This will fail on pairings nobody has ever measured. Those failures are findings, not a broken
build.** Half of this issue's deliverable is the list. Expect the S.N.I.D.E. acid family in
particular: `--faction-snide-acid-deep` reads **2.30:1** on the light wall, which is why #2343
had to mint `-wall-credit` (7.72:1 / 11.35:1) as a second name rather than repoint the first.

## Deliverables

1. The loop, replacing the curated pair list for everything the role maps cover.
2. **The findings list** — every pairing the loop fails that the old gate never measured, with its
   ratio. File the real ones; do not fix them here and do not weaken a threshold to go green.
3. A statement of what the loop still cannot see, so the next person does not over-trust it the
   way the curated list was over-trusted.

## Traps, verified

- A token block written `:root,\n[data-theme] {` is **invisible to `cssVars.ts`**, which matches
  `:root, [data-theme]` on ONE line. That also makes `frozenThemeAliases` pass vacuously. If the
  role work introduces such a block, the loop inherits the blindness.
[...]

PR #2929, opened 2026-08-30
Batch omega: ADR-0090 records the classification procedure, SPEC §4 becomes the faction checklist, and Albescent's CSS share gets a named allowance

### LINKED ISSUE #2719: Batch omega: write down the procedure - the ADR, the what-a-faction-needs checklist, and Albescent's named CSS allowance
Batch Ω of #2649, under the **"Nine Kits, One Vocabulary"** plan (the 2026-08-25 comment, which supersedes that epic's body). The nine faction lanes are merged and the vocabulary is live; this is the batch that writes down what they settled, so the next person does not re-derive it.

Three artefacts. They are independent and can land in one PR.

## 1. The ADR — a procedure, not a rule

Decision 13 is explicit: *"The ADR records a procedure, not a rule. A rule mispredicted five of six once already."* So this must not be written as "chrome resolves from tokens, ornament is drawn by an archetype" — that was the original epic's rule, and #2650's kill-test found five of six cases held by the **tree**, not the paint. That finding is the reason the epic was rewritten.

What the ADR should record instead is the **classification procedure**: four buckets — paint · tree · behaviour · content — and how to decide which bucket a per-faction difference is in before designing anything for it. The thirteen settled decisions in the plan comment are the raw material; several are already load-bearing elsewhere and should be cited rather than restated:

- **Behaviour is never slug-keyed** — abilities belong to eras and move between factions, mirroring `game_config.py`. Already has a live violation history (#2660, #2664).
- **A faction supplies a map, not values** — eight roles, not fifty CSS slots, zero new declarations.
- **A travelling piece owns no outer margin** — it owns its box, the host owns every gap (#2655).
- **The contrast gate is a loop**, not a hand-curated pair list (#2661) — and the loop caught what review missed at 1.03:1 (#2669).
- **Convergence is bounded**: the invariant layer may converge (order, rhythm, radius, busy and error states, a11y); the identity layer may not (hue, typeface, ground, ornament, copy voice).

One hazard the lanes produced that belongs in the procedure: **a role left behind when its ground moved**, twice — `onFill` at 2.07:1 and `accent` at 1.03:1. A type can force one pair to travel together but cannot know that moving `paper` invalidates the measurement of every ink on it. That is a ratio, so the loop owns it, and every lane that adds a ground override will meet it.

**ADR number:** the next free number was **0088** when this was filed (highest on disk: `0087-structural-faction-slugs-are-not-era-owned.md`). That is a snapshot, not a reservation — **run `ls docs/adr/` before writing and take the next free number then.** Follow #2536's status vocabulary; do not rewrite or delete existing ADR bodies.

## 2. The "what a faction needs" checklist

A tenth faction should be a documentation goal, not an architecture goal (decision 3). The checklist is what makes that true: the role map it must supply, which surfaces register an archetype and which get Default, what the contrast loop will measure about it, and what it does *not* need to author. Scope is the twenty surfaces of decision 10 — the full manifest plus the census of faction-dressed surfaces that were never registered.

Acceptance: someone can add a faction by following it without reading this epic.

## 3. Albescent's named allowance in `bundle-budget.mjs`

**Verified on `origin/main`: 44 of 205 top-level class rules in `index.css` are `.alb*`** — more than the other seven factions combined.

The plan's own words: *"the asymmetry is right, the silence is not."* Albescent is one ornament vocabulary over na (ADR-0083) and legitimately carries more class rules than a faction that is mostly tokens. What is wrong is that nobody decided it — the number is just what accumulated. Give it a **named allowance** in `scripts/bundle-budget.mjs` so the asymmetry is an explicit, reviewed number that fails CI when it grows, rather than an unremarked fact.

## Out of scope

- The six surface lanes (Batches 11+). The first is filed separately; the rest follow it one at a time.
- Rewriting #2649's body. It is superseded and its comment says so — that is the historical record, and the same rule that protects old ADRs protects it.
- Any appearance change.

pixieofhugs:
Closes #2719

Batch Ω of #2649, under the *"Nine Kits, One Vocabulary"* plan (the 2026-08-25 comment, which supersedes that epic's body). The nine faction lanes are merged; this writes down what they settled.

Three artefacts, two commits.

---

## 1. ADR-0090 — a procedure, not a rule

`docs/adr/0090-a-per-faction-difference-is-classified-before-it-is-designed.md`

Decision 13 is explicit that this records a **procedure**, so it deliberately does **not** restate the killed rule (*"chrome resolves from tokens, ornament is drawn by an archetype"*). #2650's kill-test found five of six survivors held by the **tree**, not the paint.

What it records instead:

- **The four buckets** — paint · tree · behaviour · content.
- **The order the question is asked**, which is the load-bearing part: behaviour first (the bucket with a live violation history, and where a slug branch freezes a rule the next era cannot re-tune), then content, then paint, and **tree as the residue** — the expensive bucket has to be *earned* by eliminating the other three rather than reached by not asking.
- **What classifying obliges**: state the split in the issue before designing; never classify from a literal count; a forked family is a failure state; payload has a veto.
- **The ratio is in no bucket**, and gets its own section. `GroundOverride` can force `fill`/`onFill` to travel together, but no type can know that moving `paper` invalidates every ink measured on it — hence `onFill` at 2.07:1 and `accent` at 1.03:1, and hence the gate being a loop.

The record's central argument is about **censuses**, not factions: three measurements in this epic could not tell the thing being counted from a thing that looks like it — the `style={{` count, the score stamp's *"nine stamps declare their own outer spacing"* (none did; two **hosts** declared no gap), and `pressGrounds` reporting two grounds instead of six, green.

Load-bearing decisions are **cited, not restated**: #2655, #2659, #2660, #2661, #2664, #2669, plus ADR-0016/0031/0038/0042/0083/0085. No existing ADR body was touched; #2649's body was not rewritten. `docs/adr/README.md` was regenerated with `python scripts/adr_index.py`, never hand-edited.

**ADR number:** took **0090** as assigned at dispatch, not "the next free number". 0089 is #2690's and is absent from the regenerated index — expected, and the index regenerates rather than resolves by hand if the two collide on rebase.

## 2. The checklist — `SPEC-faction-ui-profile.md` §4, extended

**This is the "does it need to exist at all?" call, and it is the main judgement in the PR.** §4 *is already* the new-faction registration checklist, and its own change log records that it was pruned in 2026-07-18 specifically to stop a prose registry rotting. A second, rival checklist would have rotted the same way — so §4 was extended in place rather than duplicated into a new doc.

Added:

- **Items 12–14, the role vocabulary** — and the headline is that a new faction *normally edits `factionRoles.ts` not at all*: registering the slug in `CSS_KEY` beside a `-card-*` block is the whole of joining the vocabulary. The one exception is a faction whose chrome stands on a ground that is not its card (S.N.I.D.E.'s wall is the only one).
- **Item 11 extended in place** with what the contrast loop now enumerates for you, and the five things its docblock says it still cannot see.
- **"What a faction does not author"** — undeclared surfaces, outer margins on travelling pieces, contrast rows, role-map declarations, behaviour, copy.
- **The census of faction-dressed surfaces that are not manifest surfaces**, labelled as a census rather than a registry so it is not mistaken for something that stays current. All six paths verified on disk.

Items 1–11 keep their numbers: other files cite them by number, and one such citation (`factionContrast.test.ts`'s "§4 item 15") is *already* stale.

**`SURFACE_KEYS` is 22, not the plan's twenty.** Verified against `src/factions/manifest.ts`. `docs/kit-structure.md` said twenty/21 in three live places, so it is corrected — otherwise the two docs the routing table now links together would contradict each other. One routing-table row added to `CLAUDE.md`, since `SPEC-faction-ui-profile.md` was not reachable from it at all.

## 3. Albescent's named allowance — `bundle-budget.mjs`

**The figure was re-measured, and it does not reproduce as stated.**

`index.css` became an import map in #2924, so the sheet is assembled through it the way `src/test/indexCss.ts` does. The issue's **44 of 205** reproduces as **44 of 206** at `c5ff48cb` — but only under the narrow reading that counts depth-0 rules and treats `@layer`/`@media` as opaque. **That reading fails open**: wrapping a new `.alb` rule in `@media` hides it from the count without removing one selector from the sheet.

So grouping at-rules are transparent here, and the numbers are:

| reading | at `c5ff48cb` (filing) | today |
|---|---|---|
| depth-0 only (the issue's) | 44 / 206 | 47 / 215 |
| **at-rules transparent (used here)** | 47 / 319 | **50 / 330** |

`ALBESCENT_ALLOWANCE = 50`, exceeded → exit 1. **The two existing budget lines are untouched** (#2469 / the standing #2843 ruling).

One honest correction to the issue's supporting claim: *"more than the other seven factions combined"* holds under the narrow reading (47 vs 42) but **not** under the whole-sheet reading (50 vs 66). The allowance does not depend on that claim — it pins Albescent's own number — but the claim should not be repeated as-is.

### The seam, and the test at it

**The seam is "does the budget script see the same stylesheet `readIndexCss()` sees?"** — because that is the only way this guard goes wrong, and it goes wrong *silently*. The script is plain Node and cannot import the TypeScript helper, so it re-implements the import-map read; a script that assembles half the parts reports a smaller Albescent count, which reads as a comfortable pass.

`frontend/src/__tests__/albescentCssAllowance.test.ts` re-derives the count from `readIndexCss()` with **its own walk** and asserts it equals the literal in the script. It deliberately does *not* import the script's parser — that would only prove the script agrees with itself. Two independent readings of one sheet agreeing on 50 is the assertion worth making.

Red proved before green, three ways:

1. Appending `.alb-mutation-probe { }` → `FAIL ALB 51 allow>50`, exit 1.
2. The same rule wrapped in `@media (min-width: 900px)` → **still** `FAIL ALB 51`. This is the fail-open the narrow definition would have allowed.
3. Drifting the literal to 49 → the vitest cross-check goes red (`expected 50 to be 49`).

The count is 330 on both a React 18 and a React 19 build, which is expected: it is a source measurement and no dependency tree can move it.

---

## Verification

Every step in `.github/workflows/test.yml`, read from the file rather than from memory:

- `npm run lint` — clean
- `npm run typecheck` — clean
- `npm run typecheck:design-sync` — clean
- `npm test` — **480 files, 9,959 passed, 1 skipped**
- `npm run build` — clean
- `npm run budget` — `ok CSS`, `ok ALB 50 allow>50 of 330`
- `pytest tests/test_adr_index.py` — 4 passed (the job that guards the regenerated index)

`api-schema` not run: no route signature, `response_model`, or Pydantic schema is touched — this PR contains no backend source change at all.

**Pre-existing and not mine:** `WARN JS 151.2 KB`. React 19 dissolved the react-dom vendor chunk; the warn/fail lines are ruled immovable and were not moved.

**Environment note:** an early `tsc` run through a junction to the main checkout's `node_modules` reported a phantom `TS2322` in `PortraitPicker.tsx` — main's tree is stale at React 18.3.1 against a `^19.2.8` `package.json`. The junction was removed with `cmd rmdir` (not `rm -rf`) and replaced with a real `npm ci` in this worktree; React resolves to 19.2.8 and `tsc` is clean. Every number above was measured after that.

---

## What to eyeball

**Nothing. There is no visual QA to do on this PR, and that is verifiable rather than asserted.**

- Two of the three artefacts are Markdown (`docs/adr/`, `docs/spec/`, `docs/kit-structure.md`, `CLAUDE.md`) and ship no bytes to the browser.
- The third is a **CI script and its test**. `frontend/scripts/bundle-budget.mjs` is not imported by the app; it runs from `npm run budget` after the build.
- **No stylesheet, component, or token was edited.** The only file under `frontend/src/` is a new `__tests__` file. The `.alb-mutation-probe` rules used to prove the guard red were reverted with `git checkout` before the first commit, and `git status` is clean.
- The built CSS asset is unchanged at 21.9 KB gzipped, `ok CSS`.

If you would rather spot-check one thing, make it the **judgement, not the pixels**: that `ALBESCENT_ALLOWANCE = 50` is a number you are willing to have block a future Albescent lane until someone raises it with a ledger line. That is the whole point of the artefact — *"the asymmetry is right, the silence is not"* — and it is the one part of this PR that is a ruling rather than a measurement.

---

## Rebase onto `876b3740` — the allowance did NOT move

Rebased after PR #2931 (#2690's work, ADR-0089) merged. `docs/adr/README.md` was the only conflicting path; the other six files applied clean.

**The index was regenerated, not hand-resolved.** Took one side to unblock the rebase, then ran `python scripts/adr_index.py` and committed its output — 90 records, listing both `0089-the-role-map-answers-for-every-slug-and-the-rail-keeps-its-guard` and `0090-a-per-faction-difference-is-classified-before-it-is-designed`. ADR **0090** is unchanged; nothing was renumbered.

**Re-measured the allowance against the rebased tree, and it is unchanged at `50 of 330`.** That is a verified result rather than an assumption — two merges had landed since the original measurement:

- **#2928** deleted five `--eph-*` declarations. Those are custom properties *inside* existing `:root` / `[data-theme="dark"]` blocks. `git diff` over `src/css/` between the old base and `origin/main` adds and removes **zero** lines containing a brace, so no rule was opened or closed and no class selector could move. `.eph` selectors are still 21.
- **#2931** armed twelve `Default*` surfaces. That is TSX, not stylesheet.

This is the pin behaving as designed: it counts **selectors**, so a change that only edits declarations inside existing rules correctly moves it by nothing. A byte-based pin would have drifted on both merges.

Re-ran the whole `frontend` job on the rebased tree: lint clean, `typecheck` clean, `typecheck:design-sync` clean, `npm test` **484 files / 9,947 passed / 1 skipped**, build clean, `npm run budget` → `ok CSS`, `ok ALB 50 allow>50 of 330`, exit 0. `pytest tests/test_adr_index.py` → 4 passed. `WARN JS 151.2 KB` remains the pre-existing React-19 chunk warning; the warn/fail lines are still untouched.


pixieofhugs:
## Review fix — three dead pointers, one of them in the ADR itself

Both flagged references in `bundle-budget.mjs` are corrected to the real path, `src/__tests__/albescentCssAllowance.test.ts`. They were wrong in *two* ways at once — wrong directory and wrong filename — because I wrote the ledger intending `src/css/__tests__/`, then put the test beside `motionSplit.test.ts` and never updated the prose. `src/css/__tests__/` does not exist at all.

**Auditing the rest of the diff caught a third, and it was worse.** ADR-0090 cited `familyOf()` as the migration gate's fix. **That symbol does not exist anywhere in the frontend.** I took it from #2649's comment (*"Lane 02 added `familyOf()`"*) and cited an issue-comment claim as a verdict — which is precisely the census failure the ADR is about. Caught by my own prose, which is the least comfortable way to find it.

The *mechanism* is real, so the citation now names what is actually on disk:

- **`factionRoleFallbacks.test.ts`** — its `stragglers` sweep derives the token family from `factionCssVar(surface.slug)` rather than from the slug, which is what stops it passing vacuously for `na` and `albescent` (family `default`).
- **`singularityRoleReads.test.ts`** — its `CARVED_OUT` list is the non-vacuity half, asserting each file *"still names this faction directly"*. `components/sigil/FactionSigil.tsx` is on it. This half was accurate; it just wasn't attributed.

### Audit result

All **29** path-like citations across `CLAUDE.md`, `docs/kit-structure.md`, `SPEC-faction-ui-profile.md`, ADR-0090 and `bundle-budget.mjs` now resolve on the merged tree — 21 as explicit paths, 8 as bare basenames that each resolve to exactly one tracked file. Symbols spot-checked too (`GROUND_OVERRIDES`, `SURFACE_KEYS`, `CSS_KEY`, `factionRoleVar(s)`, `pressGrounds`, `FactionConfig`, `.na-backdrop`).

### Note on the push

The remote branch had moved: `main` was merged into it as `75a24f86`, bringing in #2935. I rebased onto that rather than force-pushing over it, so the merge commit is preserved and the push was a fast-forward.

That merge also invalidated the "nothing has merged since" premise for skipping the re-measurement, so I re-checked: `git diff` over `src/css/` and `index.css` between `674d0e7c` and HEAD is **empty** — #2935 touched no stylesheet. The allowance is unchanged at **50 of 330**.

### Verification on the merged tree

lint clean · `typecheck` clean · `typecheck:design-sync` clean · `npm test` **484 files / 9,973 passed / 1 skipped** · build clean · `npm run budget` → `ok CSS`, `ok ALB 50 allow>50 of 330`, exit 0 · `pytest tests/test_adr_index.py` 4 passed.

Only comments and prose changed in this commit; no executable line was touched.

Issue #3006, opened 2026-09-02
The agent docs still describe a Windows machine: sweep the stale environment notes


pixieofhugs:
> *This was generated by AI during `/builder-bot`, 2026-09-02.*

This machine moved to Linux (Arch, Omarchy). Several agent-facing docs still
give Windows-only instructions, and one brand-new hazard appeared with the move.
Two of these were fixed in flight because the batch could not run without them;
the rest are a sweep for tomorrow.

## Already fixed in flight (this batch)

- `.claude/skills/builder-bot/SKILL.md` hazards block — `mklink` -> `ln -s`, plus
  the new `python3`/`gh` hazard below.
- `.claude/agents/frontend-feature.md` — same `mklink` -> `ln -s` swap.

## The new hazard, for the record

`python3` on this machine is a **mise shim** (`~/.local/share/mise/shims/python3`
-> `/usr/bin/mise`) that re-execs and **rewrites `PATH`**, putting
`~/.local/bin` first. `~/.local/bin/gh` is a mise *bootstrap* script:

    #!/bin/bash
    export MISE_MINIMUM_RELEASE_AGE=0
    mise use -g --quiet "gh" || exit 1
    exec mise x "gh" -- "gh" "$@"

It **hangs forever** — even `gh --version`. So any repo Python script that shells
out to `gh` hangs when run as `python3`, while the same `gh` command typed at a
Bash prompt works, because the login shell's profile puts the real binary first.

Confirmed affected: `scripts/gh_issue_comments.py` (hung at 120s; this is the
issue-reading wrapper every batch skill depends on).

**Workaround in use:** run those scripts as `/usr/bin/python3`.

**Worth deciding:** whether the real fix is repairing/removing the broken
`~/.local/bin/gh` shim (outside the repo, so it needs the owner) rather than
teaching every caller to dodge it. A machine-level fix would let the docs stay
silent about it.

## Still stale — the sweep

| file | what is wrong |
|---|---|
| `.claude/skills/git-reaper/SKILL.md` | `cmd //c rmdir` for junction teardown (~line 123); a "Windows Temp cleanup" gutted-worktree heuristic (~line 55); a "Windows paths in non-raw Python strings" hazard (~line 184). The junction hazard is real on Linux too, as a symlink — it needs restating, not deleting. |
[...]

PR #3007, opened 2026-09-02
Agent docs: the machine is Linux now, and python3 is a mise shim

pixieofhugs:
Part of #3006. Two agent-facing notes assumed Windows, and one new hazard
appeared with the OS move. Fixed in flight because tonight's `/builder-bot`
batch could not dispatch a frontend agent without them.

## What changed

**`.claude/skills/builder-bot/SKILL.md`** (hazards block, pasted verbatim into
every subagent prompt) and **`.claude/agents/frontend-feature.md`**:

- `node_modules` in a worktree: `cmd //c mklink /J` → `ln -s` with an absolute
  target. The false-green warning is kept, because a wrong target still hides
  the `tsc` binary and `tsc --noEmit | grep -c error` then reports a **false**
  0-error pass.
- The teardown warning is sharpened rather than dropped: `rm -r` follows a
  symlink into the main checkout's `node_modules` the same way
  `git worktree remove` followed a junction. `rm frontend/node_modules`, never
  `rm -r`.

**New hazard 2b**, in the hazards block only:

`python3` on this machine is a mise shim (`~/.local/share/mise/shims/python3`
→ `/usr/bin/mise`) that re-execs and **rewrites `PATH`**, putting
`~/.local/bin` first. `~/.local/bin/gh` is a mise *bootstrap* script that
hangs forever — even `gh --version`. So any repo Python script that shells out
to `gh` hangs when run as `python3`, while the same `gh` command typed at a
Bash prompt works, because the login shell's profile puts the real binary
first.

Two confirmed casualties tonight:

- `scripts/gh_issue_comments.py` — the issue-reading wrapper every batch skill
  depends on. Hung at 120s.
- `scripts/hooks/guard_pr_merge.py` — invoked as bare `python` by
  `.claude/settings.json`. It denies when it cannot answer, by design, so it
  denied **every** merge tonight.

The fix in this PR is documentation only: tell agents to use `/usr/bin/python3`
for those scripts. **The settings.json half is deliberately not in this PR** —
see #3006 for why, and for the machine-level fix that would let the docs stay
silent about this entirely.

## Verification

Docs only, no code. Verified by running the wrapper both ways:

    python3   scripts/gh_issue_comments.py 2990   # hangs, exit 124 at 120s
    /usr/bin/python3 scripts/gh_issue_comments.py 2990   # prints the issue

and the merge guard both ways, where the system interpreter reaches the guard's
real logic instead of timing out on `gh repo view`.

Not covered: `.claude/skills/git-reaper/SKILL.md` and
`docs/agents/running-locally.md` are still Windows-flavoured. They are itemised
in #3006 rather than swept here, because neither was in tonight's path and a
blind sweep of the git-reaper junction hazard risks deleting a warning that is
still real on Linux in symlink form.

## What to eyeball

Nothing rendered. Worth a human check that the `ln -s` line's hardcoded
`/home/pixie/Projects/WorldZeroPlayground` matches where you actually keep the
main checkout — #3006 flags deriving it instead.

### DIFF
--- a/.claude/agents/frontend-feature.md
+++ b/.claude/agents/frontend-feature.md
@@ -18,18 +18,20 @@ If a change needs real styling/design decisions beyond "reuse existing component
 
 - Edit and create under `frontend/src/` — pages, components, api clients, hooks, context, routes.
 - Run `npm` commands via Bash. In a fresh worktree `frontend/node_modules` does not
-  exist. Either run `npm ci`, or junction it from the main checkout. Note `mklink` is a
-  **cmd.exe builtin and is NOT on PATH** in Bash or PowerShell, so it needs the call
-  through `cmd`:
-  - Bash: `cmd //c mklink /J "$(pwd)/frontend/node_modules" "<absolute Windows path to main>\frontend\node_modules"`
-  - PowerShell: `New-Item -ItemType Junction -Path .\frontend\node_modules -Target <absolute path to main>\frontend\node_modules`
+  exist. This machine is **Linux** — either run `npm ci`, or symlink it from the main
+  checkout with an **absolute** target:
 
-  A junction pointing at the wrong target leaves the `tsc` binary absent, and
+      ln -s /home/pixie/Projects/WorldZeroPlayground/frontend/node_modules "$(pwd)/frontend/node_modules"
+
+  A link pointing at the wrong target leaves the `tsc` binary absent, and
   `tsc --noEmit | grep -c error` then reports a **false** 0-error pass. Confirm
   `tsc --version` actually runs before believing a green typecheck.
 
+  Remove the link before you finish, with `rm frontend/node_modules` — never `rm -r`,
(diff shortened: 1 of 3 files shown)


pixieofhugs:

```
# Review — PR #3007 (agent docs: the machine is Linux, python3 is a mise shim)
Reviewed 2026-09-04 against merge-base with main. Docs + .gitignore only; no code.

no findings

Nits (not blocking):
- The `rm -r` warning is conservative rather than measured: on GNU coreutils `rm -r frontend/node_modules` removes the link and `rm -r frontend/node_modules/` (trailing slash) fails with "Not a directory". Neither empties main's copy. Keeping the warning costs nothing.
- The `.gitignore` change (`frontend/node_modules/` -> `frontend/node_modules`) is correct: the trailing-slash form matches only directories, and a symlink is not one to git.
- PR is BEHIND main; `gh pr update-branch` before merge.
```

Issue #2993, opened 2026-09-01
Propose a Task: na is bespoke where seven kits are on the chassis, and factionSurfaces.ts dies with the rebuild


pixieofhugs:
**Owner ruling, 2026-09-01, during `/bug-bot`:**

> "I do want to, in general, want to have all factions read from the same chassis. The goal of my previous ruleing was to make sure each faction kept a distict look. This was before we had done all the architectural work to set things up so the devide between what is shared and what is faction spesific was more ambiguous. Now i Am concerned about the uniformity of information. The 7 factions which use the chasis all do have distinct looks and look very good, while being consistant readable experinces. I want to bring NA and albecent into that, with the idea that the albecent layer still has moving colors etc."

The third of three surfaces where na and Albescent sat out the chassis. The character forms are the sibling lanes — see the cross-link comment. **This one is genuinely different from those two and the differences are the whole brief; do not copy the character-form PR.**

## The symptom

Seven faction kits draw Propose a Task as a framed composer sheet. na draws a bespoke page: a `Breadcrumb`, a `PageTitle`, and a stack of hand-authored inline styles with no sheet, no masthead and no shared footer. Albescent renders na unchanged, so it has the same page.

Nobody reported this from a screenshot — it was found by censusing the surface that *was* reported (create character) for the same cause, and finding it on three pages instead of one.

## The cause

`DefaultProposeTask.tsx` (636 lines) mounts none of the chassis. It imports four inline-style builders from `pages/proposeTask/factionSurfaces.ts` — `proposeCardStyle`, `taskNameInputStyle`, `metaBoxStyle`, `submitButtonStyle` — each of which computes a style from a slug at runtime. The seven faction kits mount `ComposerPage` / `ComposerSheet` / `ComposerSection` / `ComposerFooter` out of `pages/editPraxis/archetypes/shared.tsx` instead.

The na dress for the chassis already exists and is shipped — `DefaultEditPraxis.tsx:167-237` runs it out of `--faction-default-composer-*` / `-card-bg` / `-card-alarm` / `-border` / `-on-accent` / `-aurora*`.

## The ruling, and exactly what it does not reverse

**One file per faction still stands.** What is forbidden is a single component with a runtime skin table rendering nine trees. ADR-0065 §"What this ADR does not do" drew that line already; ADR-0090's §Tree sentence carries the licensing clause that gets dropped when it is quoted:

[...]

Issue #2987, opened 2026-09-01
WoW's edit-slot plate reads 1.06:1 and its hairline 1.41, and only the TEXT rows are guarded


pixieofhugs:
> *Filed by AI during `/builder-bot` on 2026-09-01, out of the review of #2956 (PR #2986).*

**`wowEditCharacterContrast.test.ts` measures four TEXT rows and all clear AA. It does not measure the non-text pair on the same sheet, and that pair does not clear.**

## The shortfall

The edit-character slots' plate on WoW's card ground:

| pair | measured | 1.4.11 needs |
|---|---|---|
| `--color-bg-surface-alt` on `--faction-wow-card-bg` | **1.06:1** | 3:1 |
| the plate's hairline | **1.41:1** / **1.56:1** | 3:1 |

WCAG 1.4.11 (non-text contrast) asks 3:1 for a graphical boundary that carries meaning — which a plate edge delimiting a control region does. A text-only contrast suite cannot see this, which is why it shipped green.

Reported by the WoW lane itself during #2537's fan-out, and re-verified by `/planner-bot` on #2956.

## Why it was not fixed in #2956

#2956 gave the shared `editCharacterSlots.tsx` a dress seam (PR #2986, merged) precisely so an archetype could repaint these without editing the shared file. But #2986's whole claim is that **no dress changed** — it is behaviour-preserving by construction, and its guard asserts a mount that passes nothing renders byte-identically. Fixing WoW's plate is a dress change with its own measurements, so it belongs in its own PR.

So this is now unblocked rather than blocked: the seam exists, and WoW can pass its own plate ink and edge.

## What to build

- Repaint the plate and its hairline on `WowEditCharacter.tsx` through the seam #2986 added (`FactionRow`'s `rowStyle` / `DeleteCharacter`'s `panelStyle`, or whichever props carry the ground-dependent half — read the file, the prop names are recorded in its header).
- **Extend `wowEditCharacterContrast.test.ts` to measure the non-text pair against 3:1**, not just the four text rows against AA. A fix without the row leaves the same blind spot for the next change.
- Do not touch the shared file. That is what the seam is for.

## Watch out

- The four existing text rows must still pass. Do not trade one for the other.
- `--color-danger` / `--color-on-danger` on the filled confirm are deliberately not faction-repointed (2026-08-27 ruling). Leave them.
[...]

Issue #2816, opened 2026-08-28
The select tiles assert the 360x300 box seven times and never for the Default tile that na and albescent wear


pixieofhugs:
**The select tiles have 55 assertions and not one of them is parameterised over the kits.** Two of
those assertions are faction-invariant, are written out by hand seven times each, and are absent for
the two kits a `Default`-backed tile serves.

> Filed under #2814. Unlike its siblings, this one is measured against **both** coverage layers —
> the per-kit suites and the derived loop suites. These cells are covered by neither.

### 1. `keeps the fluid 360x300 box the directory grid is built on (#732)`

Written out in **seven** files — `covenWearsTheSlipRegister`, `ephemeristsWearsThePlateRegister`,
`everymenWearsTheBillRegister`, `singularityWearsTheTerminalRegister`, `snideWearsTheClippingRegister`,
`uaWearsTheLeafRegister`, `wowWearsTheDecreeRegister` — with the same four expectations each:

```
expect(source).toContain('width: "100%"')
expect(source).toContain("maxWidth: 360")
expect(source).toContain("minHeight: 300")
expect(source, "a fixed height would break the phone column").not.toMatch(/\bheight: 300\b/)
```

**`DefaultSelectCard.tsx` carries `maxWidth: 360` and `minHeight: 300` at lines 120-121**, so the
invariant applies to it — and nothing asserts it. `DefaultSelectCard` serves **na and albescent**,
two of the nine. Its own comment says the 375px single-column mobile directory depends on all three
values, which makes the untested tile the one the phone column would break on.

### 2. `names no token family its own task card does not`

The same seven files, again absent for `DefaultSelectCard`. This is the register rule — a tile
borrows its task card's token family and mints nothing of its own — and it is the sort of rule that
is worth exactly as much as its weakest kit.

## What the loop layer already covers, and does not

`factionSelectCardDispatch.test.tsx` iterates a derived source and covers **dispatch**: `renders
every registered faction in its own costume`, `wires the na row itself to DefaultSelectCard`,
`renders an unknown slug through DefaultSelectCard`, `keeps the legacy slugs pointed at their live
archetype`.

That is structure — which component mounts. It says nothing about the box or the register. Do not
read the dispatch suite as covering these.

## Scope

Replace the seven hand-written copies of each invariant with one derived loop over
`surfaceMap('factionSelectCard')`, covering all nine kits including the `Default` tile.

[...]

Issue #2815, opened 2026-08-28
Twenty-six faction loops, five derived lists: a typed slug list cannot notice a tenth kit, and one already misses a ninth


pixieofhugs:
**Twenty-six test files already loop over the factions. Only about five derive the list they loop
over; the rest type the slugs out by hand.** A typed list is a loop that cannot notice a tenth kit —
and one of them already omits a ninth while claiming to be complete.

> Filed under #2814. This is the same defect as its siblings from the other end: #2801, #2812 and
> #2813 are surfaces with nine copies and no loop. This is the surfaces that *have* the loop and
> hand-maintain its range.

## The rule this issue enforces

**A loop's slug list is derived from the manifest, never typed.** Where a subset is genuinely
correct, it is derived *and then filtered*, with the exclusion named in code — never by writing out
the members that survive.

`components/avatar/__tests__/factionAvatar.test.tsx` is the model and already ships:

```
const SLUGS = ["na", ALBESCENT_FACTION_SLUG, ...FACTION_RAINBOW_ORDER];
```

A new kit joins that loop by existing. Nothing has to remember it.

## The hole this already caused

`pages/fieldDesk/__tests__/homeTrackBand.test.tsx` types eight slugs and comments:

> *"Every mobile home skin paints the same track; a per-faction skin that grew its own arithmetic is
> the defect one level up, so all eight are checked."*

**There are nine.** `AlbescentFieldDesk` is registered on `mobileFieldDesk` at
`factions/albescent.ts:363` and is not in the list. The comment asserts a completeness the list does
not have, which is worse than no comment — a reader who checks the claim stops there.

## Status of every faction loop, on `origin/main` at `4d3d5e61`

**Derived — a new kit joins for free:**

| file | list |
|---|---|
| `components/avatar/…/factionAvatar` | `["na", ALBESCENT_FACTION_SLUG, ...FACTION_RAINBOW_ORDER]` |
| `components/__tests__/cardHeadingOutline` | `['na', ...FACTION_MANIFESTS.map(m => m.slug)]` |
| `pages/__tests__/taskCardOffersTheTask` | `['na', ...FACTION_MANIFESTS.map(m => m.slug)]` |
| `components/duel/…/duelSealFormFactor` (`SEAL_SLUGS`) | `Object.keys(surfaceMap('duelSeal'))` |
[...]

PR #3008, opened 2026-09-02
proposeTask's audit re-scan: the busy CTA invariant seven kits wrote by hand and the Default did not (#2994)

### LINKED ISSUE #2994: proposeTask's audit clearance expired when #2538 landed: nine archetypes, and the disabledControlContrast sweep cannot reach any of them
> *Filed during `/planner-bot`, 2026-09-01, on the owner's ruling of the same day. Child of #2814.*

**#2814's 2026-08-31 ruling named this issue before it existed:**

> "`proposeTask` is the immediate case: #2538's seven-archetype fan-out is `ready-for-agent` and unbuilt, and `proposeTask` is also marked scanned-and-clean from 2026-08-28. **Its clearance expires the moment that fan-out lands.**"

**It landed.** #2538 closed 2026-09-01T00:13 and `origin/main` now carries nine archetypes under `frontend/src/pages/proposeTask/archetypes/` — Albescent, Coven, Default, Ephemerists, Everymen, Singularity, Snide, Ua, Wow. The audit table in #2814 still marks `proposeTask` scanned-and-clean on evidence dated **three days before any of them existed**.

## What is owed

Re-run #2814's own method over `proposeTask`, at the **duplication-scan** level it originally got or deeper: is there a faction-invariant assertion hand-repeated across the nine kit files, and does the surface have at least one derived loop suite that reaches all nine? Then update #2814's audit table with a **dated** verdict, per the ruling's own instruction to *"date every surface verdict, and re-scan a surface whose registration count has changed since."*

## What I already measured, and its limit

**This is a lead, not a proven defect. Do not brief it as settled.**

`disabledControlContrast.test.ts` is the guard #2955 made derived. Its walk is:

```ts
const SITES = sourceFiles({ dir: ARCHETYPE_DIR, match: /(?:Create|Edit)Character\.tsx$/ })
```

`ARCHETYPE_DIR` is under `pages/characterPaths/`. So the sweep **structurally cannot reach** `pages/proposeTask/archetypes/` — different directory, different filename pattern. And seven of the nine proposeTask archetypes hand-carry `control-off`:

| archetype | `control-off` occurrences |
|---|---|
| Coven, Ephemerists, Singularity | 2 |
| Everymen, Snide, Ua, Wow | 1 |
| Default, Albescent | 0 |

**But the defect class is probably different, and that is the part to check first.** Those controls are gated `disabled={submitting}` — a transient busy state — and `disabledControlContrast`'s own docblock says such states are *"deliberately not swept in here"*; it sweeps `disabled={!…}`, the control that is disabled when the page opens. Zero proposeTask archetypes use that form, and none spells `opacity: canSubmit`. Albescent draws nothing because it is a pure pass-through, which is the correct answer for a delegating wrapper (#2955 records the same shape).

So the honest position: **the sweep does not reach this surface, seven kits decided `.control-off` privately, and whether that is a hole or correctly out of scope has not been established.** Establish it. If it is out of scope, say so in `disabledControlContrast`'s docblock so the next audit does not re-open it.

## Watch out

- **The clearance may be expiring again while you work.** #2993 rebuilds `DefaultProposeTask.tsx` on the chassis and **deletes `pages/proposeTask/factionSurfaces.ts`**. Its own "Doc and test consequences" already requires *"a structure guard, derived not hand-listed: for every slug in `surfaceMap('proposeTask')` × both form factors, assert the field set and the action order."* If #2993 merges first, re-measure against its result rather than against today's tree — and check whether that guard already closes what you find here, in which case this issue's exit is a dated table update in #2814 and nothing more.
- **A guard that can scan nothing passes.** Floor whatever walk you write with a real number, and prove the extraction matches a known call site.
- **A typed slug list cannot notice a tenth kit** (#2815). Derive from `surfaceMap('proposeTask')`, never from a hand-typed array — that is the failure #2955 was open for.
- `uaProposeTaskContrast.test.ts` exists and covers UA only. It is not a loop layer.

## Done when

- [ ] `proposeTask` has a **dated** verdict in #2814's audit table, replacing the 2026-08-28 row
- [ ] The question above is answered either way, and if `.control-off` on `disabled={submitting}` is out of `disabledControlContrast`'s scope, its docblock says so
- [ ] Any hole found is either fixed with a derived guard or filed as its own child of #2814
- [ ] Whatever walk is written is floored, and the floor is a real number

## Related

- **#2814** — the epic, and the ruling that predicted this.
- **#2538** — the fan-out that expired the clearance. Closed 2026-09-01.
- **#2955** — made `disabledControlContrast`'s `SITES` derived. Its walk is the one that does not reach here.
- **#2993** — rebuilds `DefaultProposeTask.tsx` and asks for a derived structure guard on this exact surface. Sequence against it.

pixieofhugs:
Closes #2994

Re-scan of `proposeTask` for #2814, whose clearance expired when #2538 landed nine archetypes on
2026-09-01 against a scan dated 2026-08-28. The dated verdict is posted as a comment on #2814
(#2814's body is not edited).

**This started as an audit and found a real defect.** The issue's lead was that
`disabledControlContrast` cannot reach this surface; that is true, and checking it turned up an
invariant seven kits wrote by hand and the Default did not.

## The seam I tested at

**The rendered `type="submit"` control of each propose archetype, in the `submitting: true`
state, over a roster read from `surfaceMap('proposeTask')`.**

Not the source text. A source scan would have to find each kit by directory and filename, which
is the exact coupling that made the sibling guard blind to this surface in the first place.
Rendering also means the Albescent pass-through is measured through what it *draws* rather than
exempted for drawing nothing of its own — and it is one of the two kits that was wrong.

## The defect

#2486 ruled that a disabled primary control drops its CTA paint for the measured
`--control-off-fill` / `--control-off-ink` pair rather than fading, because `opacity` composites
the whole element: the fill sinks toward the sheet and the label's ink fades over the faded fill,
so the label loses contrast twice. That cannot vary by faction.

Rendered evidence, before the fix — every kit draws exactly one `type="submit"` button and all
nine are `disabled` while busy:

| kit | the busy CTA |
|---|---|
| coven, ephemerists, everymen, singularity, snide, ua, wow | `class="control-off"` (Singularity adds `sg-control-off`, Ephemerists `eph-cta`), no `opacity` |
| **na — `DefaultProposeTask`** | **no class, `cursor:wait;opacity:0.6`** |
| **albescent** | **byte-identical to the above** — pure pass-through |

`na`'s enabled paint is the spectrum *frame* — `linear-gradient(card-bg, card-bg) padding-box,
var(--faction-default-rainbow) border-box` — so the fade folded the ramp, the paper and the label
into one composite with no single ratio to measure. That is #2486's own "cannot be rescued by a
gentler alpha" case, and #2816's shape exactly: seven kits found the answer privately during
#2538's seven-PR fan-out, and the file they were all copied from kept the fade.

## Why neither coverage layer caught it (#2814's two-layer rule)

**Loop layer.** `disabledControlContrast.test.ts` owns this invariant and could not reach here
twice over:

1. `sourceFiles({ dir: ARCHETYPE_DIR, match: /(?:Create|Edit)Character\.tsx$/ })` with
   `ARCHETYPE_DIR` under `pages/characterPaths/` — wrong directory, wrong filename.
2. Its predicate is `disabled={!…}`. **All 58 `disabled=` sites across the nine propose
   archetypes are `disabled={submitting}`**, and no propose archetype spells `opacity: canSubmit`
   — so widening the predicate alone would still have found nothing.

**Per-kit layer.** This surface has exactly one per-kit suite, `uaProposeTaskContrast.test.ts`.
It is UA-only, and its docblock explicitly refuses to restate measurements made elsewhere.

Neither layer, so it is work under the epic's rule rather than a duplicate of a passing loop.

## The scope question, answered

**`.control-off` on a busy control is out of `disabledControlContrast`'s scope, and the reasons
are structural rather than a preference** — so the answer is now in that file's docblock, as the
issue requires:

- its `expect(marked).toBe(gated)` is an **equality**, so a characterPaths file that also dressed
  its busy control would fail *for wearing the right thing*;
- the walk is bound to `characterPaths/archetypes` by directory and filename whatever the
  predicate says.

The docblock now also names where the busy half *is* asserted, so the next audit stops there.

## Changed

- `frontend/src/pages/proposeTask/archetypes/DefaultProposeTask.tsx` — drops
  `opacity: submitting ? 0.6 : 1` for `className="control-off"`. One property out, one class in.
  No new CSS, no new token: `.control-off:disabled` already exists and already carries
  `!important` precisely so it beats these inline skins.
- `frontend/src/pages/proposeTask/__tests__/submitControlOff.test.tsx` — **new**, 28 assertions.
- `frontend/src/pages/characterPaths/__tests__/disabledControlContrast.test.ts` — **docblock
  only**, no assertion changed. (`WowEditCharacter.tsx` and `wowEditCharacterContrast.test.ts`
  are untouched — #2987 owns those tonight.)

## The guard, and its floor

- Roster is `Object.keys(surfaceMap('proposeTask'))` via the existing shared
  `__tests__/proposeTaskState.ts` — **never a typed list** (#2815). A tenth kit is swept on
  registration with no edit here.
- **Floor: 9 slugs**, a `toBeGreaterThanOrEqual`, not an equality — a real number, and the count
  registered today.
- **The extraction is proven against a live call site per slug**: each busy row is paired with an
  enabled row asserting exactly one `type="submit"` control that is *not* disabled. A guard that
  scanned nothing, or that matched a control which is always disabled, fails there.
- **It does not re-measure the pair.** `--control-off-fill` / `--control-off-ink` and the
  Singularity's re-point are measured once in `disabledControlContrast.test.ts`, whose CSS block
  says consuming sites "should NOT restate the values".

Red first: 4 failed / 24 passed, failing on exactly `na` and `albescent`. Green after the
one-line fix.

## What I checked and found sound, so it is not re-opened

- **The loop layer is real and fully derived** — five suites walk `surfaceMap('proposeTask')`:
  `proposeTaskDispatch`, `unaffiliatedOption`, `metataskProposal`, `proposeTaskBreadcrumb`, and
  now `submitControlOff`. `proposalNotes` tests the pure `planProposalSubmission` planner above
  the archetype layer, invariant by construction.
- **No hand-duplicated invariant across kit suites** — there is only one kit suite to duplicate
  *from*. Nothing for the epic's held phase 2 to collapse here.
- **The field set does not diverge.** Rendering all nine with `canProposeMetatask` and an error
  set gives an identical census for every kit — 2 `input`, 2 `textarea`, 20 `button`,
  1 `role="checkbox"`, 8 `role="radio"`, error surfaced. So **#2993's asked-for structure guard
  has nothing to catch as of today.** I did not write it, to avoid contending with that issue.

## Sequencing against #2993

#2993 was not in flight, so this is measured against `origin/main` @ `0d2712a0`. It rebuilds
`DefaultProposeTask.tsx` on the shared chassis and deletes `pages/proposeTask/factionSurfaces.ts`.
**Whoever picks it up should carry `className="control-off"` onto the chassis' CTA, and check
whether its derived structure guard subsumes anything here** — it should not; different
invariant, same roster.

## Adjacent, reported not fixed

A tree-wide grep while establishing that proposeTask was the only propose-side instance found the
same spelling on three sites outside this surface: `components/JoinControl.tsx:270`,
`components/duel/shared.tsx:539`, `pages/CharacterProfile.tsx:328,349`. Not `proposeTask` cells,
not this issue's scope, so they are flagged on #2814 rather than touched.

## Verification

Every step in the `frontend` job of `.github/workflows/test.yml`, run locally:

| step | result |
|---|---|
| `npm run lint` | clean |
| `npm run typecheck` (`tsc --noEmit` + e2e project) | clean |
| `npm run typecheck:design-sync` | clean |
| `npm test` | **510 files, 10563 passed, 1 skipped** |
| `npm run build` | ok |
| `npm run budget` | JS **WARN** 149.6 KB (warn >130.9, fail >175.8) — pre-existing; CSS ok, ALB ok |

No backend, route, `response_model` or Pydantic change, so `api-schema` and the backend `test` job
are untouched. No `.github/` change.

The budget WARN is inherited from `main`, not caused here: this diff removes one style property
and adds one class inside `DefaultProposeTask`, which is a code-split archetype chunk and not one
of the 15 blocking assets.

## ⚠️ Visual QA is outstanding

Nothing here proves a pixel. Every assertion is `renderToStaticMarkup` with no DOM
(`SPEC-testing.md`), and I have no browser and no dev stack. The contrast *ratio* of the neutral
pair is measured against the stylesheet in `disabledControlContrast.test.ts`, but nobody has
looked at the propose page mid-submit.

## What to eyeball

1. **The `na` / unaffiliated propose page mid-submit, both themes.** The pill loses the rainbow
   frame entirely for the flat neutral while busy. That is #2486's prescription, but the
   spectrum-frame CTA is the loudest thing on this page when enabled and this is the first time
   it goes quiet — confirm the transition does not read as the button vanishing.
2. **The same page with a real faction picked.** `submitButtonStyle` keeps
   `border: 2px solid <tint>` while `.control-off` replaces the fill, so the busy pill is a
   neutral slab inside a coloured ring. The seven faction kits keep their own geometry the same
   way; confirm it reads as deliberate here too.
3. **That the busy state is still legible as busy.** The label swaps to
   `proposeTask.submit.busy` and `.control-off:disabled` forces `cursor: not-allowed`, which now
   overrides the `cursor: wait` this file still asks for — same as the seven sibling kits, left
   for symmetry rather than churn.
4. **The Albescent target page is still byte-identical to the Default.**
   `proposeTaskDispatch.test.tsx` asserts it and it passes, but it is the invariant most likely to
   be broken by the next change here.
5. **Nothing on `createCharacter` / `editCharacter` moved.** Only a docblock changed in
   `disabledControlContrast.test.ts`; its 42 assertions are unchanged and green.

### DIFF
--- a/frontend/node_modules
+++ b/frontend/node_modules
@@ -1,1 +1,0 @@
-/home/pixie/Projects/WorldZeroPlayground/frontend/node_modules
--- a/frontend/node_modules
+++ b/frontend/node_modules
@@ -1,0 +1,1 @@
+/home/pixie/Projects/WorldZeroPlayground/frontend/node_modules
--- a/frontend/src/pages/characterPaths/__tests__/disabledControlContrast.test.ts
+++ b/frontend/src/pages/characterPaths/__tests__/disabledControlContrast.test.ts
@@ -93,6 +93,23 @@ const source = (path: string): string => readFileSync(path, 'utf8')
  * `disabled={!x}` is a control gated on a form's own state — the one that is
  * disabled when the page OPENS. `disabled={deleting}` and friends are transient
  * busy states and are deliberately not swept in here.
+ *
+ * THE BUSY STATE IS OUT OF SCOPE HERE AND IN SCOPE SOMEWHERE — settled by
+ * #2994's audit so the next one does not re-open it. #2486's ruling is about
+ * the disabled PAINT, and a busy control is just as disabled and just as
+ * illegible when it fades; what is out of scope is this FILE, not the class.
+ * Two reasons it stays that way. The `marked === gated` equality below is an
+ * equality on purpose — a file that also dressed its busy control would have
+ * more class lists than gated sites and would fail for wearing the right
+ * thing — and this walk is bound to `characterPaths/archetypes` by directory
+ * and filename, so widening the predicate would still reach only this surface.
+ *
+ * The busy half is therefore asserted per surface, at the rendered control
+ * rather than the source text, over a roster read from the manifest:
+ * `proposeTask/__tests__/submitControlOff.test.tsx` does it for the nine
+ * propose kits, where seven had found `.control-off` privately in #2538 and
+ * the Default still faded. A surface that grows a busy CTA owes a guard of
+ * that shape; it does not belong in this file.
  */
 const gatedControls = (text: string): number => text.match(/disabled=\{!/g)?.length ?? 0
 
--- a/frontend/src/pages/proposeTask/__tests__/submitControlOff.test.tsx
+++ b/frontend/src/pages/proposeTask/__tests__/submitControlOff.test.tsx
@@ -1,0 +1,105 @@
+/**
+ * The propose form's primary action, once it is BUSY (#2994, audit of #2814).
+ *
+ * WHY THIS FILE EXISTS AND `disabledControlContrast` DOES NOT COVER IT.
+ * #2486 ruled that a disabled primary control must DROP its CTA paint for the
+ * measured `--control-off-fill` / `--control-off-ink` pair rather than fade,
+ * because `opacity` composites the whole element — the fill sinks toward the
+ * sheet and the label's ink fades over the faded fill, so the label loses
+ * contrast twice. `characterPaths/__tests__/disabledControlContrast.test.ts`
+ * enforces that, but it sweeps `disabled={!…}` — the control that is disabled
+ * when the page OPENS — and its docblock says busy states are deliberately out
+ * of its scope. It also walks `pages/characterPaths/archetypes` for
+ * `(Create|Edit)Character.tsx`, so it could not reach here whatever it swept.
+ *
+ * NO PROPOSE ARCHETYPE HAS A START-DISABLED CONTROL. All nine gate on
+ * `submitting`, which is the transient class that file excludes. So the
+ * invariant was unenforced on this surface, and seven of the nine kits
+ * discovered `.control-off` privately during #2538 while `DefaultProposeTask`
+ * kept `opacity: submitting ? 0.6 : 1` — the literal spelling #2486 removed —
+ * and `AlbescentProposeTask` inherited it by delegation. That is #2814's thesis
+ * exactly: an invariant hand-written seven times and absent from the Default.
+ *
+ * THE SEAM IS THE RENDERED CONTROL, NOT THE SOURCE TEXT. A source scan would
+ * have to find each kit's file by directory and filename — the coupling that
+ * made the sibling guard unable to see this surface in the first place. Walking
+ * `surfaceMap('proposeTask')` and rendering instead means a tenth kit is swept
+ * the moment it registers, and the Albescent wrapper is measured through what
+ * it actually draws rather than exempted for drawing nothing of its own.
+ *
+ * IT DOES NOT RE-MEASURE THE PAIR. `--control-off-fill` / `--control-off-ink`
+ * and the Singularity's `.sg-control-off` re-point are measured once, in
+ * `disabledControlContrast.test.ts`, and its CSS block says the consuming sites
+ * "should NOT restate the values". This file asserts only that the class is
+ * worn and nothing fades on top of it.
+ */
+import { renderToStaticMarkup } from 'react-dom/server'
+import { MemoryRouter } from 'react-router-dom'
+import { describe, it, expect } from 'vitest'
+import '../../../i18n'
+import { archetypeFor, EVERY_SLUG, proposeTaskState } from './proposeTaskState'
+
+/** The kit's page, rendered for one slug in one busy state. */
+function markupFor(slug: string, submitting: boolean): string {
+  const Archetype = archetypeFor(slug)
+  return renderToStaticMarkup(
+    <MemoryRouter>
+      <Archetype state={proposeTaskState({ factionSlug: slug, submitting })} />
+    </MemoryRouter>,
+  )
+}
+
+/**
+ * The open tags of every `type="submit"` button in `markup`.
+ *
+ * Matched on the tag rather than parsed: these pages carry no `>` inside an
+ * attribute value (React escapes it to `&gt;`), and the two things asserted —
+ * the class list and the `style` declarations — are both attributes of this one
+ * tag. The floor below proves the extraction found something at every slug.
+ */
+const submitTags = (markup: string): string[] =>
+  (markup.match(/<button[^>]*>/g) ?? []).filter((tag) => tag.includes('type="submit"'))
+
+const classesOf = (tag: string): string[] =>
+  /class="([^"]*)"/.exec(tag)?.[1].split(/\s+/).filter(Boolean) ?? []
+
+const styleOf = (tag: string): string => /style="([^"]*)"/.exec(tag)?.[1] ?? ''
+
+describe('the propose sweep reaches every registered kit', () => {
+  it('walks a roster no smaller than the one registered today', () => {
+    // A render scan that renders nothing passes. Nine kits are registered
+    // (albescent, coven, default/na, ephemerists, everymen, singularity, snide,
+    // ua, wow); a floor rather than an equality, so a tenth is swept without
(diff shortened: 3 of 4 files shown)


pixieofhugs:

```
# Review — PR #3008 (#2994: proposeTask busy-CTA invariant)
Reviewed 2026-09-04 against merge-base with main (0d2712a0).

no findings

Verified:
- `className="control-off"` is applied unconditionally, and every `.control-off` rule in `09-component-shared.css` is scoped `:disabled`, so the enabled pill's paint is unchanged.
- `.control-off:disabled` sets `cursor: not-allowed !important`, so the inline `cursor: submitting ? "wait" : "pointer"` is now inert while busy — same as the seven sibling kits; nit only.
- The guard walks `surfaceMap('proposeTask')` with a floor of 9 and pairs each busy row with an enabled row, so it cannot pass vacuously.
- Conflict note: #2993 will rebuild `DefaultProposeTask.tsx`; whoever lands second carries the class onto the chassis CTA.
```

Issue #3001, opened 2026-09-01
60% of the CSS cascade is comments: sweep the tombstones mechanically, and defer the judgement pass


pixieofhugs:
**Owner ruling, 2026-09-01, during `/bug-bot`:**

> "is there any way we can clean up the 6k lines of comments. We don't need every judgment I've ever made, we just need the ones that are relevent now."

## The measurement

`frontend/src/css/` is 10,192 lines, and **60% of it is comments**:

| | |
|---|---|
| comment blocks | 837 |
| comment lines | ~6,100 (of 10,192) |
| actual CSS | ~2,700 lines — 1,062 `--token: value;`, 1,282 declarations, 357 selectors |
| issue references | 1,174, across **367 distinct issues** |

Block sizes: 346 one-liners, 198 of 2–5 lines, 185 of 6–20, **95 of 21–50, and 13 over 50 lines**. Those 108 large blocks hold roughly half the prose.

## The danger, stated first

**Most of these comments are load-bearing, and deleting the wrong one is irreversible in practice.** They carry measured contrast ratios, owner rulings, and — most valuable — the reason not to do the obvious thing. A worked example from this week: `DefaultCreateCharacter`'s ink block records that the app's neutral tiers are used deliberately because the page has no card, with the measured ratios (`--faction-default-card-muted` 4.36 against `--color-text-secondary` 6.06 on the washed ground). Delete that and the next agent "fixes" the lint exemption and ships a contrast failure.

Git history has all of it. Nobody greps git history. **A comment moved into history is a comment deleted.**

So this issue is deliberately scoped to the half that is *provably* dead, and explicitly defers the half that needs judgement.

## In scope: tombstones — provably dead, mechanically detectable

A **tombstone** is a comment whose subject no longer exists anywhere in the tree except in the comment announcing its removal. The detector is one line of reasoning: *if the only occurrence of a name is the comment saying it's gone, the comment is the last thing keeping it alive.*

Verified samples, each occurring in exactly one file — the comment itself:

| name | where the tombstone lives | says |
|---|---|---|
| `--color-rank-accent` | `01-base-tokens.css:3198` | "went the same way" |
| `--color-level-inactive` | `01-base-tokens.css:3218` | the level track's tokens |
| `.eph-vote-star` / `.eph-vote-sparkle` | `06-faction-chrome-2.css:253` | "took them with it" |
[...]

Issue #2918, opened 2026-08-29
Tailwind v4 migration: @tailwindcss/postcss, a CSS-first theme, and a budget ratchet that must not quietly rise


pixieofhugs:
Dependabot opened #2909 to bump `tailwindcss` `3.4.19` -> `4.3.3`. That PR is closed in favour of this issue: v4 is not a version bump, it is a rewrite of how Tailwind is configured and how its output is generated, and it lands on the one file every other lane in this repo already collides on.

## Why it cannot ride in as a Dependabot PR

#2909's `frontend` job failed at the first CSS the build touched:

> `[postcss] It looks like you're trying to use tailwindcss directly as a PostCSS plugin. The PostCSS plugin has moved to a separate package, so to continue using Tailwind CSS with PostCSS you'll need to install @tailwindcss/postcss and update your PostCSS configuration.`

That is v4's *first* gate, not its last. Everything below is still ahead of it.

## What the migration actually has to touch, measured on `origin/main`

| site | today | v4 |
|---|---|---|
| `frontend/postcss.config.js` | `plugins: { tailwindcss: {}, autoprefixer: {} }` | `@tailwindcss/postcss`; autoprefixer is folded in and comes out |
| `frontend/src/index.css:27-29` | `@tailwind base; @tailwind components; @tailwind utilities;` | a single `@import "tailwindcss"` |
| `frontend/tailwind.config.ts` | `darkMode: ["selector", '[data-theme="dark"]']` | `@custom-variant` in CSS — the JS config is no longer the source of truth |
| `frontend/tailwind.config.ts` | `theme.extend.colors`, 9 `var()`-backed keys | `@theme` block |
| `frontend/tailwind.config.ts` | `theme.extend.fontFamily`, 3 stacks | `@theme` block |
| `frontend/src/index.css` | **24** `@apply` sites | `@apply` survives, but resolves against `@theme` rather than the JS config |

`index.css` is **10,201 lines**. It is also the file #2891 wants to split along its section boundaries, and the file every faction lane edits.

## The part that makes this genuinely risky, not just tedious

v4 changes **default values**, not only syntax. The border colour default moves from `gray-200` to `currentColor`, the default ring width changes, `space-y` switches selector strategy, and the preflight's own rules differ. In a normal app that is a handful of visual diffs. Here:

[...]

Issue #2887, opened 2026-08-29
Guard the walk shut: no readdirSync in a test outside sourceScan.ts


pixieofhugs:
Part of #2862. Closes the work in #2885, #2886 and #2815's share.

The walk has now been consolidated twice -- once when `src/test/sourceScan.ts` was written to replace
nine drifted copies, and once by #2885 and #2886. Nothing stopped it the first time.

## What to build

A guard that fails when a test file calls `readdirSync` (or otherwise hand-rolls a source walk)
without going through `src/test/sourceScan.ts`. This is itself a source-scanning guard, so it uses the
seam it protects.

`sourceScan.ts` is exempt by construction. Any other exemption is an allowlist entry with a reason,
following the ratchet pattern the repo already uses -- and like those ratchets it should only shrink.

## Acceptance criteria

- [ ] Adding a private `readdirSync` walk to a test reddens the guard
- [ ] `sourceScan.ts` itself passes
- [ ] Any exemption is named with a reason, and the guard fails if the count grows

## Blocked by

- #2885
- #2886
- #2815's share of the nine dual-scope files

PR #3012, opened 2026-09-02
A tombstone detector for the CSS cascade, and the report it produces (#3001)

pixieofhugs:
Part of #3001

**This PR is now one file: `docs/research/css-tombstones.md`.** The detector that produced it (`frontend/scripts/css-tombstones.mjs`) and its vitest guard (`cssTombstones.test.ts`) were removed from the branch on 2026-09-04. `frontend/src/css/` is untouched — `git diff origin/main -- frontend/src/css` is empty — and nothing else in the tree changes.

## Why the tool was cut

The owner ruled on 2026-09-04 that the **report is the deliverable, not the tool** — option (a) in the [review of this PR](https://github.com/pixieofhugs/WorldZeroPlayground/pull/3012#issuecomment-5538844131). That review found the detector's headline numbers accurate for this commit (its live output was byte-identical to the committed report) but the way they were produced unsound in three places: the KEEP-reason verdict hung on a word list containing ordinary English (`measured`, `ratio`, `instead`, `never`, `would`, `cannot`), so 4 of the 7 KEEP-reason blocks may be removal records the list happened to spare; a retirement guard written as `not.toContain('var(--x)')` was invisible to the guard route, so a "no guard" citation may be wrong; and the colour route failed silently on a stray apostrophe in surrounding prose. It also re-implemented `src/test/sourceScan.ts` just outside the `noPrivateSourceWalk` guard (#2887). ~1,100 lines plus a permanent guard to locate 3 deletable lines was not worth keeping; the report is, because it is the human-readable half #3001 asked for before any deletion.

The report's header now records its provenance — generated once on 2026-09-04 by the detector as it stood at `bdf6c4ea` of this PR, a snapshot of `origin/main` @ `0d2712a0` that will drift — and carries a **"Caveats found in review"** section stating those three findings plainly, so the tables can be read with the right amount of trust. The tables themselves are unchanged.

## What the owner should do next

1. **Read `docs/research/css-tombstones.md`**, caveats first. The CUT-candidate section is 1 block / 3 lines (`01-base-tokens.css:3218`); the caveats say why that is a floor, not a verdict, and which section (KEEP-reason, 7 blocks / 91 lines) to read by hand next.
2. **Decide the cut list on #3001** — whether it is the 3 lines, the 3 lines plus some of the KEEP-reason blocks, or "measured, not worth it".
3. **File the deletion PR separately**, against `frontend/src/css/`, once decided. That PR is the one that must not collide with #2918 (Tailwind v4), which is queued against the same files.

#3001 closes when the owner has read the report and ruled, not on this merge.

## Verification

Frontend CI steps re-run after the removal: `npm run lint`, `npm run typecheck`, `npm run typecheck:design-sync`, `npm test` (the deleted test is simply gone), `npm run build`, `npm run budget` — counts are in the trimming commit's message. No backend or schema change; `.github/` untouched. Nothing here is visual.


pixieofhugs:

# Review — PR #3012 (#3001: CSS tombstone detector + report)
Reviewed 2026-09-04 against merge-base 0d2712a0. Two finder angles (correctness, cleanup) + a live run of the detector: its output is byte-identical to the committed `docs/research/css-tombstones.md`, so the PR's numbers are accurate for this commit. The findings are about how those numbers are produced.

## Correctness (CONFIRMED)
1. `css-tombstones.mjs:316` — `quotedSpans` pairs any quote kind with any other; a possessive apostrophe in the prose before a backticked colour drops the colour. Demonstrated. The colour route (7 of the original 10 cuttable lines) is one apostrophe from silent failure; the test covers one block.
2. `css-tombstones.mjs:260` — the guard route needs the dead name as the ENTIRE quoted string, so a retirement guard written as `not.toContain('var(--font-faction-poster)')` (live today in `everymenWearsTheBillRegister.test.ts:121`) registers nothing; that block is off the cut list only by luck of position.
3. `css-tombstones.mjs:232` — `REASONING` matches ordinary English; 4 of the 7 KEEP-reason blocks (31 lines) are kept by one generic word each. The "3 lines, not worth it" verdict the owner is asked to ratify is a property of the word list, and the report does not say so.
4. `css-tombstones.mjs:126` — `attachedTo` reads the line below, mis-attributing trailing inline comments (latent; none names a dead token today).

## Reuse / altitude (CONFIRMED)
5. `css-tombstones.mjs:65` — re-implements `src/test/sourceScan.ts` (`sourceFiles`, `toRelative`, `readStripped`) just outside the `noPrivateSourceWalk` guard (#2887) that forbids a tenth copy; its `//` stripper eats `://` against the convention four sibling tests document; `liveNames` does not fold through `resolveRoleReads()` as `frontend/CLAUDE.md` asks.
6. `cssTombstones.test.ts:51` — shells out to the script and re-declares its JSON contract; the script's eight exports have no importer; nothing pins the counts or checks the committed report is fresh (a `toMatchFileSnapshot` would pass byte-for-byte today).

## The scope question (owner decision)
The issue asked for a mechanical sweep and a deferred judgement pass. The PR is ~1100 lines plus a permanent vitest guard to find 3 deletable lines in one block. Two honest ways to land #3001:
- (a) keep the REPORT as the deliverable, drop the vitest guard and the script from the tree (or park the script under `scripts/` with no test), and close #3001 with the owner's read of the report; or
- (b) keep the tool, fix 1–3 and 5, pin the report in the test, and move it under `src/test/` so it can import `sourceFiles()` / `indexCssParts()`.
(a) is the lazy answer. Not dispatched for a fix because the choice is the owner's.

## Verdict
Not mergeable as-is: finding 1 is a demonstrated bug in the headline mechanism and finding 5 lands the exact copy a repo guard exists to prevent. Ready-for-human on the (a)/(b) choice.

PR #2760, opened 2026-08-27
HAND MERGE ONLY - Metamorphosis becomes the live era: one line, and the paperwork that pins it

pixieofhugs:
Closes #2711

# ⚠️ Merging this does NOT open the era — it only deploys the config

**Merging this PR deploys Era 2's rules to production** (`main` auto-deploys). It does **not** write the new `Era` row. `scripts/era_reset.py` does, and it must be run against the target database afterwards:

```
python scripts/era_reset.py <admin-username> --env prod        # prints the plan, changes nothing
python scripts/era_reset.py <admin-username> --env prod --yes  # performs it
```

**Between the deploy and that run, the app serves Era 2's rules against Era 1's `Era` row.** The seeder warns about it on every deploy.

**There is no undo.** Scores, levels, vote budgets and faction membership reset per the era's flags; every unresolved duel freezes into a permanent result; the whole board retires.

I have not run `era_reset.py` against anything. This PR is a **draft** deliberately — the rollover is the owner's call, not this branch's.

Note (from the issue): a **wiped** database never exercises the rollover at all — `bootstrap_admin` writes the `Era` row only on an un-bootstrapped database, so a fresh seed just creates `era_2` directly. To test the rollover as an operation, test it on a database that already has an `era_1` row.

---

## The seam

`game_config.__getattr__`'s `CURRENT_ERA` branch — the single lever. Everything downstream reads `era.*` off whatever that branch returns, so this is the one place the rollover is expressible and the one place it is pinned.

The loop went red at that seam first: I flipped the assertion in `test_era_config.py` before touching `game_config.py`, and it failed on the real defect (`CURRENT_ERA` was `EraConfig(name='Renaissance', config_key='era_1', ...)`). Then the one line made it green.

## What changed — 4 files, 29 insertions

| File | Change |
|---|---|
| `backend/game_config.py` | The one line: `CURRENT_ERA`'s branch imports `ERA_2` instead of `ERA_1`. The stale "authored, not activated" comment on the `ERA_2` branch now restates *why that branch is still separate* (#2708's half-flipped-process bug). |
| `backend/eras/era_2.py` | Docstring: "Authored, not activated" → the live era, plus the deploy-vs-rollover distinction above. |
| `backend/tests/unit/test_era_config.py` | `test_era_2_is_authored_not_activated` → `test_era_2_is_the_live_era`. **Assertion flipped, not deleted** — the lever stays pinned in exactly one place, so it cannot move silently in either direction. |
[...]

Issue #2710, opened 2026-08-26
Dev and e2e fixtures break on a fresh Era 2 database


pixieofhugs:
`scripts/seed_demo_praxes.py` hardcodes Era 1 faction slugs, so seeding a fresh **dev** database under any era that lacks them crashes.

```python
METATASK_FACTION_SLUG = "ua"       # line 100  — not in Era 2
DUEL_TASK_FACTION_SLUG = "snide"   # line 125  — fine
COLLAB_TASK_FACTION_SLUG = "coven" # line 145  — not in Era 2
```

## The actual failure is smaller than it looks

The collab fixture **already skips** when its task is absent, and that skip is already tested (`test_seed_score_fixtures::test_collab_fixture_skips_when_the_board_has_no_coven_task`). The crash is in the *summary line about the skip*:

```python
collab_modifier = era.factions[COLLAB_TASK_FACTION_SLUG].collab_own_modifier   # line 516
```

Unconditional, so the report about not seeding a Coven collab is what raises `KeyError`.

Phase 5 is dev-only (`if env == "dev"` in `seed.py`), so **production is unaffected** — but it fires the moment anyone seeds a fresh dev database under Era 2.

`seed.py:305`'s `DUEL_FIXTURE_TASK_FACTION_SLUG = "ua"` is already guarded and no-ops when the row is absent; leave it.

## Decision (owner, 2026-08-25)

**Drop the perk-demonstration property.** The fixtures pick any faction the live era defines. The collab fixture seeds *a collab* — it no longer claims to demonstrate an own-faction bonus, because no Era 2 faction has one and hunting for a holder is not worth it for a dev convenience.

## Scope

1. The three constants become era-read rather than literals: any non-sentinel faction the live era defines.
2. Guard the line 516 summary print so the skip path can report itself.
3. Update the docstrings and printed summary — they currently promise "own-faction collab at xN"; that claim goes.
4. `tests/integration/test_seed_score_fixtures.py` — 7 failures under Era 2. In particular `test_seeded_collab_carries_the_own_faction_collab_modifier` re-points at *whatever this era's modifier is for the slug the fixture used*, which is the same derive-from-the-actual-slug rule as the era-agnostic fixtures issue.

## Acceptance

- `python seed.py --env dev` succeeds on an empty database under Era 1 and under Era 2.
- No fixture docstring or printed line claims to demonstrate a bonus the era does not grant.
- Full suite green under Era 1.

[...]

Issue #2708, opened 2026-08-26
Integration fixtures must ask the era for a faction, never name one


pixieofhugs:
The integration suite is written against Era 1's roster. Pointing `CURRENT_ERA` at Era 2 produces **36 failures and 3 modules that fail at import**, none of which is a bug in Era 2's config — every one is a place that named an Era 1 faction instead of asking the era a question.

## The shape of it

`conftest.faction_ua` seeds exactly two literal rows (`ua` visible, `na` hidden), and `factories.py` defaults every character, task and duel side to `faction_slug="ua"` — **257 `"ua"` literals across 82 files**. The factory docstring admits the slug is arbitrary:

> "`ua` is the default slug because it is the only faction the shared fixtures seed a row for, and most callers do not read the slug at all."

Failure breakdown:

| # | Cause |
|---|---|
| 10 | FK violation — a fixture seeds a `Task`/`Character` into a slug with no `Faction` row |
| 8 | `KeyError` on `CURRENT_ERA.factions["ua" / "coven" / "singularity"]` — 3 are module-level, so the file dies at import (`test_collab_vote_recalc:33`, `test_era_scoped_recalc:42`, `test_habit_bonus:37`) |
| 9 | Scoring assertions that only hold under Coven's 1.1 collab modifier |
| 7 | `test_level_jump_signup` fixture errors |
| 4 | Route rejections (400/422/404/403) — **correct behaviour**, the faction does not exist |
| 1 | Ephemerists' retired-task carve-out |

## Decision (owner, 2026-08-25)

**A test may know only that it got *some* faction, never which one.**

1. **Fixture default** = the first non-sentinel faction the live era defines (excluding `na` and `albescent`, which are structural — ADR-0087). Under Era 1 this resolves to `'ua'`, **byte-identical to the literal it replaces**, so the migration is provably behaviour-preserving while Era 1 is live.
2. **Score assertions derive from the actual slug**: `era.factions[character.faction_slug].own_task_modifier`, never a named slug and never a hardcoded product. **7 sites already do this** (`[character.faction_slug]`, `[challenger.faction_slug]`, `[winner.faction_slug]`) and survive the flip untouched — they are the pattern to copy.
3. **Tests whose subject is a specific perk** do not belong here. They find the holder by the perk and skip if the era has none, or they move to that era's rules module (see the per-era module issue).

## Two traps — both cost a full test run when attempted

[...]

Issue #2536, opened 2026-08-24
81 ADRs, no index, and no way to ask "which of these are still rules"


pixieofhugs:
81 ADRs, no index, and no way to ask "which of these are still rules"

## State

`docs/adr/` holds 81 records. There is **no index file** — no `README.md`, no table.

Status lines are inconsistent to the point of being unparseable:

- **28 records have no `Status` line at all** — 0001–0006, 0009–0014, 0017–0023, 0027, 0030, 0032, 0033, 0035, 0036, 0037, 0047, 0052
- of the 53 that do, **49** spell it `**Status:**` and **4** spell it `Status:`
- the value is free prose, ranging from `Accepted (2026-06-24)` to `Accepted (2026-06-24); the *Merit (the card number)* section superseded by …` to `Reversed` to `Accepted — **built, except presence.**`

So a machine cannot answer "which ADRs are live", and a person answering it has to read all 81.

## Why it matters now

The game is in its final stretch. Most rules are made. The question being asked of this directory has changed from *"what did we decide"* to *"what is still true"* — and the directory cannot answer it.

Superseded records are not marked at the top where a reader lands. Known reversals and supersessions today, none of them discoverable without a full read:

- **ADR-0009** superseded by ADR-0077 (blocks became their own record)
- **ADR-0021** superseded by ADR-0080 (Albescent's unlock)
- **ADR-0046** `Reversed`
- **ADR-0047** superseded by ADR-0053
- **ADR-0035** "superseded per surface — see the banner above"
- **ADR-0020**, **ADR-0027** — check both; ADR-0027's hiding was superseded by #2409's redaction

Stale statuses too. **ADR-0073** reads *"Still design only: presence (#1744)"*. `frontend/src/pages/editPraxis/roomPresence.ts` is 250 shipped lines with two tests beside it (`roomPresence.test.ts`, `roomPresenceContrast.test.ts`). The record describes a build that finished.

## Fix

**1. One status vocabulary, in a fixed position.** A line-2 field, one of exactly:

```
**Status:** Accepted | Superseded by ADR-NNNN | Reversed | Amended by ADR-NNNN
**Date:** YYYY-MM-DD
```

Prose about *which clause* was superseded goes in the body, under an `## Amendment` heading — not in the status value. Backfill all 81, including the 28 with none.

[...]

PR #3013, opened 2026-09-02
The latest Era row chooses the live ruleset, and a mod picks it from the admin page (#827)

### LINKED ISSUE #827: Mod control: queue multiple eras and choose which to roll into
@
Molly wants **several era configs queued in the codebase** and a **mod-UX control to choose which era to switch the game into**. This is the control that *ends* an era — the trigger that freezes duels (#824) and starts the next era. Surfaced while closing #719.

## Today
- `CURRENT_ERA` is **hardcoded to era_1** (`backend/game_config.py:170-178`); only `era_1.py` exists (+ `_template.py`).
- The admin reset endpoint (`backend/routers/admin.py:224` `PUT /era/reset`) re-instantiates the **same** era (`CURRENT_ERA.config_key`) — no next-era or selection concept.
- No era registry, no ordering, no runtime selection. There **is** an admin UI to host the control (`frontend/src/pages/Admin.tsx`, `frontend/src/api/admin.ts`).

## The crux — needs a decision/ADR before build
For "choose between eras" to actually change how the game **plays** (not just reset stats), the **active ruleset must resolve at runtime from the chosen era**. Today every service defaults `era: EraConfig = CURRENT_ERA` — a compile-time constant. CLAUDE.md states *"Changing CURRENT_ERA is the one lever"* and *"Era.config_key ... never owns the rules."* Making the active era data-driven (latest `Era` rows `config_key` → registry lookup) **reverses that invariant**.

Decision needed: does the active ruleset become a runtime resolution of the latest `Era` rows `config_key`? (Almost certainly yes for this feature.) → ADR.

## Scope (pending the decision)
- **Era registry:** enumerate available era configs by `config_key` — an explicit registry (each era file registers itself; avoid filesystem magic). "Error unless a valid era file" = the chosen `config_key` must be in the registry.
- **Active-era resolver:** replace the hardcoded `CURRENT_ERA` default with a resolver returning the config for the latest `Era` rows `config_key` (default fallback for a fresh DB). Services keep taking `era: EraConfig`; only the default *source* changes. **This is the large refactor** — every `= CURRENT_ERA` default.
- **Endpoint:** admin picks a target `config_key` → validate its registered (error otherwise) → create the `Era` row + run `apply_era_reset` (which freezes duels per #824). Incoming eras reset flags apply.
- **Mod UX:** a selector on the Admin page listing available eras + a confirm (destructive — resets everyone and freezes all duels). Add a GET to list available eras for the dropdown.

## Composes with
- #824 (duel freeze fires when this control ends the era).
- Downstream: #823 (badge reads frozen prior-era winners).

Needs triage/grill + an active-era-resolution ADR before its agent-ready.
@

pixieofhugs:
Closes #827

Mods can now end an era from the admin page and choose which era it opens into. The live ruleset resolves from the latest `Era` row, per the owner's 2026-09-01 ruling.

**Visual QA is outstanding.** I have no browser and no dev stack in this environment; everything below was verified through tests, `tsc`, eslint and the build. See "What to eyeball".

**The backend integration tests in this PR are unrun.** There is no Postgres available here — no docker group membership, no local server, and `sudo` needs a password. The 523 backend *unit* tests pass, all 1237 integration tests collect cleanly (so nothing is broken at import), and every frontend CI step passes. CI is the first thing to actually execute `backend/tests/integration/test_admin_era_rollover.py`.

---

## The seam I tested at

`game_config.CURRENT_ERA` as an **object identity** rather than a name, and `services.era.apply_era_reset` as the one operation that moves it.

`backend/tests/unit/test_live_era_binding.py` is the red-first test. Its core case defines, at module scope, a function with `era: EraConfig = CURRENT_ERA` — exactly as the 157 real ones are written — and asserts that after a flip it reports the new era.

## One correction to the ruling, and why it is a correction rather than a re-litigation

The ruling's decision stands untouched: the latest `Era` row chooses the ruleset, bound at start-up and after `apply_era_reset`, services keep taking `era: EraConfig`, 157 call sites unchanged. What does not survive contact with Python is the ruling's *mechanism* — "rebind the name `CURRENT_ERA`":

> Option A leaves every one of them untouched **and correct**.

A default argument is evaluated **once, when the `def` executes**. All 157 functions therefore hold the *object* named at import time, for the life of the process. Four more sites read `CURRENT_ERA` as a module global inside a function body, which resolves against the *importing* module's globals, not `game_config`'s. `game_config.CURRENT_ERA = other` moves neither population — it rebinds one name in one module while every holder keeps what it captured. The process would end up **half-flipped**: a handful of readers on the new era, the overwhelming majority on the old. That is the same failure mode #2708 found, arriving from the other direction.

So `CURRENT_ERA` is now one stable `EraConfig` instance, distinct from every era file's config, whose fields are refreshed in place. Every holder — default argument, module global, closure — sees the new ruleset at once. This is the only mechanism under which the ruling's own promise ("157 call sites untouched and correct") is actually true rather than merely stated, and it makes start-up ordering a non-issue as a bonus: the lifespan can bind after the routers have imported, because they all point at the same object.

Two consequences, both taken deliberately and both documented in ADR-0091:

- `CURRENT_ERA is ERA_1` is now **false**; `CURRENT_ERA == ERA_1` is the question. Correct on the merits — `CURRENT_ERA` means "the live binding", not "era 1" — and it moved three assertions (one unit test, two integration fixtures) that compared eras by identity.
- The live instance must never *be* an era file's config, or the first flip would overwrite `ERA_1` in the process. Pinned by a test.

## The central hazard (#2708) survives, and is re-pinned

`game_config.__getattr__` still keeps `CURRENT_ERA` and `ERA_N` in separate branches. Two tests now cover it — one in `test_live_era_binding.py` that resolves `era_1` while Era 2 is live and asserts both the module attribute and a def-time-bound default are unmoved, and the existing `test_resolving_a_past_era_does_not_rebind_the_live_one`, rewritten to flip through the real lever instead of a stand-in name assignment (that stand-in tested a name nothing reads any more).

## Three regressions this change would otherwise have shipped

None was in the brief; all three were found by tracing what reads `CURRENT_ERA` at runtime, and all three are silent.

1. **`backend/seed.py` runs on every deploy** (`start.sh`) and its Phase 1 is a two-way mirror: it marks the era's roster `visible` and every other slug `retired`. It read the *compile-time* era. So the first deploy after a mod rolled the game into Era 2 would have re-marked Era 1's factions visible and Era 2's retired — silently undoing the roster half of the rollover, with the `Era` row still correctly naming Era 2. It now seeds the live era.
2. **`backend/scripts/era_reset.py`** would have re-opened the compile-time era, i.e. silently rolled the game *back* to Era 1 after a mod rolled it forward. It now resolves the live era first — one call, and `print_plan` and the row it builds needed no edit, because the live era is one object refreshed in place.
3. **`seed.DUEL_FIXTURE_TASK_FACTION_SLUG`** was a module constant, evaluated when `seed` is imported — which is before `seed()` resolves the live era. Its value is `ua`, which Era 2 does not carry. This is #2710's exact symptom reached by a new route, and the comment sitting directly above that line says so ("A slug read off the era cannot go absent, so the guard stays as a backstop rather than as the thing that decides whether the suite has a task at all"). Worse than #2710, in fact: the `Faction` row for `ua` still exists, because rows retire and never delete, so the `faction is None` backstop would wave it through and seat the duel fixture task in a faction the live era does not carry. It is now `duel_fixture_task_faction_slug(era=CURRENT_ERA)`, read at call time. Demonstrated directly: before the flip it returns `ua`, after `bind_live_era(era_2)` it returns `snide`.

A read-only sweep of all 675 `CURRENT_ERA` occurrences under `backend/` found no *other* shipped-code site of either class — no remaining identity comparisons, and no module-level derivations under `services/`, `routers/`, `models/` or `scripts/`. It did find ~30 module-level constants in the **test** suite with the same import-time-freeze shape (`tests/integration/factories.py::DEFAULT_FACTION_SLUG` being the widest). All are latent — the autouse `restore_live_era` fixture contains them and the one file that genuinely flips to Era 2 imports none of them — so they are deliberately **not** in this PR; converting ~30 constants is churn that would bury the change. Flagged separately; a guard test that fails when a new module-level derivation appears is probably the better-value half of that work.

`seed.stale_era_warning` also had its advice inverted. It used to warn "era_2 is configured but the live row is era_1 — run `era_reset.py`", which was the deploy window between an owner's code flip and the hand rollover. That window no longer exists, and the warning would now be telling an operator to undo a mod's rollover. It now fires on the one state still genuinely wrong: a row naming a `config_key` no era file registers, where the process has fallen back and the game is quietly not in the era the database says it is in.

## What is in the PR

**Backend**
- `game_config.py` — `bind_live_era` (the primitive), `_new_live_era`, `registered_era_config_keys()`, `COMPILE_TIME_ERA_CONFIG_KEY`, and the module docstring corrected.
- `services/era.py` — `rebind_live_era(session)`, the **only** caller of the primitive; falls back to the compile-time era with a log line both when there is no `Era` row (fresh database — does not invent one, `Era.started_by` is a not-null audit trail) and when the row names an unregistered key. Called from the app lifespan and from the end of `apply_era_reset`.
- `services/admin_service.py` — `list_registered_eras()`, `roll_into_era()`.
- `routers/admin.py` — `GET /admin/eras`, `PUT /admin/eras/live`. An unregistered key is a coded 422 (`ERA_CONFIG_UNKNOWN`) rather than an uncoded raise, because the uncoded-error allowlist only shrinks.
- `backend/openapi.json` + `frontend/src/api/generated/schema.d.ts` regenerated by `scripts/regen_api_client.py`; `git diff --exit-code` on both is clean.

**The guard the ruling asked for.** `test_only_services_era_rebinds_the_live_era` greps the shipped backend and fails if anything but `game_config.py` and `services/era.py` calls `bind_live_era`. (Word-boundary matched, so `rebind_live_era` — which callers legitimately use — is not counted.)

**Test isolation.** `apply_era_reset` rebinds a process-wide object, so an autouse fixture in `backend/tests/conftest.py` restores the live era after every test. Without it a rollover in one test would hand Era 2 to every test after it in the same session, in file order, silently.

**Docs.** New ADR-0091 (number taken from `ls docs/adr/` at authoring time, not from the issue). It amends ADR-0042's "the DB owns only history" clause — status line updated to `Amended by ADR-0087, ADR-0091` plus an `## Amendment` paragraph, per #2536's vocabulary; no other ADR body rewritten. `docs/adr/README.md` regenerated with `scripts/adr_index.py`. The three records that asserted the reversed invariant are corrected in the same commit: `CLAUDE.md`'s "Config architecture", `game_config.py`'s docstring, `models/era.py`'s `config_key` comment.

**Frontend** — a new Era tab (last in the row, the destructive one), a `<select>` of registered eras, and a two-step confirm that spells out in words what a rollover does. The live era stays in the select as a `disabled`, marked option and is never a target — rolling into the era you are already in is a second reset for nothing. The decision logic is extracted to `eraRollTargets.ts` and unit-tested, matching how `adminTaskRows.ts` is split out of `TasksTab.tsx`.

## Verification actually run

Backend (`backend/.venv/bin/pytest`, dedicated test DB not needed for these):
- `tests/unit` — **523 passed**, including the ruff ratchet, the uncoded-error ratchet, the i18n catalog coverage guard and the ADR index test.
- `tests/integration --collect-only` — 1237 collected, no import errors. **Not executed: no Postgres in this environment.**
- `game_config.py` coverage from unit tests alone: 100%.
- `python scripts/regen_api_client.py` then `git diff --exit-code` — clean (the `api-schema` job).

Frontend, every step named in `.github/workflows/test.yml`:
- `npm run lint` — clean
- `npm run typecheck` — clean
- `npm run typecheck:design-sync` — clean
- `npm test` — 510 files, **10543 passed**, 1 skipped
- `npm run build` — succeeds
- `npm run budget` — `WARN JS 150.1 KB (warn>130.9, fail>175.8)`, exit 0. **Pre-existing**: the baseline is ~149.7 KB, already well past the warn line before this branch. This change adds ~0.4 KB gzip, all of it the new copy landing in the eagerly-imported i18n chunk; `EraTab` itself is behind the already-lazy `Admin` route. 25 KB under the fail ceiling.

## Not touched

PR #2760 and its branch, as instructed.

---

## What to eyeball

I could not open a browser. Everything here is a *visual* or *interaction* claim that tests cannot make:

1. **The Era tab in the tab bar** — it is last, after Overview. Check the underline/active treatment matches its four neighbours, and that `#era` in the URL selects it on reload (the tab bar drives `window.location.hash`).
2. **Step one, the selector.** The live era appears in the `<select>` as a `disabled` option labelled "*{name}* — live now", and the dropdown defaults to the first non-live era. Judgement call worth a second opinion: I kept the live era visible-but-unpickable because the list is the register and where the game sits in it is the fact you need before choosing. If you would rather it were dropped from the select and stated in prose above, that is a one-line change.
3. **Step two, the confirm card.** Red border (`--color-danger-edge`), red heading, three effect bullets, bolded no-undo line. The confirm button is `btn-primary` (strong neutral ink), **not** a red button — `--color-danger` measures 4.40:1 on light, so red-on-white at `text-xs` would be under AA. Please sanity-check that it still *reads* as dangerous enough for an irreversible action, in both light and dark.
4. **Dark mode** on both the confirm card and the select. No `dark ? a : b` ternaries were used; it is all CSS variables, but nobody has looked at it.
5. **In flight.** The confirm button really disables and swaps to the working copy; Cancel is *hidden* rather than disabled while the request is out, since cancelling cannot recall a rollover already happening. Check that the buttons do not jump distractingly when Cancel disappears.
6. **The success readout**, after a real rollover on a scratch database: "{name} is open" plus the pluralised character count, and the selector re-fetches so the new era shows as live and is no longer pickable.
7. **The error path** — send a bad key (or point at a stale build) and confirm the 422 renders the `ERA_CONFIG_UNKNOWN` sentence from `errors.json`, not a raw body.
8. **The one-era case** (`era.noTarget`) is unreachable today with two eras registered, so it is untested by eye. Worth a look if you ever cut back to one.

And the thing worth confirming on a scratch database rather than by eye: **that a rollover actually changes how the game plays in the live process** — roll into Era 2 and then check a surface that reads a rule off the era (the faction roster, a level threshold, the signup cap) *without* restarting the backend. That is the whole point of the change, it is the part the unit tests model rather than prove end-to-end, and it is the one I could not run.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,7 +17,7 @@ identity + era-as-ruleset in ADR-0041 / ADR-0042.
 
 | Need... | Go to |
 |---|---|
-| Active rule values (signup cap, vote budget, level thresholds, resets) | `backend/eras/era_1.py` (live `ERA_1`; `CURRENT_ERA` resolves here) |
+| Active rule values (signup cap, vote budget, level thresholds, resets) | `backend/eras/era_1.py` (`ERA_1`; which era is live is the latest DB `Era` row — ADR-0091) |
 | Factions, tasks, level ranks/unlocks + taunt structure for the live era | `backend/eras/era_1.py` |
 | Taunt & rank/unlock **wording** (ADR-0031: backend emits keys) | `frontend/src/locales/en/{taunts,progression}.json` |
 | Faction **name/description** wording (ADR-0038: backend emits slug) | `frontend/src/locales/en/factions.json` (`names.<slug>`, `descriptions.<slug>`) |
@@ -54,7 +54,14 @@ Read only what your task needs.
 
 ## Config architecture
 - `game_config.py` = dataclass shape. `eras/era_N.py` = values. `CURRENT_ERA` = active era.
-- DB `Era.config_key` records which era was active; it does not own rules.
+- **The latest DB `Era` row's `config_key` chooses which era is active** (ADR-0091).
+  It owns no rule *value* — those stay in `eras/` — but it decides which set is
+  live, and a mod sets it from the admin page. `CURRENT_ERA` is still the one
+  lever; the hand on it moved off a code edit.
+- `CURRENT_ERA` is a stable **object identity**, refreshed in place. Never
+  reassign `game_config.CURRENT_ERA` — a default argument binds at `def` time,
+  so that moves one name and leaves 157 call sites on the old era. The only
+  rebinding site is `services.era.rebind_live_era`, and a test keeps it at one.
 - Services take `era: EraConfig = CURRENT_ERA`. Never import `CURRENT_ERA`
   inside a service body.
 - Never hardcode a value that lives in `EraConfig`. Read `era.*`.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -60,8 +60,8 @@ Read only what your task needs.
   lever; the hand on it moved off a code edit.
 - `CURRENT_ERA` is a stable **object identity**, refreshed in place. Never
   reassign `game_config.CURRENT_ERA` — a default argument binds at `def` time,
-  so that moves one name and leaves 157 call sites on the old era. The only
-  rebinding site is `services.era.rebind_live_era`, and a test keeps it at one.
+  so that moves one name and leaves every call site on the old era. The only
+  module that rebinds it is `services/era.py`, and a test keeps it there.
 - Services take `era: EraConfig = CURRENT_ERA`. Never import `CURRENT_ERA`
   inside a service body.
 - Never hardcode a value that lives in `EraConfig`. Read `era.*`.
--- a/backend/db.py
+++ b/backend/db.py
@@ -27,6 +27,15 @@ async def get_db() -> AsyncGenerator[AsyncSession, None]:
     Services call ``session.flush()`` only — the router's dependency owns the
     single per-request commit. If the handler raises, the transaction rolls back
     and the exception propagates unchanged.
+
+    **One deliberate exception**, and it is the era rollover:
+    ``services.era.commit_and_bind_live_era`` commits inside the service,
+    because the process-wide live-era binding has to land *after* the write is
+    durable and this dependency's commit runs after the handler has already
+    returned (ADR-0091, #827). The commit below then finds nothing left to
+    write. Do not read it as a precedent — anything else that wants its own
+    commit needs the same "a non-transactional side effect must follow it"
+    argument.
     """
     async with AsyncSessionLocal() as session:
         try:
--- a/backend/eras/_template.py
+++ b/backend/eras/_template.py
@@ -8,10 +8,23 @@ Steps:
   2. Rename all ERA_N references to match (e.g. ERA_2, ERA_2_FACTIONS, etc.)
   3. Fill in every section marked with TODO
   4. Update backend/game_config.py:
-       - Add: from eras.era_N import ERA_N
-       - Add the key to _ERA_ATTRIBUTE_BY_CONFIG_KEY ("era_N": "ERA_N") -- this
-         is the registry every era-aware tool walks, step 7 included
-       - Change: CURRENT_ERA = ERA_N
+       - Add an `if name in ("ERA_N", "ERA_N_FACTIONS"):` branch to `__getattr__`,
+         copying the ERA_2 one. The lazy import is what keeps eras/ out of a
+         circular import with game_config; never import the era file at module
+         scope here.
+       - APPEND the key to _ERA_ATTRIBUTE_BY_CONFIG_KEY ("era_N": "ERA_N") --
+         this is the registry every era-aware tool walks, step 7 included, and
+         its order is the order the admin era selector offers. Append, never
+         insert.
+       - Do NOT assign CURRENT_ERA. Which era is LIVE is a database fact now
+         (ADR-0091, #827): the latest `Era` row's config_key chooses it, and
+         `services.era.rebind_live_era` binds it at start-up. Reassigning the
+         name would move nothing anyway -- `era: EraConfig = CURRENT_ERA`
+         defaults bind at def time, so the process would end up half-flipped.
+         A mod opens the new era from the admin page (Admin > Era), or an
+         operator runs `python scripts/era_reset.py USERNAME --yes` from a
+         shell. Adding the era file makes it *offerable*; it does not make it
+         live, and nothing about this step touches production's ruleset.
   5. Run: python seed.py --env dev
   6. Run tests: pytest tests/unit/test_era_config.py
   7. Write this era's rules module (#2709). It is not optional:
--- a/backend/errors.py
+++ b/backend/errors.py
@@ -245,6 +245,16 @@ class ErrorCode(str, enum.Enum):
     #: is no sign-in. The player starts one.
     returning_player_consent_expired = "RETURNING_PLAYER_CONSENT_EXPIRED"
 
+    # -- Era rollover (#827, ADR-0091) --------------------------------------
+    #: A rollover naming a ``config_key`` no era file registers. Coded despite
+    #: ``routers/admin.py`` being an otherwise-uncoded staff surface, for the
+    #: same reason as :attr:`character_departed_not_restorable`: this raise is
+    #: new and the uncoded-raise allowlist only shrinks. It is also the one
+    #: refusal standing between a typo and an ``Era`` row that names no
+    #: ruleset — a row whose rules the process cannot resolve, on the most
+    #: destructive operation in the system.
+    era_config_unknown = "ERA_CONFIG_UNKNOWN"
+
 
 #: The keys of a coded ``detail`` body. Named so the frontend contract is
 #: greppable from Python and no raise site spells them as bare literals.
--- a/backend/errors.py
+++ b/backend/errors.py
@@ -254,6 +254,14 @@ class ErrorCode(str, enum.Enum):
     #: ruleset — a row whose rules the process cannot resolve, on the most
     #: destructive operation in the system.
     era_config_unknown = "ERA_CONFIG_UNKNOWN"
+    #: A rollover naming the era that is already live. Refused because opening
+    #: an era is an append, not an assignment: re-opening the live one wipes
+    #: every score, level, vote budget and faction, freezes every duel and
+    #: retires the board, to arrive at the ruleset the game was already on.
+    #: Distinct from :attr:`era_config_unknown` — the key is perfectly valid,
+    #: the *target* is. 409 rather than 422 because nothing about the request is
+    #: malformed; it is the state of the world that refuses it.
+    era_already_live = "ERA_ALREADY_LIVE"
 
 
 #: The keys of a coded ``detail`` body. Named so the frontend contract is
--- a/backend/faction_slugs.py
+++ b/backend/faction_slugs.py
@@ -150,7 +150,7 @@ def real_faction_slugs(era) -> list[str]:
     lists is a faction a player could actually be in.
 
     The predicate had spelled itself out twice before it got a home (#2708):
-    ``seed.DUEL_FIXTURE_TASK_FACTION_SLUG`` and
+    ``seed.duel_fixture_task_faction_slug`` and
     ``scripts.seed_demo_praxes.fixture_faction_slugs``, both of which left a note
     naming this module as its place — beside :func:`faction_filter_slugs`, the
     leaf every layer may import — the day a third site asked. The integration
--- a/backend/game_config.py
+++ b/backend/game_config.py
@@ -1,10 +1,18 @@
 """
 Single source of truth for all game rules.
 
-Services import EraConfig instances from this file. The database stores
-config_key to record which ruleset was active during each era, but never
-owns the rules themselves. Changing CURRENT_ERA is the one lever that
-switches all live game mechanics.
+Services import EraConfig instances from this file, which owns the *rules*. The
+database owns the **choice**: the latest ``Era`` row's ``config_key`` names which
+of these rulesets is live, and ``services.era.rebind_live_era`` resolves it at
+start-up and again the moment a rollover writes a new row (ADR-0091, #827).
+
+That reverses the older invariant this docstring used to state — "the database
+never owns the rules; changing CURRENT_ERA is the one lever". A mod now ends an
+era from the admin console instead of an owner editing this file and deploying.
+CURRENT_ERA is still the one lever; the hand on it moved.
+
+See "The live era binding" at the foot of this file for why CURRENT_ERA is a
+stable object identity rather than a name that gets reassigned.
 """
 from dataclasses import dataclass, field, replace
 from enum import Enum
@@ -512,24 +520,26 @@ def __getattr__(name: str):
     one.
     """
     if name == "CURRENT_ERA":
-        # The ONE line a rollover moves. Deliberately its own branch: resolving
-        # ERA_1 must not also (re)bind CURRENT_ERA, or the flip is undone by a
-        # side effect. `services.activity_feed` calls `era_config_for_key` on a
-        # stored row's key to label a PAST era, and with the two folded together
-        # the first such call after the flip would rebind CURRENT_ERA to Era 1
-        # for the life of the process — while every module that imported it at
-        # start-up kept the new one. Found rehearsing the flip in #2708.
-        from eras.era_1 import ERA_1 as _era_1
-        globals()["CURRENT_ERA"] = _era_1
-        return _era_1
+        # Deliberately its own branch: resolving ERA_1 must not also (re)bind
+        # CURRENT_ERA, or a flip is undone by a side effect.
+        # `services.activity_feed` calls `era_config_for_key` on a stored row's
+        # key to label a PAST era, and with the two folded together the first
+        # such call after a flip would put the process back on Era 1's rules for
+        # the life of it — while every module that imported CURRENT_ERA at
+        # start-up kept the new one. Found rehearsing the flip in #2708, and
+        # still the reason these are two branches now that #827 makes the
+        # rebinding deliberate.
+        live = _new_live_era(era_config_for_key(COMPILE_TIME_ERA_CONFIG_KEY))
+        globals()["CURRENT_ERA"] = live
+        return live
     if name in ("ERA_1", "ERA_1_FACTIONS"):
         from eras.era_1 import ERA_1 as _era_1
         globals()["ERA_1"] = _era_1
         globals()["ERA_1_FACTIONS"] = _era_1.factions
         return globals()[name]
     if name in ("ERA_2", "ERA_2_FACTIONS"):
-        # Era 2 (Metamorphosis) is authored, not activated: CURRENT_ERA stays
-        # Era 1 above.
+        # Resolving Era 2's config does not make Era 2 live. Only the database
+        # decides that, through `bind_live_era` below (ADR-0091).
         from eras.era_2 import ERA_2 as _era_2
         globals()["ERA_2"] = _era_2
         globals()["ERA_2_FACTIONS"] = _era_2.factions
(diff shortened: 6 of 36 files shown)


pixieofhugs:
> *Review pass generated by AI during `/builder-bot`, 2026-09-04 (overnight run). Not a merge — the merge guard cannot run tonight (#3006). A fix agent is being dispatched onto this branch for the blocking items; CI is the first thing that will run the integration tests.*

# Review — PR #3013 (#827: the latest Era row chooses the live ruleset)
Reviewed 2026-09-04 against merge-base 0d2712a0. Six finder angles + direct verification on the branch.

## Blocking (correctness, CONFIRMED)
1. `backend/services/era.py:487` — `apply_era_reset` rebinds the process-wide `CURRENT_ERA` BEFORE `get_db` commits. A failed commit rolls the DB back and leaves the process on the new era. Bind after the commit, from the era object already in hand.
2. `backend/services/admin_service.py:541` — no server-side refusal of rolling into the era that is already live; only the disabled `<option>` in EraTab prevents a second full destructive reset. Refuse with a coded error when `config_key` equals the live row's key; consider an `expected_current_era_id` token to serialize concurrent rollovers.
3. `backend/services/admin_service.py:559` — the rollover never calls `seed.upsert_era_factions`, so the `Faction` visible/retired mirror stays on the old era until the next deploy (joins refused on visible tiles; a new slug has no FK row). Call it inside `roll_into_era`; `test_admin_era_rollover.py` should assert on `Faction`.
4. `backend/main.py:50` — startup rebind sits between `acquire_single_instance_lock` and the `try/finally` that releases it. Move it inside the `try`.
5. `backend/eras/_template.py:14` — still says `Change: CURRENT_ERA = ERA_N`, the assignment ADR-0091 forbids, and omits the `__getattr__` branch.

## Should fix before merge (docs that contradict the code in the same PR, CONFIRMED)
- ADR-0091 line 164 says `era_reset.py` rolls into the compile-time era; the PR's script now rebinds to the DB's live era.
- `services/era.py:38` docstring: "two moments call this one, and there is no third" — four callers on the branch (main, era_reset, seed, apply_era_reset).
- `schemas/admin.py:226` docstring names `POST /admin/eras/roll`; the route is `PUT /admin/eras/live`, and the docstring ships into openapi.json + schema.d.ts.
- `game_config.py:359` still describes the pre-#827 procedure.
- "157" default-argument sites cited in CLAUDE.md / game_config.py / ADR-0091; the branch has 114. CLAUDE.md should not carry a count at all ("keep it thin so it can't drift").

## PLAUSIBLE / nits
- `scripts/era_reset.py:121` — `rebind_live_era` falls back to `era_1` on an unregistered live key, so the "re-open the live era" script can silently roll back. Refuse instead.
- `EraTab.tsx:100` — hand-rolled confirm instead of `components/confirm/ConfirmDialog.tsx`; `selectableEras` is tested to exclude the live era but the `<select>` renders it disabled (frontend/CLAUDE.md: "Hide unusable controls; don't show them disabled").
- `services/era.py:447` — pre-existing N+1 per character now on an HTTP path with no lock; a timeout + retry opens two eras.
- `tests/conftest.py:21` duplicates `bind_live_era`'s refresh (with a prune loop the PR's own test proves never fires); `seed.stale_era_warning` re-runs the query `rebind_live_era` just made.
- Owner judgement (from the previous batch's close-out, still open): the PR corrects the ruling's MECHANISM (stable identity refreshed in place, not a name rebind). ADR-0091 §2 argues it. Ratify or reject.

## Verdict
Not mergeable as-is: items 1–3 are production-affecting on the most destructive operation in the app. A backend fix agent is being dispatched onto the PR branch; re-review after it lands.

pixieofhugs:
# Review fixes, 2026-09-04

Head: `1c7a180c`, four commits on this branch. Every finding was re-verified against the code before acting; two of them turned out slightly different from the report and are called out below.

---

## 1. Bind after commit — BLOCKING, fixed

`apply_era_reset` no longer touches the live era at all. It only flushes, so there is nothing durable for a bind to stand on.

New `services.era.commit_and_bind_live_era(session, era)` commits and *then* binds, from the `EraConfig` the caller already holds — no re-query, so there is no window in which it could bind something other than what was written. `roll_into_era` calls it as its last act; `scripts/era_reset.py` calls it too, so the script still ends bound to exactly what it wrote.

**Which of the two shapes, and why.** The service owns the commit, not the router. The router cannot own the boundary without also owning *which* era to bind — that is the rollover's decision, not the route's — and `get_db`'s commit lands after the handler has already returned, too late for the handler to bind behind it. `get_db` still commits afterwards; that second commit finds nothing left to write. The exception is documented in both places: in `commit_and_bind_live_era`'s docstring and in `db.get_db`'s, next to the rule it excepts, with an explicit "do not read this as a precedent".

The guard test is untouched and still honest: its allowlist is by *file*, and both callers of the primitive live in `services/era.py`. I updated its docstring to name both rather than one.

Tests added, all at the seam:
- `test_a_failed_commit_leaves_the_live_era_where_it_was` — the defect. A session stub that raises on commit; `CURRENT_ERA` and a def-time-bound default argument are both unmoved.
- `test_the_bind_happens_after_a_successful_commit` — the ordering asserted as an order, by reading the live era from inside the stub's `commit()`.
- `test_apply_era_reset_does_not_bind_anything_itself` — greps the function body, so a future edit cannot quietly move the rebind back in.

## 2. Refuse rolling into the live era — BLOCKING, fixed

New `ErrorCode.era_already_live` = `ERA_ALREADY_LIVE`, **409** rather than 422: nothing about the request is malformed, it is the state of the world that refuses it. Matching sentence in `frontend/src/locales/en/errors.json`.

**Serialization: an advisory lock, not `FOR UPDATE`.** I checked the row-lock option and it does not hold. `SELECT ... FROM era ORDER BY id DESC LIMIT 1 FOR UPDATE` locks nothing on a table with no rows, and — worse — under READ COMMITTED the waiter's *locked row* is re-checked after the lock is granted but its `ORDER BY id DESC LIMIT 1` is not re-planned, so after the winner inserts a new latest row the waiter is still holding the old one and would decide against it. `pg_advisory_xact_lock` taken **before** the live row is read makes the read-decide-insert one indivisible decision and has no empty-table hole. Key `17400827`, the same issue-number convention as `praxis_room._SINGLE_INSTANCE_LOCK_KEY`.

Tests (integration): the refusal writes no row and does not move the live era; the refusal reads the *row* rather than the process's binding, so a drifted process cannot open a duplicate; two overlapping rollovers open exactly one era.

## 3. Re-mirror the faction roster — BLOCKING, fixed

`roll_into_era` now calls `seed.upsert_era_factions(session, era_config)`, after the `Era` row is flushed and **before** `apply_era_reset`, since the reset seats every character in `era.reset_faction_slug`, an FK onto that table.

**No move was needed and no cycle exists.** `seed` imports models, `game_config` and `config` and never a service; `services/era.py` already imports it for `ONBOARDING_TASK_TITLE`, and `admin_service` imports `services.era`, so `seed` is loaded before `admin_service` either way. The mirror stays in `seed` because the seeder is what runs it on every deploy; the rollover is the second caller, not the owner.

Test: `test_the_rollover_re_mirrors_the_faction_roster` asserts the slugs Era 2 drops end `retired` (never deleted — `character.faction_slug` is an FK) and every slug Era 2 carries ends `visible`, or `hidden` for a system row. Which slugs those are is read off the two configs, not restated in the test.

## 4. Startup rebind inside the lock's try/finally — fixed

Moved inside the `try`. The comment no longer says "Never raises", which was false — it now says what is actually guaranteed: a *row* it cannot resolve is survivable (missing row, unregistered key, both fall back and log), but the query itself can fail, and a database down at boot must not leave the single-instance advisory lock held by a process that then dies.

## 5. `eras/_template.py` step 4 — fixed

`Change: CURRENT_ERA = ERA_N` is gone. Step 4 now names the `__getattr__` branch (with why the import is lazy), says the registry key is **appended** and why order matters, and states that adding an era file makes it *offerable*, not live — a mod opens it from Admin > Era, or an operator runs `era_reset.py`. Reassigning the name is called out as moving nothing anyway.

## 6. `scripts/era_reset.py` — fixed, and one test had to move with it

The script now resolves `era_config_for_key(live_row.config_key)` itself and **refuses** with a non-zero exit and a three-line message when that is `None`, rather than inheriting `rebind_live_era`'s compile-time fallback. The resolver falls back because start-up must survive that state; a rollover must not, or "re-open the live era" silently becomes "roll the game back to Era 1" on the operation with no undo.

Nothing binds on the plan-only path any more — the script does not call `rebind_live_era` at all. `print_plan` takes the resolved config as an argument instead of reading `CURRENT_ERA`, so "Nothing was changed" is now literally true, including of this process's live era.

**A consequence worth flagging:** `test_a_rollover_into_a_new_config_key_opens_it_and_closes_the_old` asserted the *old* contract — a live row carrying the previous era's key, rolled forward into the compile-time era (#2705). That window is exactly what ADR-0091 abolishes, and this PR's own `stale_era_warning` rewrite already says the state cannot occur. The test now pins the refusal, keeping the #2705 read it was also protecting (the live era is the latest row whatever key it carries, and it must not 500). A second test was added: a confirmed run ends bound to what it wrote.

Also flagged in the script's docstring and in ADR-0091: **the route refuses what this script does**, deliberately. From the admin page, re-opening the live era is a mis-click; from a shell behind `--yes`, a new season under the same ruleset is the operation asked for by name.

## 7. Docs that contradicted the code — all fixed

- **ADR-0091 ~line 164** — `era_reset.py` re-opens the era the latest `Era` row names, refuses on an unresolvable key, cannot choose a different one. The route/script asymmetry is stated so it does not read as an oversight.
- **ADR-0091 "bound at exactly two moments"** — rewritten to name the real callers: start-up and `seed.py` ask the database via `rebind_live_era`; the rollover binds after its own commit via `commit_and_bind_live_era`, explicitly not inside `apply_era_reset` and not before the commit.
- **`services/era.py:38`** — "two moments call this one, and there is no third" was wrong on the branch. Rewritten around what actually calls it.
- **`schemas/admin.py:226`** — `POST /admin/eras/roll` → `PUT /admin/eras/live`. `python scripts/regen_api_client.py` rerun from the repo root with the venv python; `backend/openapi.json` and `frontend/src/api/generated/schema.d.ts` committed (one line each).
- **`game_config.py:359`** — the `reset_faction_slug` comment described the pre-#827 procedure. Both doors pass the era explicitly now; the historical note is kept as a parenthetical.
- **The "157" count** — measures **114** on this branch (`grep -rn --include='*.py' "era: EraConfig = CURRENT_ERA" backend/`, all of them non-test). Removed from `CLAUDE.md` entirely, since it is a pointer doc that must not carry counts. `game_config.py` and ADR-0091 keep a figure, now with the grep that produces it and a note that it drifts; the incidental repetitions inside those files say "every" instead.

## 8. Nits

- **`tests/conftest.py`** — done. Snapshots via `game_config._new_live_era` and restores with one `bind_live_era(snapshot)` call, so the fixture goes through the real lever instead of re-implementing the refresh (prune loop included).
- **`rebind_live_era`'s two identical fallback branches** — done, collapsed to one `bind_live_era` return with the log line chosen above it.
- **`seed.stale_era_warning` re-querying** — **not done, deliberately.** The code disagrees with the finding here. `rebind_live_era` runs at the *top* of `seed()`, and Phase 2 (`bootstrap_admin`) can insert the very first `Era` row between that read and the warning at the end. Passing the earlier row down would make the "no Era row exists" branch report a state that is no longer true on exactly the run where it matters — a fresh install. The second query is correct, not redundant.
- **EraTab / ConfirmDialog** — left alone as instructed; out of this agent's lane.

---

## What ran, and what CI will run first

Local, in this worktree:

- `backend/.venv/bin/python -m pytest tests/unit -q` from `backend/` — **529 passed**. (Env supplied inline: no `.env` exists on this machine, and unit tests need only the three `Settings` fields.)
- `pytest tests/integration --collect-only -q` — **1242 collected**, no import errors.
- `ruff` clean on every file touched; the ruff ratchet and the uncoded-error ratchet both pass inside `tests/unit`.
- `python scripts/regen_api_client.py` then `git diff --exit-code` on both artifacts — clean (the `api-schema` job).
- Frontend, every step in `.github/workflows/test.yml`: `npm run lint` clean, `npm run typecheck` clean, `npm run typecheck:design-sync` clean, `npm test` **510 files / 10543 passed / 1 skipped**, `npm run build` succeeds, `npm run budget` `WARN JS 150.1 KB (warn>130.9, fail>175.8)` exit 0 — the same pre-existing warning the PR body reports, unmoved by this batch.

**CI is the first thing to execute any of the backend integration tests**, the six new ones included. There is no Postgres in this environment — no docker group membership, no local server, `sudo` needs a password. The new integration cases are:

| File | Test |
|---|---|
| `test_admin_era_rollover.py` | `test_rolling_into_the_era_that_is_already_live_is_refused` |
| | `test_the_refusal_reads_the_row_not_the_process` |
| | `test_two_overlapping_rollovers_open_exactly_one_era` |
| | `test_the_rollover_re_mirrors_the_faction_roster` |
| `test_era_reset_script.py` | `test_a_live_row_naming_no_era_file_is_refused` (rewritten from the pre-#827 assertion) |
| | `test_a_confirmed_run_ends_bound_to_what_it_wrote` |

The unit tests for finding 1 do run here and do pass, which is where the blocking defect's own regression lives.

**Visual QA is still outstanding** — unchanged from the PR body, and nothing in this batch touches a rendered surface except one new sentence in `errors.json`. Worth adding to the eyeball list: the `ERA_ALREADY_LIVE` sentence has no reachable path from the admin page today (the live era is a `disabled` option), so it will only be seen against a stale build or a direct API call. That is the point of it being server-side.

